PR #12, opened 2022-05-14
WIP: Runspaces

DeveloperDurp:
I created a sample of how I do runspaces in the scripts I have out in production. I like to keep the main form inside the initial thread and call runspaces from there. Not as pretty but I also don't use the xaml files like you do.


DeveloperDurp:
Figured out the runspaces with your design, working on converting it into a working version.

ChrisTitusTech:
Thank you so much for this.

PR #19, opened 2022-05-15
Added DISM command and button to tool

Rexzarrax:
This request adds a button into the config(tab 3)/fixes section to run the DISM command in a new elevated PowerShell window. The window then will remain open until the user press enter.

Tested in:
[x] Windows 10
[ ] Windows 11

DISM seems to be a little better than ```SFC /scannow```


Carterpersall:
It might be better to have a 'Scan for Corruption' button instead that runs Chkdsk, SFC /scanow, DISM, then SFC /scannow again as that is the best way to catch everything

Rexzarrax:
Good one!. I have added that in the commit above.

PR #50, opened 2022-05-19
Installs working under runspaces

DeveloperDurp:
Got the installer to work under a runspace.

All output is sent to a log file located in the %appdata% temp folder. I believe it should install winget based off the original code (made the modification to output to the logfile). There is however no error handling if winget is having an issue (like my win10 instance) 

image
image

image
image

One con is I am not sure of a way to pass functions to a runspace without creating a separate module first. Open to ideas.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,27 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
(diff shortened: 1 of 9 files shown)


DeveloperDurp:
Random shower thought. Could use the invoke dispatcher and have the main form thread write to the logs instead of defining a write-log function in each runspace section. Since writing to a file is fairly lightweight it shouldn't impact gui performance. That also brings up the question of what it would take to keep a runspace open waiting for commands.

DeveloperDurp:
Tested out the winget install on a fresh Win10 install 

image
image

However looks like a wait timer should be added as it failed right away once the install was finished

image

Carterpersall:
I've run into this issue in the past. Sometimes just pausing for a few seconds works, sometimes it requires a shell restart, and sometimes it even requires a computer restart. I previously used this to force the winget alias to be set:
```
$installationPath = (Get-AppxPackage Microsoft.DesktopAppInstaller).InstallLocation
               Set-ExecutionAlias -Path "C:\Windows\System32\winget.exe" -PackageName "Microsoft.DesktopAppInstaller_8wekyb3d8bbwe" -EntryPoint "Microsoft.DesktopAppInstaller_8wekyb3d8bbwe!winget" -Target "$installationPath\AppInstallerCLI.exe" -AppType Desktop -Version 3
```
This might be outdated though so it definitely needs to be tested.
This also only applies to the manual installation method, so it would need to be adapted for the automatic install method.

PR #52, opened 2022-05-20
Arguments

### LINKED ISSUE #51: Feature Request: Arguments and Unattended Running
**Is your feature request related to a problem? Please describe.**
It would be very useful if the script had the ability to run unattended and silent. This would be useful for people who know what they want to run and want to run a command and wait for it to finish. It would also be helpful for people like IT's and System Admins to use if they wanted to deploy the script during the setup of a network of computers.

**Describe the solution you'd like**
The solution would be much like what I implemented in ChrisTitusTech/win10script#331 where you pass arguments into the script and the script runs them and then exits (if the accompanying arg is set). It will be much more difficult to implement in the new variant of the script though. 
I will attempt to implement it in a pull request when the current restructuring of the script for runspaces is complete as I don't want to interfere with that.

**Additional context**
I'm putting this here for others to refer to and offer suggestions and to prevent someone else from opening an issue on it while I'm working on it.

Carterpersall:
- Closes #51
- Depends on #67
- Using $env:args, the script scans for arguments; then if found, Invokes the codeblock associated with said arg.
  - \b is to prevent mutual words from causing issues (Ex: Install and InstallUpgrade would conflict without \b)

This will be expanded for the rest of the script as the script restructure for runspaces is finished.

### DIFF
--- a/winutil-test.ps1
+++ b/winutil-test.ps1
@@ -1301,17 +1301,17 @@ $xaml.SelectNodes("//*[@Name]") | ForEach-Object {$global:sync["$("$($_.Name)")"
 # Arguments
 If($env:args){
     Write-Host "Arguments Detected, Running Args"
-    If($env:args -match '\bInstallUpgrade\b'){Invoke-Runspace $InstallUpgrade}
-    If($env:args -match '\bUndoTweaks\b'){Invoke-Runspace $undotweaks}
+    If($env:args -match '\bInstallUpgrade\b'){Invoke-Command -scriptblock $InstallUpgrade}
+    If($env:args -match '\bUndoTweaks\b'){Invoke-Command -scriptblock $undotweaks}
     If($env:args -match '\bPanelControl\b'){cmd /c control}
     If($env:args -match '\bPanelNetwork\b'){cmd /c ncpa.cpl}
     If($env:args -match '\bPanelPower\b'){cmd /c powercfg.cpl}
     If($env:args -match '\bPanelSound\b'){cmd /c mmsys.cpl}
     If($env:args -match '\bPanelSystem\b'){cmd /c sysdm.cpl}
     If($env:args -match '\bPanelUser\b'){cmd /c "control userpasswords2"}
-    If($env:args -match '\bDefaultUpdates\b'){Invoke-Runspace $Updatesdefault}
-    If($env:args -match '\bDisableUpdates\b'){Invoke-Runspace $Updatesdisable}
-    If($env:args -match '\bEnableSecurity\b'){Invoke-Runspace $Updatessecurity}
+    If($env:args -match '\bDefaultUpdates\b'){Invoke-Command -scriptblock $Updatesdefault}
+    If($env:args -match '\bDisableUpdates\b'){Invoke-Command -scriptblock $Updatesdisable}
+    If($env:args -match '\bEnableSecurity\b'){Invoke-Command -scriptblock $Updatessecurity}
     If($env:args -match '\bQuitAfter\b'){Break}
 }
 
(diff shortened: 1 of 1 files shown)


DeveloperDurp:
Looks like it should work though it might need to be lower in the script as the functions/scripts wouldn't be loaded into memory yet. Might even need to go until right before line 1316. 

Hmm....

We might want to move everything that generates the GUI lower in the script as well. That way the functions/scripts load first and if we want CLI only it doesn't have to spend time loading things it doesn't need.

DeveloperDurp:
Well that's interesting lol, ran it twice. But I had to put it at the bottom of the script.

image

(Code at top)
image

Carterpersall:
The positioning issue should be fixed

DeveloperDurp:
Quick test of the quit after I think we will need to add some logic so it can wait for the runspaces to finish. Or we might be able to do an invoke-command -scriptblock $InstallUpgrade since we don't have to worry about a gui

I will see what's going on with the double running of the install upgrade. Might need to rework the runspaces a little.

Carterpersall:
Doing that seems to work
For me though it ran a _few_ more than 2 times:
image

DeveloperDurp:
Perfect! I will work on the runspaces with the CLI in mind

Carterpersall:
> Doing that seems to work For me though it ran a _few_ more than 2 times: image

I can't seem to recreate this issue so I dunno what's causing it to happen only occasionally

DeveloperDurp:
Perhaps lingering runspaces from the last run, it works perfectly on my win10 instance.

Carterpersall:
> Perhaps lingering runspaces from the last run, it works perfectly on my win10 instance.

Oh yeah, that would make sense as that was my first time trying the different method.

Carterpersall:
I might want to make another arg where if it's passed, none of the message boxes show. Currently, the message boxes pause the script making it not unattended. Another way to do it would be to have a check every time a message box is shown that checks if the argument is passed, then doesn't show the box when if it is passed.
Though this would mean that if QuitAfter isn't passed and the corresponding button is pressed, no message boxes would show. This could be prevented though by setting $env:args to an empty string before the form is opened and after the argument check.

DeveloperDurp:
Yeah, that's something I am going to work on. I might have to separate the scripts a little and have a gui portion that controls the gui and formats the data into the same arguments you would use with say winget. That way you could pass $arg = "install:git.git" and it will install it the same way. 

Also going to play around with the error handling to have it output errors to the terminal window and also allow you the option to do verbose in case you want to see what the script is doing.

DeveloperDurp:
@Carterpersall FYI, the new script to call is $sync.ScriptsInstallPrograms, separated the GUI portion from the install script to make it easier to get the CLI functions to work without spamming message boxes.

DeveloperDurp:
Once #50 gets merged you can add Installs to your arguments. Though we would need to add some logic so you could add in programs

Something like 
```
$env:args = "UndoTweaks QuitAfter Install:git.git,winscp.winscp"

If($env:args -match '\bInstall\b'){
    Write-Verbose "Running"
    $Programstoinstall = (($env:args-split " " | Where-Object {$_ -like "install*"} ) -split ":")[1]
    Invoke-Command -scriptblock  $sync.ScriptsInstallPrograms -ArgumentList $Programstoinstall
}
```

We can also add this to get output in the console window
`If($env:args -match '\bVerbose\b'){$VerbosePreference = "Continue"}`

Carterpersall:
Great Idea! When I find the time, I'll implement it.

Carterpersall:
Added it in the most recent commit

DeveloperDurp:
$sync.ScriptsInstallUpgrade will also be available in the same MR

Issue #39, opened 2022-05-18
ERROR: The variable '$arch' cannot be retrieved because it has not been set.


techroy23:
**Describe the bug**

**To Reproduce**
1. Go to 'updates'
2. Click on 'Default OOB Settings'
3. See error

**Expected behavior**
Will show the done button

**Screenshots**
![image](https://user-images.githubusercontent.com/53241974/169048341-29b87e89-2e90-4fc3-8a2c-b7f4b7cfacb1.png)

**Additional context**

```
Setting BITS StartupType to Automatic
Setting wuauserv StartupType to Automatic
Enabling driver offering through Windows Update...
Enabling Windows Update automatic restart...
Enabled driver offering through Windows Update
1. Stopping Windows Update Services...
2. Remove QMGR Data file...
3. Renaming the Software Distribution and CatRoot Folder...
4. Removing old Windows Update log...
5. Resetting the Windows Update Services to defualt settings...
6. Registering some DLLs...
7) Removing WSUS client settings...
ERROR: The system was unable to find the specified registry key or value.
ERROR: The system was unable to find the specified registry key or value.
8) Resetting the WinSock...
9) Delete all BITS jobs...
10) Attempting to install the Windows Update Agent...

ERROR: The variable '$arch' cannot be retrieved because it has not been set.
ERROR: at <ScriptBlock>, <No file>: line 1342 ERROR: at <ScriptBlock>, <No file>: line 1489 ERROR: at <ScriptBlock>, <No file>: line 1
ERROR EXCEPTION: System.Management.Automation.RuntimeException: The variable '$arch' cannot be retrieved because it has not been set. ERROR EXCEPTION:    at System.Management.Automation.VariableOps.GetVariableValue(VariablePath variablePath, ExecutionContext executionContext, VariableExpressionAst varAst) ERROR EXCEPTION:    at System.Management.Automation.Interpreter.FuncCallInstruction`4.Run(InterpretedFrame frame) ERROR EXCEPTION:    at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)

Sleeping for 60m to give you time to look around the virtual machine before self-destruction...
```

sergiodevelop78:
Hi, the $arch variable is not defined.
I put these lines after line 1341 , and now it works.
```

   $arch64=[Environment]::Is64BitOperatingSystem
    if ($arch64 -eq $true) {
        $arch=64
    }
    else {
[...]

PR #61, opened 2022-05-27
Fix bug #39

sergiodevelop78:
Define variable $arch and get the value (64 bits or 32 bits)
   $arch64=[Environment]::Is64BitOperatingSystem
    if ($arch64 -eq $true) {
        $arch=64
    }
    else {
        $arch=32
    }


Carterpersall:
Duplicate of #40

Implementation of the fix is different though

ChrisTitusTech:
Not needed and incorrect fix for dumping all the xaml into the main file. Denying PR.

PR #67, opened 2022-06-08
All functions working with runspaces

DeveloperDurp:
Made a couple big changes that made the older PRs incompatible. Creating this as a draft so you can keep tabs of progress. Will release it when it's ready for a "1.0" release


ChrisTitusTech:
Looks incredible... Thanks for all the work you are doing on this.

DeveloperDurp:
@ChrisTitusTech @Carterpersall 

This is the issue I am running into and could probably use another pair of eyes, I merged the "completed" tweaks code (starts at 431) into this MR as I might be missing something. GUI seems to crash but things seem to work if called via command line. GUI will look like it's working and then freeze after one of the foreach loops. (IE if I comment out Registries it will then proceed to lockup after Services) If you attempt to interact with the gui you will be greeted with the error invoke-button not recognized so I am curious if something is happening to the original thread. 

I will start working on the config/updates tab to keep things moving

image

Carterpersall:
@DeveloperDurp I'm definitely getting the same issue as you here. It seems to usually lock up at the same point which also is weird. It's as if some commands work but others don't for some arbitrary reason. The only thing I can think of doing is going through the script and commenting out every command that causes a lockup and trying to find a common denominator. I unfortunately don't have the time to go through that process at the moment and won't be able to for at least the next two weeks.

Maybe you could try another PowerShell multithreaded implementation like PSJobs and see if that works but in my attempts I couldn't get it to work at all.

Carterpersall:
@DeveloperDurp I got an implementation using Jobs working on my branch here: https://github.com/Carterpersall/winutil/tree/Multithreaded
It seems that Jobs lacks the issues with elevation and quirkiness the previous implementation had so it might end up being the better option.

DeveloperDurp:
@Carterpersall Might be the route to be taken if Chris wants to switch over quickly. First downside I can think of as downsides to using jobs instead of runspaces is powershell spins up a new instance of powershell which does increase CPU and memory utilization. Question of how much of an impact I guess is up for debate.

Been working on a work related project that also uses Powershell GUIs and have some ideas I will try and implement to streamline this.

ChrisTitusTech:
I'm still on the fence about Runspaces vs Jobs. 

Right now I am brushing up on my C# to implement a better interface. I will come back to this at some point.

DeveloperDurp:
Another idea that I have is calling a brand new powershell.exe and feeding it what script to run and the options selected. Would allow the gui to run as a normal user and then escalate permissions per script. Since it would pop up a new shell you could have the logs output directly so users could see if the script ran into errors instead of programming it all into the gui.

Carterpersall:
I might take another stab at this as I've been learning runspaces while working on https://github.com/kiedtl/winfetch/pull/150

I might start from the ground up and see at what points I run into issues

DeveloperDurp:
In my testing it seems limited to the tweaks section. It's possible one of the methods I used to cut down on repeat code is causing the issue. I haven't noticed too much fuss on the other sections of the script so far.

I had to step away for a while to prevent burn out from my day job and now my Windows instance is acting a little buggy. Once I get that going again I will take a stab at it again.

Carterpersall:
In converting winfetch for use with runspaces, I ran into an issue on two of the functions within it that would randomly cause almost completely unrelated issues where two types of commands within a section of the script (outside of the runspace) would suddenly stop being recognized. Maybe the issue we're having here is related to that?

DeveloperDurp:
Sounds like the same issue, my guess is probably something powershell specific but if we were writing this in c# this likely would have no problems. Perhaps a powershell garbage collection is clearing those functions to make room for the runspace.

Carterpersall:
The issue in winfetch seems to appear only sometimes when a specific function is called. It is called the exact same way in multiple other runspaces within the same script with no issues, it only causes problems within two of them. What makes it even weirder is that it seems to be the function call that does it and not anything that occurs within the function. So in two seemingly normal runspaces, a call to a function that is also used in other runspaces causes PowerShell to absolutely implode for seemingly no reason.

Carterpersall:
The issue in winfetch seems to appear only sometimes when a specific function is called. It is called the exact same way in multiple other runspaces within the same script with no issues, it only causes problems within two of them. What makes it even weirder is that it seems to be the function call that does it and not anything that occurs within the function. So in two seemingly random runspaces, a call to a function that is also used in other runspaces causes PowerShell to absolutely implode for seemingly no reason.

DeveloperDurp:
Anyone run into this issue with winget? I run it once, get this error and upon retrying it works. 

image

Carterpersall:
Found this while working on something else: https://github.com/PowerShell/PowerShell/issues/4003

Understood it enough to see that it could be the source of the issues we've been having, but it's a bit too technical for me to know for sure.

DeveloperDurp:
@Carterpersall I think I found the exact cause for the lockup. Doing a rewrite of the tweaks section and it seems when using a try block and using -erroraction stop locks up the runspace when you encounter an error.
 
Will be interesting to do error handling without try/catch

DeveloperDurp:
Seems to be any errors will break the runspace. Removed all error handling and seems to be working

image

DeveloperDurp:
@ChrisTitusTech @Carterpersall take a look at the script. Everything seems to be in working order. Only things left is working with error handling inside runspaces to avoid crashes. 

I also updated it to work with the arguments like we discussed some time ago. It also will not load the GUI if arguments are detected 

Not sure of everything that has been done since I started this branch but I have made some updates to bring it to feature parity. 

image
image
image

ChrisTitusTech:
Very cool, I'll dive more into this and see if we can eventually get it merged into main.

ChrisTitusTech:
Manually rebased with new old archive folder in commit d5862191926dfaeccb71a71372e31f8d8ed0b652

ChrisTitusTech:
Added new commit to first check if service exists before trying to set it. 

Script was getting stuck on setting services without this modification.

PR #94, opened 2022-07-27
Test

FLKaplan20271:
Hey Chris - Its Toradon,
In one of my 9 lives, I was a deveoper for a short time.  Have you thought about making nicer screens through c#?  If you have, then I guess this is all for naught.  I want to see if I can do it so I am asking if I could have access to the test branch.  I will get permission and I probably will not end up with time to do it, but it has been a thing of mine for a while that I have wanted to add my contributions to github.  So anyways, really good progress so far and I want to help - forgive my rambling I have had 3/4 of a beer and I am a bit more relaxed than I was 20 minutes ago.
So there it is.


mrg9999:
I see now it runs the installer direct from the source but it's always selecting unattended installs. 
Is there a log file? It just closed the console without even asking

Example
Freecad wants me to click next, next etc like a Neanderthal
2022-09-14 - freecad

Nice work so far, but needs a tad more 

tad = English for a little bit

PR #101, opened 2022-08-10
Translucenttb Installation fixed

Crusader-Boss:
I found out the 'TB' at the end should be in small letter s or else winget dosent install it.


ChrisTitusTech:
The releases for this program has moved to the ms-store and can't be installed via this fix the actually name for install is this `9PF4KZ2VN4W9`
and can be installed via `winget install -e 9PF4KZ2VN4W9`

PR #132, opened 2022-09-14
Added additional gaming clients/tools

eSKylezZ:
Added additional gaming clients/tools

### DIFF
--- a/MainWindow.xaml
+++ b/MainWindow.xaml
@@ -78,6 +78,14 @@
 							<CheckBox Name="Installepicgames" Content="Epic Games Launcher" Margin="5,0"/>
 							<CheckBox Name="Installgog" Content="GOG Galaxy" Margin="5,0"/>
 							<CheckBox Name="Installsteam" Content="Steam" Margin="5,0"/>
+							<CheckBox Name="InstallItch" Content="Itch.io" Margin="5,0"/>
+							<CheckBox Name="InstallHumble" Content="HumbleBundle" Margin="5,0"/>
+							<CheckBox Name="InstallAmazonGames" Content="Amazon Games" Margin="5,0"/>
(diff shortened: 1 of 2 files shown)


eSKylezZ:
Additional notes:

Main website links for each client/tool:

- Ubisoft: https://ubisoftconnect.com/
- Vortex: https://www.nexusmods.com/about/vortex/
- Loot: https://loot.github.io/
- Itch: https://itch.io/app
- Humble: https://www.humblebundle.com/membership/collection
- CurseForge: https://www.curseforge.com/
- Amazon: https://www.amazongames.com/en-us/support/prime-gaming/articles/download-and-install-the-amazon-games-app
- Paradox: https://www.paradoxinteractive.com/our-games/launcher

Triplesweet:
i don't think loot should be included, not many people even in the Skyrim modding community uses it, it would just make the list longer

LinuxFurry:
> i don't think loot should be included, not many people even in the Skyrim modding community uses it, it would just make the list longer

Not to play devils advocate with your suggestion, but if that is the case, then Vortex should go as well, as not every user will be modding their games.

Triplesweet:
Understandable, but people who use loot most probably use vortex as well, or another mod manager, but not everyone that uses vortex uses loot, but if you believe having loot there can be helpful, by all means, thank for the hard work 👍

LinuxFurry:
> Understandable, but people who use loot most probably use vortex as well, or another mod manager, but not everyone that uses vortex uses loot, but if you believe having loot there can be helpful, by all means, thank for the hard work 👍

Your scenario lines up with my use case for Vortex, as I don't heavily rely on it due to the built-in functionality of LOOT that is within Vortex already when it sorts out load orders.

However, I do still use it every once and a while just so I can see any warnings/errors that Vortex doesn't catch or cannot display through the UI they have.

eSKylezZ:
Missing applications.json file.

PR #133, opened 2022-09-14
Add a check for $isLinux

couleurm:
This is a very simple if statement, I don't think there's much to explain

```PowerShell
if ($IsLinux){ # PowerShell on Linux has that read-only bool variable set to $true
   Write-Host "This utility is exclusively designed for Windows"
   return
}
```

### DIFF
--- a/winutil-test.ps1
+++ b/winutil-test.ps1
@@ -5,6 +5,11 @@
    Version 0.0.1
 #>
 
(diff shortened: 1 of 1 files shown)


vulcan-dev:
I don't see the need for this, it's already clearly stated that it's meant to be used by windows.

couleurm:
doesn't hurt to add :shrug:

PR #135, opened 2022-09-14
Added UAC Disable

PrincessAkira:
Hello Chris,

What did i change?
So instead of using EnableLUA to 0, im using the Method of Disabling the Prompt for Admin Users.
What im doing here is changing `HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\System` to 0 (5 is default)

Why do you ask? 
Some tools tend to break or you maybe want someone else to use your PC.
It is annoying to be alerted by certain Programs to ,,Your UAC is turned off,, and this works around that completly.

Tldr
If you find a way to edit Group Policy via Powershell it makes this way Easier.
All im doing is changing this value here.
image


PrincessAkira:
Tldr
VSCode merged in wrong branch. Lemme fix

PR #143, opened 2022-09-15
added software development related packages

bbSnavy:
coming from linux, this project is kind of confusing for me, I hope I haven't missed anything. this should install rust and alacritty from winget.
the packages on there are quite outdated when I browsed it, some are even behind apt while some others straight up don't show up in PATH. perhaps something has to be done... at least it's working so far, I managed to install the two packages on my windows laptop.


ChrisTitusTech:
Looks good, just make sure to do future PR's to the test branch

PR #148, opened 2022-09-15
Add dotnet install

SteffenCarlsen:
Added dotnet install into the utility. 

I am in doubt about the explicitness of the checkbox test description. I've named it ".NET SDK" - But realistically the goal is to make it install "The current LTS version of the .NET SDK" - Hence not .NET Framework.

I also took the liberty (And let me know if I should revert this) to reformat the json file according to the standard formatting rules.


ChrisTitusTech:
I'll add this in the next update, but I'm seeing a bunch of conflicts on my end. So I will close this PR.

PR #153, opened 2022-09-15
Add Nahimic Service

thecuz1:
Added Nahimic Service to list of services to set to manual. Nahimic seems to provide little improvement to audio for some resources consumed and tends to cause issues with specific games. Comes preinstalled on lots of prebuilts and laptops.


ChrisTitusTech:
Good commit, I hate that service for Realtek users. Causes nothing but problems.

PR #165, opened 2022-09-18
Forgot to add the XML Commit :D

PrincessAkira:
please double check chris bc im not good with XML here :D


d1payan (commented):
It will be "MiscTweaksDisableUAC"
Not "WPFMiscTweaksDisableUAC"

PR #177, opened 2022-09-21
Longer Thing

PrincessAkira:
- Added Undo for TPM and UAC
- Added TPM Check Skip
- Used the actual right values for GameDVR and FSO
- Added Long Paths

### DIFF
--- a/winutil.ps1
+++ b/winutil.ps1
@@ -612,11 +612,13 @@ $WPFtweaksbutton.Add_Click({
 
         If ( $WPFEssTweaksDVR.IsChecked -eq $true ) {
             If (!(Test-Path "HKCU:\System\GameConfigStore")) {
-            Set-ItemProperty -Path "HKCU:\System\GameConfigStore" -Name "GameDVR_DXGIHonorFSEWindowsCompatible" -Type Hex -Value 00000000
-            Set-ItemProperty -Path "HKCU:\System\GameConfigStore" -Name "GameDVR_HonorUserFSEBehaviorMode" -Type Hex -Value 00000000
-            Set-ItemProperty -Path "HKCU:\System\GameConfigStore" -Name "GameDVR_EFSEFeatureFlags" -Type Hex -Value 00000000
-            Set-ItemProperty -Path "HKCU:\System\GameConfigStore" -Name "GameDVR_Enabled" -Type DWord -Value 00000000
+                New-Item -Path "HKCU:\System\GameConfigStore" -Force
             }
+            Set-ItemProperty -Path "HKCU:\System\GameConfigStore" -Name "GameDVR_DXGIHonorFSEWindowsCompatible" -Type DWord -Value 0 -Force
+            Set-ItemProperty -Path "HKCU:\System\GameConfigStore" -Name "GameDVR_HonorUserFSEBehaviorMode" -Type DWord -Value 0 -Force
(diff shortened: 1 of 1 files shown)


PrincessAkira:
tldr i dont agree with the values from https://github.com/ChrisTitusTech/winutil/commit/a785c75bee99a2fc9d4e00a280293582e849960d

d1payan:
> tldr i dont agree with the values from [a785c75](https://github.com/ChrisTitusTech/winutil/commit/a785c75bee99a2fc9d4e00a280293582e849960d)

You don't have to agree with my values. Just Put your all registry values outside if condition and inside if condition only put the new item code, like what I did. Otherwise, any of the registry tweaks will not work. 

Because the code is saying if the path (HKCU:\System\GameConfigStore) is NOT there then only apply these tweaks. But that is not what we want , right ?

So, pull the code first from latest changes then create a pull request with whatever value you want.

PrincessAkira:
> > tldr i dont agree with the values from [a785c75](https://github.com/ChrisTitusTech/winutil/commit/a785c75bee99a2fc9d4e00a280293582e849960d)
> 
> You don't have to agree with my values. Just Put your all registry values outside if condition and inside if condition only put the new item code, like what I did. Otherwise, any of the registry tweaks will not work.
> 
> Because the code is saying if the path (HKCU:\System\GameConfigStore) is NOT there then only apply these tweaks. But that is not what we want , right ?
> 
> So, pull the code first from latest changes then create a pull request with whatever value you want.

i meant the one i used instead of the hex ones!!!
:D
everythings dandy ^-^
theres just other values one can use. should have worded better mb <3

d1payan:
> > > tldr i dont agree with the values from [a785c75](https://github.com/ChrisTitusTech/winutil/commit/a785c75bee99a2fc9d4e00a280293582e849960d)
> > 
> > 
> > You don't have to agree with my values. Just Put your all registry values outside if condition and inside if condition only put the new item code, like what I did. Otherwise, any of the registry tweaks will not work.
> > Because the code is saying if the path (HKCU:\System\GameConfigStore) is NOT there then only apply these tweaks. But that is not what we want , right ?
> > So, pull the code first from latest changes then create a pull request with whatever value you want.
> 
> i meant the one i used instead of the hex ones!!! :D everythings dandy ^-^ theres just other values one can use. should have worded better mb <3 I wanted to pull yours but then it wouldnt let me PR mine and i would have needed to discard it full due to Github logic _grrr_

I am not saying your values are bad :D
But put it outside if loop or just remove the if loop because this path ((HKCU:\System\GameConfigStore) exists by default. No need to test it.

But you can still do it right ? Just do the changes what I told and then commit again. It will be automatically in the PR. Or, you can just sqash the new commit with previous commit.

PrincessAkira:
> d1payan

Thats what i meant.
I wanted to pull your commit but Github wouldnt let me or it discards my Pushes :D
image

PrincessAkira:
Just did it by hand now. God i hate github

d1payan (commented):
Yes that is what I was trying to say.

d1payan:
> Just did it by hand now. God i hate github

I am new to it. So, just learnt some things in these few days :D

PrincessAkira:
Idk why it has merge errors. ima fix em somehow

Issue #189, opened 2022-09-24
Winget doesn't install on Fresh install of Win11


DeveloperDurp:
**Describe the bug**
I have a clean install of Windows 11 and when trying to install an application it goes to install winget as it's not installed. However windows complains.

**To Reproduce**
1. Check an application
2. Click Start Install

**Expected behavior**
Winget gets installed and then installs the application

**Screenshots**
![image](https://user-images.githubusercontent.com/74198206/192098925-081f9481-7568-4673-88d6-41bfe99d9623.png)

It also goes through and still tries to install everything even though winget fails to install.

DeveloperDurp:
Might have resolved it by changing the paths to $env:userprofile\AppData\Local\Temp

![image](https://user-images.githubusercontent.com/74198206/192111251-8a52a13a-d83b-4bb6-b824-b90410c1a948.png)

DeveloperDurp:
Most likely could have been linked to Time being off as with other issues I was having. Keeping MR open as it's probably better practice to use the temp folder instead of the current directory.

PR #192, opened 2022-09-24
Move winget install to user temp folder

DeveloperDurp:
This fixed the issue I was having in https://github.com/ChrisTitusTech/winutil/issues/189#issue-1384663355


ChrisTitusTech:
@DeveloperDurp Going to modify this PR the variable you were looking for is `$env:TEMP` instead of the long $env:userprofile one.

PR #213, opened 2022-09-30
Added Opera GX for install

padsalatushal:


ChrisTitusTech:
I absolutely hate Opera GX and wouldn't recommend my worst enemy installing that browser. I won't allow this to be added to my project. 
From the shady loan agreements the opera company made, to the bloat that Opera GX has, and the fact they heavily promoted it through YouTube and sent me 10+ emails... I just absolutely despise anything with the Opera name.

PR #223, opened 2022-10-03
Added option for set classic right click menu

padsalatushal:
Added option for set classic right click menu under tweak section
also added into undo tweaks for restore default right click menu layout
also fix disable notification variable name 
i tested both features on different windows 11 vm multiple times its works fine


padsalatushal:
Regrading setting classic right click menu layout it need explorer.exe to restart for it to work

PR #224, opened 2022-10-04
Added option for remove microsoft edge

padsalatushal:
Added option for remove microsoft edge
i put script that remove microsoft edge in my github repo and in this script i added option for remove microsoft edge that run that bat file through 
iwr -useb https://raw.githubusercontent.com/padsalatushal/Edge-Removal/main/Edge_Removal.bat | iex

i tested this it work fine but for some reason it work slow in vm but it work fast in bare matal windows

also fix minor issue in line 1078


ChrisTitusTech:
bringing edge.bat inside the project. I may adapt this from bat file to ps1 in the future. I don't want to link to external projects that I won't know when they are updated.

PR #238, opened 2022-10-08
Added Apply Google's DNS to all connections under disable Telemetry s…

padsalatushal:
…ection


ChrisTitusTech:
Love the idea, but this would mess all my systems up as I host my own DNS and don't use google dns. This can be added as an option, but never part of the essential tweaks.

Personally Cloudflare and Level3 will be more performant if doing this to systems.
1.1.1.1
4.2.2.2 
Respectively.

PR #262, opened 2022-10-12
Confirmation dialog to warn of read-only mode when not running as administrator

craigmjackson:
Users may not understand the error messages in the command line


ChrisTitusTech:
PR Shows no changed files

PR #273, opened 2022-10-16
Added Programs For Install!

josephistired:
_Added Krita For Install_
_Added Kodi For Install_ 
_Added Telegram For Install_
_Added EA Origin For Install_
_Added Classic Open Shell For Install_
_Added Strawberry For Install_
_Added SQL Management Studio For Install_

**Also removed Spotify due to error with installation.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,10 @@ winutil.pdb
 *.zip
 package.psd1
 /.idea/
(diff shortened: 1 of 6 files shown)


ChrisTitusTech:
Will need to redo with new format and structure.

I am using runspaces now.

PR #275, opened 2022-10-17
Added confirmation of readonly when running as unprivileged user

craigmjackson:
Users may not understand the error messages in the command line

Apologies for my Github ignorance.  I am a daily GitLab user.

Thanks!


DeveloperDurp:
I do have something similar in my PR #67 once that all gets merged in. I also have it where it will attempt to relaunch itself as admin.

PR #294, opened 2022-10-19
Github actions to update script based on branch

DeveloperDurp:
@ChrisTitusTech you might like this addition

I figured out the github actions needed to modify the script so it injects the branch name into a variable. Meaning we should now be able to call the script from any branch (with this updated code) and it will run with everything inside that branch. Note I only modified the runspace.ps1 and a POC. 

Then you could move to a release branching structure instead of running everything off main. From here you change your link to point that specific release. That way if you that something broke it's easy to revert back. Plus it would make testing test branches much easier.
IE https://github.com/DeveloperDurp/winutil/tree/release/1.0.0


ChrisTitusTech:
Love the idea. Also will make sure I don't botch a merge by forgetting to change the XAML variable.

DeveloperDurp:
pipeline failed, I think you have to make a modification to the repo to allow the pipeline secret.

ChrisTitusTech:
Ok

DeveloperDurp:
I think going to settings > actions > general and setting this should do the trick

image

ChrisTitusTech:
should be good now

DeveloperDurp:
Looks to be working with no issues, I'll create a PR to update the main script 

image

padsalatushal:
@DeveloperDurp when i try to install glary utility with runspace script it show error 
image

DeveloperDurp:
@padsalatushal Yep, that's something I missed in my conversion. Pretty much needs to be added to the applications.json file, and if you find any other applications that give the same error. 

If you want to take a shot at updating the code you can use the helper script and use lines 36-50

doing a quick look your variables should be.
$NameofButton = "Installglaryutilities"
$WingetCommand = "Glarysoft.GlaryUtilities"

padsalatushal:
@DeveloperDurp thanks for the help and tip. It takes time to understand the runspace different files and structures but now I understand it is amazing. I am able to fix the issue. First, I fix it without running the helper script then try with the helper script both worked. Your work is amazing.

PR #300, opened 2022-10-19
Save and Load option in install menu

NotBugJustFeature:
You can now export the selected programs to a file and restore it later.


DeveloperDurp:
It might be worthwhile to hook into the config/applications.json file since you are going through the effort of replacing all the if statements. I have been using that with the runspace conversion but haven't ported it over to the current version.

NotBugJustFeature:
Ok.
For ease of use I reworked the application.json and implemented it.

ChrisTitusTech:
Ran into issues with the FindForm not running properly I left them on main-test for now and I'll change the branch to bad-test

Installs kept failing and wasn't grabbing the WPF variables properly.

Issue #307, opened 2022-10-20
Missing items in runspace.ps1


DeveloperDurp:
The following items have been found to not be migrated over to the runspace.ps1 script

let me know if any others are found

Thank you,

-----
Will be merged in with #308 
- [x] Setup Autologin
- [x] Reset Windows Update
- [x] Set Dark mode
- [x] System Corruption Scan

padsalatushal:
system corruption scan

DeveloperDurp:
@padsalatushal feel free to take a look at the PR, the existing config files wouldn't work in these cases.

PR #308, opened 2022-10-20
Feautre/updatetweaks

DeveloperDurp:
Fixes for #307


ChrisTitusTech:
Accepted a bad PR right before I merged this, you might need to do another PR to grab your changes. 

That is the last time I accept more than one PR at a time. Sorry!

DeveloperDurp:
I think I should expand on the variables and do the github URL as well. That way you can test 100% their branch to make sure things are working. Might also implement some pester scripts

ChrisTitusTech:
I should have just checked out the PR and ran it. 

I keep forgetting some folks push PRs that are broken or untested. I'm against using a variable for the entire URL that can lead to some malicious use.

PR #357, opened 2022-11-04
Add Toggle for System Dark Mode

lavavex:
Adds toggle for System dark mode as well as App Dark mode to Tweaks. This is to fix issue #354 . Tested and Working on Windows 10 22H2. Was unable to figure out how to add the functionality to the current buttons so I made new buttons for system and renamed the current buttons to specify App dark mode. Feel free deny merge and do it better if you want as i have no clue how to use github properly and this is my first pull req. This was originally forked from main and it seems that some confilcts with main and test are present... The only edits i actually made were lines 219-222 in winutil.ps1 and lines 1395-1411 in MainWindow.xaml (sorry)

Issue #354, opened 2022-11-04
Dark Mode Toggle does not change Windows Dark Mode, Just App Mode


lavavex:
**Describe the bug**
When enabling the "dark theme" It only applies to apps, and not to windows.

**To Reproduce**
1. Go to 'Tweaks'
2. Click on 'Enable' under Dark Mode

**Expected behavior**
This should set App AND System to Dark mode, and not just Apps

**Screenshots**
![Screen Shot 2022-11-04 at 4 28 14 PM](https://user-images.githubusercontent.com/27239435/200077598-50666182-016d-4e2d-a2ca-a6480820862f.png)

**Additional context**
I have sent a pull request #358 to fix this

lavavex:
Fix Merged into Test with Pull Req #358. Issue being closed

PR #358, opened 2022-11-04
Add Toggle for System Dark Mode

lavavex:
Adds toggle for System dark mode as well as App Dark mode to Tweaks. This is to fix issue https://github.com/ChrisTitusTech/winutil/issues/354 . Tested and Working on Windows 10 22H2. Was unable to figure out how to add the functionality to the current buttons so I made new buttons for system and renamed the current buttons to specify App dark mode. Feel free deny merge and do it better if you want as i have no clue how to use github properly and this is my first pull req. Fixed #357 (based on wrong branch sorry)


ChrisTitusTech:
I think most people would prefer these to just be the same click. 
Having a system and app dark mode separate seems like it would cause too many questions

lavavex:
> I think most people would prefer these to just be the same click. Having a system and app dark mode separate seems like it would cause too many questions

I have updated the code and have combined the buttons. There should be no change to MainWindow.xaml and only change to winutil.ps1 now.

ChrisTitusTech:
Looks great, Thanks @lavavex 
merging now.

PR #362, opened 2022-11-05
Add Disable and Enable Mouse Acceleration Tweaks

sidibu:
Adds tweaks to toggle mouse acceleration.


ChrisTitusTech:
Looks good, very useful when using Windows in a VM as well.

Issue #366, opened 2022-11-08
Ultimate Performance Addition


ChrisTitusTech:
Can you add an "Ultimate Performance" plan to tweaks?

powercfg -duplicatescheme e9a42b02-d5df-448d-aa00-03f14749eb61

PR #367, opened 2022-11-08
Add Ultimate Performance add/remove buttons

lavavex:
in response to issue #366 , added buttons for adding and removing the Ultimate Performance Power Plan.
needs testing on more than just windows 10 22h2.


lavavex:
If someone can do this better, be my guest, I bet there is a better way that probably looks nicer too and does checks to see if it already exists. (But I'm sadly not sure how to implement that)

Bhooty:
> If someone can do this better, be my guest, I bet there is a better way that probably looks nicer too and does checks to see if it already exists. (But I'm sadly not sure how to implement that)

not a pro person regarding these scripts but can you explain if this is better or just using desktop button in tweaks is better?

PR #372, opened 2022-11-11
PScriptAnalyzer recomendadtions

solomoncyj:


ChrisTitusTech:
Old script version and commit to main. 

Also, optimization tools and formatting fixes like this are hard to track and often get shot down during the PR stage unless it only fixes subtle things. As is, it is very hard to track any changes even if it were submitted to test branch.

Issue #401, opened 2022-11-23
Tool installs Git instead of selected program(s)


teacher-svb:
**Describe the bug**

When I select VS Code, Github Desktop or Unity to be installed through the install section, Git is installed instead.

**To Reproduce**
1. Go to 'Install'
2. Click on the checkbox for 'Github Desktop'
3. Click on 'Start Install'
4. Git is installed instead
5. The tool reports that Github Desktop is installed, allthough it isn't

**Expected behavior**

The selected programs are installed.

**Screenshots**

![image](https://user-images.githubusercontent.com/39157063/203595956-bf177a26-e975-4ecc-afd1-8a7f0ece6409.png)

![image](https://user-images.githubusercontent.com/39157063/203596172-e1437cd0-eb76-4475-b876-9819ba87d9a5.png)

![image](https://user-images.githubusercontent.com/39157063/203596230-40ac5869-868c-4ce6-a1f1-8df8ed491e7a.png)

**Additional context**

DeveloperDurp:
I know exactly what's wrong, hotfix should be available soon.

PR #405, opened 2022-11-24
Fix multiple winget entry in json file

DeveloperDurp:
I updated the logic to split the semi colon as this was missed in the original implementation of the Get-Checkboxes function. I also updated the logic to only pass unique values.

Also updated the unit tests to detect if duplicate values there.

#401


padsalatushal (approved):

padsalatushal (approved):

ChrisTitusTech (approved):
Looks good

Issue #430, opened 2022-11-26
Auto-generated ("scalable") WPF GUI


couleurm:
### **Is your feature request related to a problem?**
Having to manually modify the GUI to add something new (e.g a new app to the installer's section) when you could have have structured data, like one big hashtable with each value for an app being @{DisplayName = "official.winget.PackageName"}, it could also be equal to a hashtable which contains stuff like tooltips, package names..). It could also be used to store scriptblock that contains tweaks

Also throwing that out there: you could have sections for common apps for each known package managers (Chocolatey, Scoop, Winget) and it'd be much easier making something like this by doing the following:

### **Describe the solution you'd like**
I'm a WPF noob, so consider this with a grain of salt (I'd love helping out and learning more 👍)

The honest reason I'm making this issue is because I'm struggling with a clunky WPF function of mine I've frankensteined to make work (from farag2's Sophia Script). I've turned into a neat and user-friendly selector I can use within my scripts to prompt users, and helping you making it more useable (and maybe also useable outside of winutil for my usecase) would also help me:

It loops over each objects you pass to it (via `-Items`) in order to generate the GUI, it's logic may be interesting to implement into winutil to make it simpler to add new features

[Here's the code for `Invoke-Checkbox`](https://github.com/couleur-tweak-tips/TweakList/blob/master/helpers/Invoke-Checkbox.ps1) and a video demonstrating it:

https://user-images.githubusercontent.com/82747632/204089651-6bea07b5-fed4-470e-84b6-a10714d63f34.mp4

(didn't show it here but yes you can tick multiple checkboxes and it'll return them like you'd expect it to)

DeveloperDurp:
I 100% agree, only downside is in the end we would still want to load the GUI from a static file as there is a performance hit. I support a powershell GUI at my day job which dynamically builds the GUI and it can take several seconds to finish depending on how much it's doing. I am not WPF expert either but I like the idea, I think the best way would be to do something similar to how the helper script is setup. 

https://github.com/ChrisTitusTech/winutil/blob/f83ffaf0a9acd343dc7470d03d640fceade7450b/config/helperscript.ps1

couleurm:
If it's slow to run it could be compiled into a static state with CI, kinda like Hugo?

DeveloperDurp:
Yeah, that's kinda what I was eluding to. Only thing is we would want to keep in mind when people are developing locally IE #428 so additions should also be able to be ran manually.

[...]

Issue #428, opened 2022-11-26
Make winutil self contained, and dev/prod agnostic


couleurm:
**Is your feature request related to a problem?**
More of a nitpick from me / quality of life improvement for contributors

**Describe the solutions you're proposing**
Instead of having to comment and uncomment a line, why not leverage PowerShell and CI to either:

* "Compile" (combine to a single file) MainWindow.xaml and winutil.ps1 into one file to have it self contained (a single file to irm and worry about) and speed up startup time (one less request to make to github's servers, could also make it available offline easily (of course the app installer would be unusable))

Quick and dirty way I've wrapped my brain about making it work through CI, feel free to copy my [main.yaml](https://github.com/couleur-tweak-tips/TweakList/blob/master/.github/workflows/main.yml)
```PowerShell
Set-Content ./main.ps1 @"
`$inputXML = @'
$(Get-Content ./MainWindow.xaml -Raw)
'@
$(Get-Content ./winutil.ps1 -Raw)
"@
# If you're testing this out yourself don't forget to comment the current routine to not overwrite it
```
* Make use of `MyInvocation` to use `./MainWindow.xaml` if available, else make a request to GitHub to grab it:
```PowerShell
if (Test-Path "$PSScriptRoot/MainWindow.Xaml"){
	$inputXML = Get-Content "MainWindow.xaml"
	Write-Verbose "Using XAML from current directory" -Verbose
} else {
	$inputXML = (new-object Net.WebClient).DownloadString("https://raw.githubusercontent.com/ChrisTitusTech/winutil/$BranchToUse/MainWindow.xaml")
	Write-Verbose "Using latest XAML from GitHub"
}
# Presence of -Verbose on the first switch exclusivly on the first instance of Write-Verbose is intentional
```

**Additional info**
That may hurt your $10 EXE wrapper's sales, my apologies if I'm touching a sensitive subject

DeveloperDurp:
Something I tried implementing in the runspace version where you set an environment variable and it will load the files locally. 
https://github.com/ChrisTitusTech/winutil/blob/f83ffaf0a9acd343dc7470d03d640fceade7450b/runspace.ps1#L1362

Likely easily moveable to the main script. Currently there is a CI pipeline to modify the variable $BranchToUse so that main and test are more easily testable. Also I agree, files being pulled from gitlab is a quality of life improvement for using the script but sometimes can be a pain to develop for.

couleurm:
[...]

PR #435, opened 2022-11-26
Proof of concept of an auto generated script in CI

DeveloperDurp:
#430 
#428

Goals:

- [ ] Winutil be self contained and generated by either CI or a helper script
- [ ] xaml dynamically created through CI or a helper script


DeveloperDurp:
it works locally but not via invoke-restmethod lol. Like the idea though, should make unit testing individual files much easier.

PR #445, opened 2022-11-28
Replaced Dark Theme Enable and Disable buttons with a toggle switch

joshackland:
I thought it would look better if we changed the two dark mode buttons to a toggle switch.
When you open winutil, it checks to see what the system settings are and selects the correct toggle.

Before:
image

After:
image
image

To create this toggle switch, a style was created which can be applied to any checkbox. This can be re-used in the future if any additional toggle switches are needed.


ChrisTitusTech:
Oh man this is AWESOME! Thank you so much for the contribution.

lavavex:
Nice! I'm really new to coding stuff so the buttons were the best i could do, lol. glad someone was able to make it better

PR #498, opened 2022-12-06
Fixed If Chocolatey was not installed and the shell did not run as admin

widkit:
If Chocolatey was not installed and the shell did not run as admin, the script would return an error.  Just added a small privilege check.


ChrisTitusTech (changes requested):
Wrong branch, several of the formatting issues have already been addressed on test branch.

PR #519, opened 2022-12-17
Update check for Chocolatey

jstok84:
Let user decide if he wants to install Chocolatey.


jstok84:
I need to apply to test not main, sorry

Issue #516, opened 2022-12-15
Windows 11 Taskbar Tooltip display on mouseover have zero latency


meky84:
**Describe the bug**

When I mouseover the taskbar (for example on the clock icon) the tooltip appears without latency.

**To Reproduce**

In 2 different installations of Windows 11 22H2 after using the "Tweaks" section with "Recommended selections: Desktop" plus some other changes (I attach screenshots with the selected options) the bug happens,

**Expected behavior**

Normal latency to show tooltips is 400 ms .

**Screenshots**
![1](https://user-images.githubusercontent.com/37713275/207742319-7f7e24fd-bf33-4ebd-8ab6-5fca4a1623d5.png)
![2](https://user-images.githubusercontent.com/37713275/207742338-2d6a64da-6419-4c37-8d2e-d93d761b1cec.png)
![3](https://user-images.githubusercontent.com/37713275/207742348-e48961a6-c254-46d3-9a7b-baafecafc194.png)

Additional context

After several searches I discovered that one of the tweaks changes the registry key "ComputerHKEY_CURRENT_USERControl PanelMouseMouseHoverTime" from "REG_SZ" to "REG_DWORD" bringing the latency of the tooltip popup to zero instead of the default 400 ms (I attach screenshot).

Blaqwolf:
+1

I deleted the specified reg key and created a new REG_SZ as detailed above and it fixed the error.

Win11 Pro 22H2 on a clean install. I actually went back to W10P because of this issue, didn't realize it was due to this script! LOL Very glad I can fix it!

Thanks Christitus

ChrisTitusTech:
Fixed in commit a1c7501b98d56c5b554f9cfe70cb5a0767385796

PR #568, opened 2023-01-10
Fix data type of MouseHoverTime

Starfiresg1:
Setting Computer\HKEY_CURRENT_USER\Control Panel\Mouse\MouseHoverTime needs to be REG_SZ instead of REG_DWORD as suggested by [meky84](https://github.com/meky84) in issue #516. With the wrong type the Taskbar and the Start Menu become hard to use because everything pops up instantly, probably because it's then treated as a 0 ms delay.
This fixes issue #516.


ChrisTitusTech:
Good catch!

frostbreeze:
thank you very much dude, I was having the same problem , and now I could finally solve it.

PR #576, opened 2023-01-15
Feature/tweakstofunctions

DeveloperDurp:


DeveloperDurp:
@ChrisTitusTech I noticed when updating the restore point tweak I got. Should we modify that key automatically to ensure we are able to set a restore point on every run? 

WARNING: A new system restore point cannot be created because one has already been created within the past 1440 minutes. The frequency of restore point creation can be changed by creating the DWORD value 'SystemRestorePointCreationFrequency' under the registry key 'HKLM\Software\Microsoft\Windows NT\CurrentVersion\SystemRestore'. The value of this registry key indicates the necessary time interval (in minutes) between two restore point creation. The default value is 1440 minutes (24 hours).

DeveloperDurp:
Everything migrated to the tweaks.json file

ChrisTitusTech:
Sweet I'm going to review it all this week.

PR #639, opened 2023-03-07
If you want to keep "Undo All"

DeveloperDurp:
if you look at the Get-WinUtilCheckBoxes function it will show you how I grab all the check boxes from the GUI. 

https://github.com/ChrisTitusTech/winutil/blob/test-03-2023/functions/private/Get-WinUtilCheckBoxes.ps1#L39

### DIFF
--- a/functions/public/Invoke-WPFundoall.ps1
+++ b/functions/public/Invoke-WPFundoall.ps1
@@ -12,11 +12,12 @@ function Invoke-WPFundoall {
         return
     }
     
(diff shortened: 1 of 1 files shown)


DeveloperDurp:
Also if you wanted to go down the path of having the tool export and import automatically you can use the same code in the import/export function. Then you would only need logic to keep the json file update to date what has been ran and undone. 

https://github.com/ChrisTitusTech/winutil/blob/test-03-2023/functions/public/Invoke-WPFImpex.ps1#L29-L36

PR #641, opened 2023-03-08
Fix DarkMode

DeveloperDurp:
Noticed darkmode was not toggling, looks like it doesn't get picked up by the following line as the dark mode toggle is considered a checkbox and not a button. Will have to think of a way to pull those in automatically. Easiest way might be using a naming convention to pull all checkboxes that use that style.

https://github.com/ChrisTitusTech/winutil/blob/a1c7501b98d56c5b554f9cfe70cb5a0767385796/scripts/main.ps1#L48

### DIFF
--- a/scripts/main.ps1
+++ b/scripts/main.ps1
@@ -53,6 +53,10 @@ foreach ($button in $buttons){
     })
 }
 
+$WPFToggleDarkMode.Add_Click({    
+  Invoke-WPFDarkMode -DarkMoveEnabled $(Get-WinUtilDarkMode)
+})
+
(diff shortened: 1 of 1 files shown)


ChrisTitusTech (approved):

martin-braun:
@DeveloperDurp @ChrisTitusTech For me this didn't work. I'm testing on an non-activated Windows, but this should work nonetheless. I had to modify the values in `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Themes\Personalize` to toggle it.

DeveloperDurp:
This only got merged into the test branch, it would need to be merged into the main branch for the fix to work

PR #684, opened 2023-04-08
adding the code from asheroto to install winget

mignochrono:
with Server versions of Windows, that come without winget , the script seems unable to install it and then refuses to install any programs. so by adding two lines of code (hopefully) you can include the installation of winget and then fulfill the requirement and let the program run and do all the cool things. to my limited understanding, the echo command should forward the accept button and run the two commands that are in the asheroto repo.


ChrisTitusTech (approved):
Thank you, this look like a much better implementation.

Issue #638, opened 2023-03-07
Xbox Controller Xbox Button won't open Game Bar


Tommy2tone1982:
**Describe the bug**
After running recommended tweaks even unchecking gamedvr the 
xbox button on xbox controller no longer will open xbox game bar. It will still open with windows+G but for some reason it is breaking the buttons link to open it from the controller
**To Reproduce**
1. Run Reccomended Tweaks
2. Click on Xbox Controller XBOX button
3. Nothing Happens

**Expected behavior**
Xbox Controller Button should open game bar or controller bar, works fine before tweaks, and after undoing
**Screenshots**

**Additional context**

dennyamarojr:
Please open powershell as admin and use the following command
Get-AppxPackage | Select Name, PackageFullName >"$env:userprofile\Desktop\Apps_List.txt"

After you get the file Apps_list.txt attach in the issue so we can check your apps, to see if Xbox Game Bar app is installed.

dreamsyntax:
Hijacking this as I've reproduced this issue on 3 different computers.
[Apps_List.txt](https://github.com/ChrisTitusTech/winutil/files/11237190/Apps_List.txt)

The app itself still exists, and can be launched in Start with "Xbox Game Bar" app manually.

![image](https://user-images.githubusercontent.com/14857235/232157082-91b84c56-c823-4fce-b571-e4470ceb8998.png)

I suspect this setting - which always turns itself off once you re-navigate to the page is the issue.

So likely the DVR disable is also preventing the hotkeys for gamebar, despite gamebar still being installed.

dreamsyntax:
I've confirmed this is the issue.

I reverted the following:

```
"WPFEssTweaksDVR": {
    "registry": [
      {
        "Path": "HKCU:\\System\\GameConfigStore",
        "Name": "GameDVR_FSEBehavior",
        "Value": "2",
        "OriginalValue": "1",
        "Type": "DWord"
      },
      {
        "Path": "HKCU:\\System\\GameConfigStore",
        "Name": "GameDVR_Enabled",
        "Value": "0",
        "OriginalValue": "1",
        "Type": "DWord"
      },
      {
        "Path": "HKCU:\\System\\GameConfigStore",
        "Name": "GameDVR_DXGIHonorFSEWindowsCompatible",
        "Value": "1",
        "OriginalValue": "0",
        "Type": "DWord"
      },
      {
        "Path": "HKCU:\\System\\GameConfigStore",
[...]

Issue #595, opened 2023-01-31
I want to undo everything for it


Birdley:
I tried it to see if i liked it and knew it would let me do a restore point but i simply do not fully like it since it causes annoying issues like the xbox game bar to break and full screen games to always  go black for 3 seconds and is very annoying and i want to uninstall the entire thing because it doesn't fix when i revert or undo it

Issue #371, opened 2022-11-10
Game Bar


Cameron6897:
**Describe the bug**
Running the desktop tweaks disables game bar, not allowing me to press Windows + G to open it. I have used this tool for the tweaks and many fresh windows installs and game bar has always worked after. For some reason now it doesn't. I tried reinstalling windows 10 home, game bar works before running the tweaks. It does not work after running the tweaks. I can open it from the search bar but not Windows + G. I go to enable it in windows settings but everytime I come back, it is disabled. I tried changing registry and group policy values. I tried to use the undo all tweaks button, but that did not work either.

thamudi:
I am facing the same issue, going through the script shows that it will be removed by default if [WEPFssDebloat is enabled](https://github.com/ChrisTitusTech/winutil/blob/main/winutil.ps1#L871)

I am trying to use [this](https://answers.microsoft.com/en-us/insider/forum/all/help-i-lost-my-game-bar/95dadd76-227b-4ed6-aaef-24a78fab6e1d) solutio. Will update if it works.

ChrisTitusTech:
Fixed in #692

PR #692, opened 2023-04-14
Split up AllowGameDVR to fix breaking Xbox Game Bar

dreamsyntax:
The `AllowGameDVR` regkey not only prevents GameDVR, but it also stops the hotkeys for Xbox Game Bar.
Winkey+G and Xbox Controller -> Xbox Button will not work when this key is set to 0.

I propose splitting this key into its own category. By keeping this separate, a user who wants it completely off can still enable this key.

Resolves:
#638
#371
#595

Worth noting to get GameBar working again on machines that already ran the Disable DVR script must:

1. Change the AllowGameDVR regkey to 1:  
`Set-ItemProperty -Path "HKLM:\SOFTWARE\Policies\Microsoft\Windows\GameDVR" -Name "AllowGameDVR" -Type DWord -Value 1`
2. Reboot PC
3. Explicitly Toggle ON the option in Settings -> Gaming -> Xbox Game Bar

--------

This moves the AllowGameDVR reg edit to its own category. 
There is plenty of use cases for Xbox Game Bar (e.x. using as a hotkey for Sound Mixer), where the user would otherwise not care about GameDVR. Most users probably do not use the recording feature but may use Xbox Game Bar for audio management, performance overlay, etc...

Splitting this option apart allows the user to still disable the GameDVR portion while keeping Xbox Game Bar itself functional for other uses.

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -1268,7 +1268,11 @@
         "Value": "0",
         "OriginalValue": "1",
         "Type": "DWord"
-      },
+      }
+    ]
+  },
+  "WPFDisableGameBar": {
+    "registry": [
       {
         "Path": "HKLM:\\SOFTWARE\\Policies\\Microsoft\\Windows\\GameDVR",
         "Name": "AllowGameDVR",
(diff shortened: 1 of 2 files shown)


ChrisTitusTech (changes requested):
Tweaks.json looks good, but winutil.ps1 is the auto generated compiled file and the changes will need to come from one of the modules. 

Also you need to submit this to test branch and not the main branch.

ChrisTitusTech (approved):

PR #708, opened 2023-05-01
Added Guilded

EdiValentin101:
Guilded is a Discord competitor that is primarily focused on video gaming communities, such as those focused on competitive gaming and esports. It provides features intended for video gaming clans, such as scheduling tools and integrated calendars.
Guilded's UI is almost identical to Discord, which makes it easier for users to switch. When it comes down to it, Guilded is actually very different from Discord because it offers more for free.

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -55,6 +55,10 @@
     "winget": "Discord.Discord",
     "choco": "discord"
   },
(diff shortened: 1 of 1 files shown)


ChrisTitusTech:
No commits to main

PR #713, opened 2023-05-03
Test 03 2023

iBeny:
Hi Titus,

Hope this is the fastest way to catch your attention about a "fundamental" aspect of your OTFE app you haven't discussed in your videos : is your app language aware or english only (not in the UI texts but in its actions(inter-) with the OS) ? If yes, then mark ALL language specific tweak/features of OTFE with an asterisx so we don't wreck our OS !

Are the program installations you implement language aware or english only ? Please specify in your OTFE GUI.

Best regards
Benoit


iBeny:
Hi Chris,

The follow-up/response to my question/concern is as obscure as can be. This must be a secret society cabale with secret languages or so. Can you make you answer understandable to the non... initiated to your secret society please ?

Best regards,
Benoit

PR #726, opened 2023-05-09
Test 03 2023

ChrisTitusTech:


Alilo1P:
good util

amine-051:
good i advise

Nikoskampourakis:
Can you add a feature in microwin that you can add exes and it makes a app in your custom iso

PR #773, opened 2023-05-19
add verbose logon

padsalatushal:
Enable or disable verbose sign in status messages. Verbose status messages may be helpful when you are troubleshooting slow startup, shutdown, logon, or logoff behavior.

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2238,6 +2238,28 @@
       }
     ]
   },
+  "WPFMiscTweaksEnableVerboselogon": {
+    "registry": [
+      {
+        "path": "HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\policies\\system",
+        "OriginalValue": "0",
+        "name": "VerboseStatus",
+        "value": "1",
+        "type": "DWord"
+      }
+    ]
+  },
+  "WPFMiscTweaksDisableVerboselogon": {
+    "registry": [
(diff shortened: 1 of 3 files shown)


padsalatushal:
Source : https://learn.microsoft.com/en-us/troubleshoot/windows-server/performance/enable-verbose-startup-shutdown-logon-logoff-status-messages

DeveloperDurp:
@padsalatushal We can compress this into one checkbox, especially if it's that one registry value. With the undo function users would be able to enable and disable this without the need of 2 checkboxes

padsalatushal:
@DeveloperDurp thanks for the suggestion. I will make changes.

padsalatushal:
can we do same with disable power throttling and disable mouse acceleration feature.?

DeveloperDurp:
Yes, those can be converted to a single checkbox as well

padsalatushal:
one more question :  
is it a good idea to turn the checkbox button to a toggle button like 
image

DeveloperDurp:
We could, however dark mode is a little special. It has logic that when it's "checked" or "unchecked" it will execute a function. There is also a function to get the state when winutil starts. I have ideas about handling this but because that is also a checkbox you can't make a blanket function like I do for every button, otherwise every checkbox would start executing code every time you check or uncheck a box.

ChrisTitusTech (approved):

PR #796, opened 2023-05-23
Update Set-WinUtilDNS.ps1

iDereje:
Removed the unnecessary Out-String command when displaying the list of interfaces.
Replaced the Out-String with Format-Table for better readability.
Replaced $psitem with $Error[0] to access the caught exception.
Removed the comment block inside the function, as it wasn't affecting the code's functionality.

**Update the prompt

### DIFF
--- a/functions/private/Set-WinUtilDNS.ps1
+++ b/functions/private/Set-WinUtilDNS.ps1
@@ -1,32 +1,33 @@
 function Set-WinUtilDNS {
     <#
-    
         .DESCRIPTION
-        This function will set the DNS of all interfaces that are in the "Up" state. It will lookup the values from the DNS.Json file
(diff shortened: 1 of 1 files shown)


DeveloperDurp:
Please make sure you switch this to the latest test branch.

out-string is required otherwise powershell would display the object type instead of the contents. I also tend to avoid format-table as it has the chance of cutting information off. 

Is there much difference between $psitem and $error[0]?

DeveloperDurp:
There is no output in this current version
image

You need to change your $adapters | format-table to include write-host in front of it which gives you this output
image

for reference this is the original
image

ChrisTitusTech (commented):
Wrong branch, and not sure if this would be an improvement over the current method.

PR #800, opened 2023-05-26
Update applications.json

0scvr:
Add [Malwarebytes Windows Firewall Control](https://www.binisoft.org/wfc.php) in applications install list


ChrisTitusTech:
Not a fan of this package, not going to merge and it's on the wrong branch.

PR #852, opened 2023-06-20
UI theme support

KonTy:
Adding theme support so people with bad vision could have better contrasting themes potentially

Later when this is tested well, we can do one of 2 things.
1. Pass theme on a command line so people with bad vision will have an option to have a better (more visible theme)
2. Create a drop down with themes and a save button which would save it ti some local config file


ChrisTitusTech:
Closing because it was submitted to main branch

PR #854, opened 2023-06-23
Update vcredit package names for winget winutil.ps1

shljsl75891:
Microsoft Visual C++ 2015-2022 Redistributable (32bit, 64bit) package names for winget were incorrect. Due to which, these redistributables were not installing after selection. I have changed them with right names. Please check.... BTW this is a great tool. Keep it up, Titus sir.


shljsl75891:
I will pull it in test branch

Issue #544, opened 2022-12-28
Program Update: Replace Origin with EA Play


Scallywer:
**Describe the bug**
Origin is now legacy and EA Play is to be used

**To Reproduce**
None

**Expected behavior**
Self explanatory

**Screenshots**

**Additional context**

ChrisTitusTech:
Thanks for the issue, I'll get this swapped out.

PR #865, opened 2023-07-06
Replace Origin with EA App, add DDU

supplefrog:
#544 #866

### DIFF
--- a/winutil.ps1
+++ b/winutil.ps1
@@ -2761,11 +2761,11 @@ $sync.configs.applications = '{
     "choco": "dotnet-6.0-runtime"
   },
   "WPFInstallvc2015_64": {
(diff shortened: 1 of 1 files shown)


winutil.ps1
```diff
@@ -2761,11 +2766,11 @@ $sync.configs.applications = '{
     "choco": "dotnet-6.0-runtime"
   },
   "WPFInstallvc2015_64": {
-    "winget": "Microsoft.VC++2015-2022Redist-x64",
+    "winget": "Microsoft.VCRedist.2015+.x64",
```
ChrisTitusTech, line 2769: Missing commit from prior commit

supplefrog: Fixed in new pr

PR #900, opened 2023-07-21
Removes ugly blue text when script is running

b421337:
Adding this to all ps1 scripts isn't necessary because not all scripts create the blue text. But, noticed in your live stream that upon first launch of winutil, when it checks/installs winget, the blue texts appears on powershell window. This change is purely aesthetics.


ChrisTitusTech:
This is about to revamped and a different install method will be used.

PR #908, opened 2023-07-25
Added Bing Search toggle plus some support functions

DeveloperDurp:
Added a framework to make toggle switches like dark mode easier. As long as the checkbox is called "WPFToggle" it will automatically be given the Invoke-WPFToggle function (similar function to Invoke-Button) and automatically checked if the status is enabled or disabled.


ChrisTitusTech (approved):

ChrisTitusTech:
I like the addition of the toggle. I might start coding these type of function into other parts of the script as well.

DeveloperDurp:
That's the idea, hopefully this will make it easier to migrate over to using toggle switches in the future. I will try to think of a way to use the config files as well but so far don't see a good enough pattern.

I would be careful of what you convert to a toggle switch, right now it performs a check on load so if you have a complex toggle it will impact the initial load times.

PR #910, opened 2023-07-26
Update applications.json

dejjavu:


ChrisTitusTech (commented):
Not a fan of ccleaner
Pretty much malware at this point and I don't recommend anyone install it.

PR #918, opened 2023-07-28
Updated Edge Removal Script to the more up-to-date one.

antunnitraj:
Github user "padsalatushal" forked the script from the original owner "AveYo" and "padsalatushal" did not update it for almost two months. You can see that its the same script because "padsalatushal" referenced the original owner in the README.md in the repo [here](https://github.com/padsalatushal/Edge-Removal#resources). Here is a more recent script with my modifications to the "extra cleanup" part.

padsalatushal's repo: https://github.com/padsalatushal/Edge-Removal
AveYo's repo: https://github.com/AveYo/fox

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2178,7 +2178,7 @@
   "WPFEssTweaksRemoveEdge": {
     "InvokeScript": [
         "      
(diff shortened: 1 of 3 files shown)


ChrisTitusTech (approved):
Thanks for the update

PR #949, opened 2023-08-08
Add uninstall edgehtml tweak

ChaseKnowlden:
Helps get rid of the white tile on the start menu.


ChrisTitusTech (changes requested):
Can't replace an open source cmd file with a closed binary.

PR #953, opened 2023-08-09
Add Message confirming restore point created

hypertux001:


ChrisTitusTech (approved):
Good idea, I overlooked this

PR #986, opened 2023-08-26
Make irm command even shorter based on issue 144

crazycheetah42:


ChrisTitusTech:
It actually used to be like this, but I changed it to https:// as people were getting SSL errors. This solves it in some instances.

Issue #926, opened 2023-08-01
Translucent Taskbar - No package found matching input criteria


influential-eliot:
# Description
When trying to install 'Translucent Taskbar' Winget gives the error message:
``` No package found matching input criteria ```

Would assume that this is low priority. 👍 

## Reproduction steps
1. Check only '_Translucent Taskbar_'
2. Tap '_Install Selection_'

# Expected behavior
The installer grabs the package and installs it.

# Screenshots

![ChrisTitusTech-winutil--Translucent-Taskbar--Does-Not-Find--001](https://github.com/ChrisTitusTech/winutil/assets/112952958/daaad290-8e56-412e-a130-645e5aee0a1d)

# Additional context

- **OS**: Windows 11 Pro
- **Version**: 22H2
- **Build**: 22621.2070 ni_release
- **Experience**: Windows Feature Experience Pack 1000.22659.1000.0
- **Status**: Queued for unenrollment from preview builds
- **Note 1**: As per #890 Winget is updated
- **Note 2**: FWIW, I've not been able to get 'Security (Recommended)Settings' to stick and only push sec updates. I assume that's because of the darned preview builds. Thought I was being clever there, when I CLEARLY wasn't. 😩

Velosofy:
TranslucentTB has to be installed using choco.

It has been removed from winget repo since 2022 https://github.com/microsoft/winget-pkgs/commit/23a8fca370f9119bd9af1b826331d037222c5e03

influential-eliot:
@Velosofy interesting ... because choco is installed. Thanks, though!

Velosofy:
The toolbox doesn't have a failover to choco, so it failed. You have to do it yourself.

influential-eliot:
@Velosofy , aha ... I see ... sorry, I wasn't aware that this was a thing that you had to do when tapping an install. When it starts up it checks for (_and finds_) choco, so I assumed that was fine.

Not saying any of that sarcastically, I assumed that it would just install, is all.

So in ITIL terms:

1. This was an **incident** (_tt doesn't install_)
        - Closable with a **workaround** (_install manually_)
2. Inferring a **problem** be raised (_tt installs via wrong medium_) 
        - Requiring a **fix** (_point script at choco_) at a later date

In normal people terms:

**Close this, add a task to switch the script.** 😅 

((( _obvious jovial tone is hopefully obvious, not speaking for anyone i've no skin in the game_ )))

PR #1001, opened 2023-09-06
fix package name

### LINKED ISSUE #997: Cant install Translucent Taskbar
Cant install Translucent Taskbar on the install tab.
Screenshot 2023-09-01 205721

padsalatushal:
fix #997 #926


ChrisTitusTech (approved):
Strange winget does funny things like this from time to time. The package name used to work, but you are right looks like they broke it.

PR #1010, opened 2023-09-09
Added tweaks for IPv6

Joanty24:
Added tweaks to disable and enable IPv6.

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2402,5 +2402,21 @@
         "Value": "0"
       }
     ]
+  },
+  "WPFMiscTweaksDisableipsix": {
+    "InvokeScript": [
+      "Disable-NetAdapterBinding -Name '*' -ComponentID ms_tcpip6"
+    ],
+    "UndoScript": [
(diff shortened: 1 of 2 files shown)


ChrisTitusTech (approved):

ChrisTitusTech:
I might actually change this to a different section like config instead of tweaks. I'll probably push DNS over there too. It makes more sense in the other tab. I'm about to revamp the tweaks/config section to be more user friendly.

Joanty24:
If you want I can move it to the config tab but I'd keep it as it is until the GUI update.  
I'll do what you thing is best.

ChrisTitusTech:
> If you want I can move it to the config tab but I'd keep it as it is until the GUI update. I'll do what you thing is best.

I'm just thinking of the most user friendly way of approaching it. I am going to remove all the "Disable/Enable" Tweaks and make those toggles. Ipv6 will be one of those. You can see Dark Mode as a reference.

ChrisTitusTech (approved):

ChrisTitusTech:
Going to just merge it so the code is there and I'll move it later.

PR #1027, opened 2023-09-14
Update edgeremoval.bat

crian:


crian:
@ChrisTitusTech it would be easier to maintain the edgeremoval.bat updates if the script gets pulled directly from source.

ChrisTitusTech (approved):

PR #1028, opened 2023-09-14
Pull edgeremoval.bat directly from source

crian:


crian:
The script can now also be used to restore everything.
So with the appropriate arguments edge, webview and xsocial can be reinstalled.

ChrisTitusTech (approved):

ChrisTitusTech:
Yeah Aveyo has done a great job with it.

PR #1041, opened 2023-09-20
Add Discord PTB

elli0t43:
Added discord PTB (for the people who use PTB instead of stable.)


ChrisTitusTech:
Don't want to bloat up the utilities section with dev builds. It would triple the list size.

Issue #1052, opened 2023-09-24
Add ZeroTier One to Utilities install menu options


tsdevau:
I would love to see ZeroTier One included as an install option in Utilities.  It is an easy to configure VPN utility with a generous free tier.  I use it on most client machines to enable secure VPN access to the client for remote maintenance and/or access.

I appreciate that you need to consider additions carefully, so as not to bloat the utility, but I think this tool really fits with what you're trying to achieve.

If you're not familiar with the utility here is a link to their [website](https://www.zerotier.com) and the [Github repo](https://github.com/zerotier/ZeroTierOne).

tsdevau:
PR #1054 resolves this issue.

PR #1053, opened 2023-09-24
Adds ZeroTier One for #1052

tsdevau:
- Adds install option for ZeroTier One to Utilities menu.

ZeroTier One is a popular, easy to configure VPN utility with an extremely generous free tier.


tsdevau:
Made PR to main instead of Test branch as preferred.  Deleting this PR to replace it with PR on Test branch.

PR #1054, opened 2023-09-24
Adds ZeroTier One to utilities menu #1052

tsdevau:
Adds install option for ZeroTier One to Utilities menu.  Refer to [Issue #1052](https://github.com/ChrisTitusTech/winutil/issues/1052).

ZeroTier One is a popular, easy to configure VPN utility with an extremely generous free tier.

Not sure why PR went through with so many line changes, but all that actually changed is the addition of ZeroTier One in both files.

Compiles and runs & installs OK on my tests.

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -143,6 +143,10 @@
 		"winget": "GIMP.GIMP",
 		"choco": "gimp"
 	},
+	"WPFInstallguilded": {
+		"winget": "Guilded.Guilded",
+		"choco": ""
+	},
 	"WPFInstallgit": {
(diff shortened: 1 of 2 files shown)


DeveloperDurp:
> Not sure why PR went through with so many line changes, but all that actually changed is the addition of ZeroTier One in both files.

Looks like a difference in the amount of indents, I guess the existing file has a few too many.

ChrisTitusTech (changes requested):
Can't accept PRs with formatting changes. impossible to verify what has changed.

PR #1129, opened 2023-10-26
MicroWin and fixes to common bugs

KonTy:
This new feature allows you to create debloated ISO images of Windows 10 and 11. It removes many components that you don't need just like tiny11 and tiny10 do. Or scripts like MSMG and NTlite. At the end of the install you end up with about 85 process running, depending on how much you want to delete. Overtime, we can make it into a really cool seamless tool that is easy to use and very flexible. 

Some highlights from MicroWin

1. Remove Telemetry and Tracking
2. Remove TPM requirement
3. Add ability to use local accounts 
4. Remove Wifi requirement to finish install
5. Ability to remove Edge 
6. Ability to remove Defender 
7. Apps debloat
And more....

image

Some other features:
1. Fixed the Icon now it shows CTT logo vs PowerShell logo
2. You can clearly see which tab you are on
3. Ctrl-F now allows you to filter features for quick find
4. Fixed Admin issue with elevation and redownload
5. Added a scrollbar to the Apps screen https://github.com/ChrisTitusTech/winutil/issues/1130
6. Use Ctrl-I, C  T, U, M to jump between tabs
7. More ...


KonTy:
@ChrisTitusTech feel free to change the text and layout on the MicroWin tab I'm not good at writing instructions etc.

KonTy:
@ChrisTitusTech  there are lots of changes in this PR, I fixed tons of bugs and added a new feature. Right now it generates unattend.xml and FirstRun.ps1 that runs after the install is done, The problem is that it is static and it doesn't benefit from all the updates that you make to the winutil. I think the next thing that we should do it make winutil be able to run from a command line. Images if you could save all the config files with apps to install and tweaks and then run it like winutil.ps1 --run someconfig.js. If we had this ability, people would save thier configs and then inject them into an iso so that when setup finishes it would run all the tweaks and install all the apps. This should be relatively easy to do, I could probably finish it in November.

KonTy:
@ChrisTitusTech I just saw your stream when you looked at this PR. I think you misunderstood the purpose of MicroWin. I'm sorry I was not clean in the instructions above. This tool allows you to create tiny11 like ISO, very light weight Windows version without all the problems that Tiny11 has, you can modify any Windows ISO and you are not tied into a particular version of Windows ISO. You can also use it to create Win10 version etc. This will be our own way to create tiny images of windows, delete telemetry, defender and other things. Overtime this can grow into a very flexible tool to remove bloat and prep images.

This also doesn't suffer from oscdimg problems that other tools have because we dynamically download it from choco store.

PS this besically replaces ntlite/msmg/tiny10/tiny11 and other scripts and tools out there that allow you to create debloated Windows.

ChrisTitusTech (commented):
There are several PRs and commits since this was first pushed. Can you pull the changes and resend the PR on the new test branch for November?

KonTy:
Oh it will be extremely hard now to split it into different PRs because some of the features I've added depend of the fixes and some fixes depend of features that I've added. :-)

ChrisTitusTech:
Yeah I love the addition, but was hard to manage with all the other PRs that came in.

KonTy:
Ye totally, I will figure it out man, give me a week or so, just busy with work atm. I will create a PR this weekend.

ChrisTitusTech:
Thanks so much for this work @KonTy, I'll be watching the repo this week and try not to do any big PR's until I see yours come through. Since it's a bigger PR I'd like to do it separate instead of including it with a bunch of smaller PRs.

PR #1195, opened 2023-11-24
Create wrapper.cmd

Build-0-Matic:
This is an easy to use wrapper for downloading and running the Chris Titus Tech Debloater.
It's written as a simple to use Batch file. It verifies that all the basic requirements are met before invoking the powershell command.


ChrisTitusTech:
We already have the desktop shortcut to powershell to launch it. The cmd file is a bit overkill.

Build-0-Matic:
That's OK with me as I did it mostly for fun.
Cheers!

Issue #1216, opened 2023-11-30
netlogon service will be set to manual but it's needed by windows time sync service (in domain environment)


popy2k14:
**Describe the bug**
netlogon service will be set to manual but it's needed by windows time sync service (in domain environment)

**To Reproduce**
1. run tweaks (netlogon is set to manual)
2. Reboot
3. In windows settings under "Date/Time" you can see a source "CMOS clock" for timesync
4. time sync from DC is not working

**Expected behavior**
Not disable netlogon in domain environments

**Screenshots**

**Additional context**
It would be nice if there would be an detection, if the computer is joined to a domain and skip some parts of the tweaks.

PS.: i know that normally you do such tweaks with GPO on the DC, but hey i have local admin rights and want to tweak my PC like i would :-)

lavavex:
Made a commit request to just remove netlogon from being touched by script, as according the script, it should already be manual on clean systems, so it seems redundant to touch it again especially if it may cause an issue...

popy2k14:
In my case it was "manual" and not started, when needed by windows time sync service!
I think it was set to manual by the script (on other machines, where the script was not executed, it is on "auto").
After setting it manually to "auto" and started it manually, windows time sync service worked again, against the DC.

So yes, dont touch "netlogon" would be nice.

lavavex:
> In my case it was "manual" and not started, when needed by windows time sync service! I think it was set to manual by the script (on other machines, where the script was not executed, it is on "auto"). After setting it manually to "auto" and started it manually, windows time sync service worked again, against the DC.
> 
> So yes, dont touch "netlogon" would be nice.
> 
> PS.: thx fot the pull request.

all good, it was something i could do with no knowledge of how to code... now to hope i dont get flamed for pulling to main as test branch is non existant

munkk01:
> > In my case it was "manual" and not started, when needed by windows time sync service! I think it was set to manual by the script (on other machines, where the script was not executed, it is on "auto"). After setting it manually to "auto" and started it manually, windows time sync service worked again, against the DC.
> > So yes, dont touch "netlogon" would be nice.
> > PS.: thx fot the pull request.
> 
> all good, it was something i could do with no knowledge of how to code... now to hope i dont get flamed for pulling to main as test branch is non existant

That's what I did so we're in the same boat :)

lavavex:
Netlogon has been changed to be set to Automatic in new test branch per PR #1220 being merged

PR #1220, opened 2023-11-30
Dont touch netlogon?

lavavex:
To Address issue #1216 

- Remove setting the netlogon service to manual as it interferes with domain joined systems needing time sync

- As the "Original" is also set to manual in the script, this seems somewhat redundant as domain joined systems should set this to automatic, where as it should already be manual on non domain joined systems.

- If netlogon is actually originally automatic on non domain joined systems, then ignore commit and we can find a way to pull AD status or something before running.

### DIFF
--- a/winutil.ps1
+++ b/winutil.ps1
@@ -6150,7 +6150,7 @@ $sync.configs.tweaks = '{
       },
       {
         "Name": "OneSyncSvc_dc2a4",
-        "StartupType": "matic",
+        "StartupType": "Automatic",
         "OriginalType": "Automatic"
       },
       {


lavavex:
also, test branch is missing, thats why i pushed to main, dont think i cant read please

ChrisTitusTech (changes requested):
Might just move this to Auto just because on non-domain joined systems I believe netlogon is a dependency of group policy.

lavavex:
Changes made to set netlogon to automatic per request

ChrisTitusTech (approved):

PR #1223, opened 2023-12-01
Fixed Microwin Mount & Unmount, Reformatted

NovaWilliamClark:
After mounting drive, the value of the mounted drive letter was valued null - I Have replace this with a txt prompt asking for the drive letter where it mounted.

Also fixed the unmounting of the ISO after copying files to the microwin folder in C:\

Ran through default Powershell formatter in VS Code

### DIFF
--- a/winutil.ps1
+++ b/winutil.ps1
@@ -28,23 +28,21 @@ $sync.ProcessRunning = $false
 
 $currentPid = [System.Security.Principal.WindowsIdentity]::GetCurrent()
 $principal = new-object System.Security.Principal.WindowsPrincipal($currentPid)
-$adminRole=[System.Security.Principal.WindowsBuiltInRole]::Administrator
+$adminRole = [System.Security.Principal.WindowsBuiltInRole]::Administrator
 
-if ($principal.IsInRole($adminRole))
-{
-    $Host.UI.RawUI.WindowTitle = $myInvocation.MyCommand.Definition + "(Admin)"
-    clear-host
+if ($principal.IsInRole($adminRole)) {
+  $Host.UI.RawUI.WindowTitle = $myInvocation.MyCommand.Definition + "(Admin)"
+  clear-host
 }
-else
-{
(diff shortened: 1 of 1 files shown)


NovaWilliamClark:
> Hey @NovaWilliamClark first of all thank you for getting engaged, because of people like you projects like this thrive. A couple of things, please take a look at a projects as a whole, winutil is a compiled script, it contains many scripts, so you can't directly "fix" winutil you have to fix corresponding functions that get compiled into winutil. Another thing is instead of presenting people with a dialog for choosing the drive (fixing the symptom) it would be better to track down the issue and figure out why it is failing and fix it. Please take a look at Invoke-WPFGetIso.ps1 in the public folder. This line suppose to mount the iso and return it's object, if that is not happening something is going on
> 
> ```powershell
> $mountedISO = Mount-DiskImage -PassThru "$filePath"
> ```
> 
> I will also try to repro this on my machines, so far I was unsuccessful.

Thank you for looking into it with me, turns out I suspect a windows service or something regarding the `Mount-DiskImage`  & `Get-Volume` functionality might have not been active or was corrupt on my machine. After a fresh Windows install, it the Mount-DiskImage is now returning a value and seems to be mounting an dismounting properly. Not entirely sure why but there we go!

KonTy:
@NovaWilliamClark I still think it is a legit issue, if someone doesn't have a couple of services that are not running, which is a huge probability in this community, we should catch that issue and show a warning and tell people what to do and which services to start. I would still encourage you to keep looking to see how we can find the issue and show a warning to the user. Or if you could tell me how to repro this issue I could look into it.

NovaWilliamClark:
> @NovaWilliamClark I still think it is a legit issue, if someone doesn't have a couple of services that are not running, which is a huge probability in this community, we should catch that issue and show a warning and tell people what to do and which services to start. I would still encourage you to keep looking to see how we can find the issue and show a warning to the user. Or if you could tell me how to repro this issue I could look into it.

I'll continue to look into it and see if I can find what might have been causing the issue

PR #1254, opened 2023-12-05
Improve Runtime installer of Winutil

SteavenGamerYT:
Improve Runtime installer of Winutil

### DIFF
--- a/winutil.ps1
+++ b/winutil.ps1
@@ -371,33 +371,29 @@ function Get-WinUtilVariables {
     return $keys
 }
 function Install-WinUtilChoco {
-
-    <#
-
-    .SYNOPSIS
-        Installs Chocolatey if it is not already installed
-
-    #>
-
-    try {
-        Write-Host "Checking if Chocolatey is Installed..."
-
-        if((Test-WinUtilPackageManager -choco)){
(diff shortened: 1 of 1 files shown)


munkk01:
Just to confirm, this modification is directly for `winutil.ps1`, which will get overwritten when compiled. Correct?

SteavenGamerYT:
its not really for Install-WinUtilChoco.ps1, chris said he will only use the 

```sh
function Install-WindowsUpdateCli {
    # Store the current ConfirmPreference
    $currentConfirmPreference = $ConfirmPreference

    # Set the ConfirmPreference to "Yes" only for this function
    $ConfirmPreference = 'Yes'

    # Check if PSWindowsUpdate module is installed
    if (-not (Get-Module -Name PSWindowsUpdate -ListAvailable)) {
        Install-Module -Name PSWindowsUpdate -Force -Confirm:$false
    }

    # Check if WUServiceManager has Microsoft Update configured
    $wuService = Get-WUServiceManager
    if (-not ($wuService.MicrosoftUpdate -eq $true)) {
        Add-WUServiceManager -MicrosoftUpdate -Confirm:$false
    }

    # Restore the original ConfirmPreference
    $ConfirmPreference = $currentConfirmPreference
}```

PR #1275, opened 2023-12-09
[MicroWin] Get-LocalizedYesNo does not work correctly for some localizations

beehunt9r:
I tried to use MicroWin to create a compact image, however, I encountered the following error.

Invoke-WPFMicrowin

I had to do some research, which revealed the problem that on my localization (ru) the help has only 2 spaces instead of 3.

takeown.exe

This caused an error that did not allow to return values for Yes and No correctly. After replacing the fixed number of spaces with a regular expression, everything worked correctly.  I can assume that this problem occurs not only on my localization.

### DIFF
--- a/functions/private/Get-LocalizedYesNo.ps1
+++ b/functions/private/Get-LocalizedYesNo.ps1
@@ -30,7 +30,7 @@ function Get-LocalizedYesNo {
                 break
             }    
         }
-        elseif ($line -match "/D   ") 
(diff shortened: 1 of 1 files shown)


KonTy:
@beehunt9r great find, thank you for the fix. @ChrisTitusTech I tested it on English it seem to work fine.

ChrisTitusTech (approved):

PR #1291, opened 2023-12-17
Also inject drivers into boot.wim

0SkillAllLuck:
For touch screen devices and devices with special storage having the drivers in boot.wim can be really helpful. Good example are handheld devices and Intel RST/VMD devices.


ChrisTitusTech:
copied to another branch

PR #1305, opened 2023-12-21
Fix grammar error in microwin

xemulat:
Change oscdimge.exe to oscdimg.exe


KonTy:
Thanks for doing this, you might want to also fix another typo while you are at it, the 'do' should probably be a 'to'. :)

"now attempt *do* download and install it using choco or github. This might take a long time."

ChrisTitusTech (approved):

ChrisTitusTech (changes requested):
This is to the compiled file and will be overwritten. The correction should happen in the functions folder.

PR #1348, opened 2024-01-04
Increase performance during loading.

KonTy:
Add a clear button to the search box.
Add link and description attributes to the applications JSON. Use the link for linking to the app website.
Use the description as a tooltip for each app.
Add a clickable link to the website for each application (this took a long time; don't kick me if I got some wrong). Pressing Escape now clears the filter box.
Pressing Alt-P prints your PID.


ChrisTitusTech (approved):
Looks great, I need to create a local wiki here and redo that link in the future. However, that is work for future me... until then we can use your wiki page hehe.

PR #1360, opened 2024-01-09
Add a few apps

AlbydST:
Added FireFox Developer Edition, FireFox ESR, Qute Browser, SuperTux, SuperTuxKart, RipGrep, Kali Linux for WSL and GNU Icecat and did some minor changes

PS:  I don't know why it doesn't show the changes, but they are there. You might want to check my fork.


ChrisTitusTech:
pull and submit to test branch

KonTy:
@ChrisTitusTech imho the list of apps is getting very big and unruly, maybe it would make sense to create some kind of voting mechanism??? As I was working on the "links" feature I had to go through each and every app, what I saw -- there are so many one-off apps that majority people will never use. dunno man. Maybe we should split it into several pages, or create treeview like UI? Not sure...
We could also introduce some telemetry to gather which apps and installed and which once are not, and start removing apps that are not used. As it is now, the list is getting huge and UI will inevitably slow down on "slower" machines after all this is still a script and not a compiled app :)

Issue #1367, opened 2024-01-10
Feature Request: Restore Point Functionality


saharankamal:
I would like to suggest the implementation of a restore point feature in your developed by Chris Titus. This feature would allow users to revert any changes made by the tool to a previous state, providing a safety net in case something goes wrong.

Benefits
- Adds an extra layer of security for users experimenting with the tool.
- Enables quick and easy rollback in case of unintended changes.

munkk01:
I didn't realize this was already in there. Ignore the closed PR.

munkk01:
This was addressed and merged. Can be closed.

PR #1369, opened 2024-01-10
add restore point function

munkk01:
I'm sure there is much more elegant solution than `Out-GridView` but I couldn't seem to get my tiny brain to figure something else out. 

Feel free to not use this and perhaps take inspiration for a better method.

Created in reference to:
https://github.com/ChrisTitusTech/winutil/issues/1367#issue-2074952480

### DIFF
--- a/functions/public/Invoke-WPFButton.ps1
+++ b/functions/public/Invoke-WPFButton.ps1
@@ -36,6 +36,7 @@ function Invoke-WPFButton {
         "WPFundoall" {Invoke-WPFundoall}
         "WPFFeatureInstall" {Invoke-WPFFeatureInstall}
         "WPFPanelDISM" {Invoke-WPFPanelDISM}
+        "WPFCreateRestorePoint" {Invoke-WPFCreateRestorePoint}
         "WPFPanelAutologin" {Invoke-WPFPanelAutologin}
         "WPFPanelcontrol" {Invoke-WPFControlPanel -Panel $button}
         "WPFPanelnetwork" {Invoke-WPFControlPanel -Panel $button}
--- a/functions/public/Invoke-WPFCreateRestorePoint.ps1
+++ b/functions/public/Invoke-WPFCreateRestorePoint.ps1
@@ -1,0 +1,24 @@
+function Invoke-WPFCreateRestorePoint {
+    <#
+    .SYNOPSIS
+        Enables the ability to create Windows Restore Points
+    #>
+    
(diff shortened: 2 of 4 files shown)


KonTy:
@munkk01 @ChrisTitusTech  imho this is a bit underdeveloped, this doesn't provide any additional functionality to just going to Windows and creating restore point using Windows UI. A better solution would be to have a check box in the tweaks section to create a Restore Point that could be "checked" by default and when you click "apply tweaks" it would just create it and if you don't want it you would uncheck the check box. As it is done now, it has a limited value. :-)

munkk01:
> @munkk01 @ChrisTitusTech imho this is a bit underdeveloped, this doesn't provide any additional functionality to just going to Windows and creating restore point using Windows UI. A better solution would be to have a check box in the tweaks section to create a Restore Point that could be "checked" by default and when you click "apply tweaks" it would just create it and if you don't want it you would uncheck the check box. As it is done now, it has a limited value. :-)

Totally understandable! I just threw together something quickly to get the ball rolling and address the issue that was opened. I'm totally fine with closing this and not using it.

munkk01:
Done deal.

KonTy:
@munkk01 I think the opposite :-), I think we should add one more this to it (a checkbox) and push it out. If you are not familiar with WPF assign it to me, I will look at adding that logic when I have a chance

munkk01:
> @munkk01 I think the opposite :-), I think we should add one more this to it (a checkbox) and push it out. If you are not familiar with WPF assign it to me, I will look at adding that logic when I have a chance

None of this is REALLY in my wheelhouse (I'm not a programmer and just started using git late last year). Hence the underdevelopment, incorrect usage, etc. I'm just trying to learn!

However, I'd like to try and tackle it myself for the sake of doing it. I appreciate any feedback or helpful advice. 

On that note though, I didn't realize that Chris already has a `Set-WinUtilRestorePoint.ps1` implemented and calls it in the `Invoke-WPFRunspace`. If we wanted to continue this route, would we just move that into the `Tweaks` tab itself and default it to `True` but allow users to uncheck it if they choose?

KonTy:
> On that note though, I didn't realize that Chris already has a `Set-WinUtilRestorePoint.ps1` implemented and calls it in the `Invoke-WPFRunspace`. If we wanted to continue this route, would we just move that into the `Tweaks` tab itself and default it to `True` but allow users to uncheck it if they choose?

Good point, I don't know Chris is the ultimate decision maker but I would say so, if it was up to me, I would add a checkbox, "Create a Restore Point" somewhere close to Run Tweaks button and have it be checked by default :) and if it's checked call that function that Chris has or something. Also, thank you for doing all the work and please please keep working on this. Community is how we move projects forward and make things better for everyone.

Issue #1343, opened 2024-01-03
Copy to Ventoy malfuntioning


andreinoue:
**Describe the bug**
"Copy to Ventoy" setting in MicroWin is malfuntioning: creates a unusable file.

**To Reproduce**
1. Run irm https://raw.githubusercontent.com/CodingWonders/winutil/CW-MicroWinFix-Batch2/winutil.ps1 | iex in PowerShell (admin).
2. Download Win11_23H2_EnglishInternational_x64v2.iso from https://www.microsoft.com/en-us/software-download/windows11/.
3. Select Win11_23H2_EnglishInternational_x64v2.iso in MicroWin.
4. Select Windows 11 Pro as the SKU.
5. Tick box "Copy to Ventoy".
6. Try to reboot to Ventoy.
7. See error.

**Expected behavior**
Expected image to work.

**Screenshots**
Didn't take any before replacing image.

**Additional context**
Left MicroWin running and went away from PC. When I came back, PC had turned off. Probably because of an error, but can't tell for sure because I wasn't watching. Turned PC on, File Explorer said there was a microwin.iso file in Ventoy, whose size and date of creation matched microwin.iso in Temp. When trying to reboot to Ventoy, got an error. When deleting the image from Ventoy, Explorer said it had 0GB size.

PR #1371, opened 2024-01-11
Winutil take a long time to create iso file and goes to sleep

KonTy:
This PR fixes issue #1343 when computer does to sleep during a long iso creation.


ChrisTitusTech (approved):
Very Nice, I've never seen the Type added like this. Thank you for teaching me this method.

Issue #1372, opened 2024-01-11
Ventoy not installing


CasalenoGroup:
**Describe the bug**
Ventoy app is not installing from the "Pro Tools" list. Terminal says "No packages found with matching input criteria"

**To Reproduce**
1. Go to 'Install" on Windows Utility
2. Click on 'Ventoy' under Pro Tools
3. Click Install selected
4. See error in terminal window

**Expected behavior**
Programm to install

**Screenshots**
![Screenshot 2024-01-11 102824](https://github.com/ChrisTitusTech/winutil/assets/156302917/4f3eb13b-800f-4806-ac2f-293e92e7d5db)

**Additional context**
Nothing else.

KonTy:
I can't repro this issue, just installed it on my system using winutil and it works fine

``` script
C:\Users\CollinSteward>where ventoy2disk.exe
C:\Users\CollinSteward\AppData\Local\Microsoft\WinGet\Links\Ventoy2Disk.exe
```

godalming123:
Cannot reporoduce this @CasalenoGroup, it works with the same command that KonTy posted for me

CasalenoGroup:
I tried again this morning and now it's working. Thanks

PR #1374, opened 2024-01-11
add winget ventoy package

munkk01:
Addressing open issue.

Issue:
https://github.com/ChrisTitusTech/winutil/issues/1372#issue-2076140240

EDIT:
I'm also not sure what's causing the file changes as a binary file now? Seemed to happen to a previous PR or two when modifying the `applications.json`. Not sure if this will happen to everyone?


ChrisTitusTech (commented):
working on adding a .gitattributes diff to explicitly make these show differences instead of showing binary. Someone is using a mac and added extended attributes or some crap I think... I'll hunt it down today

munkk01:
@ChrisTitusTech You might want to double check the extra commits I inadvertently added here. I was testing and it seems that `applications.json` was at some point converted to `UTF-16 LE` which may have caused that issue to pop up? 

I converted it back to `UTF-8` and had to fix a few more things with special chars. I'm not sure if your `.gitattributes` addition is now forcing it to render properly or if the last commit `update applications.json` did it?

ChrisTitusTech (approved):

PR #1397, opened 2024-01-14
Anoter one of those huge PRs

KonTy:
- Fix version inefficiency the version is already stored in sync, no need to slow down loading by one extra replace.
- Created custom dialog and About message
- Create a menu with Import/Export values
	- press on teh Settings cog in the right upper corner and save all the checkboxes
	- then you can either load them or load and run automatically
- Made Import Export load accross the whole app
- Optimized the way checkbox controls are looked up, it is 20% faster now
- Added a switch to load all the boxes from a config file
	- example: .winutil.ps1 -Config "C:UsersasdfDesktop\111.json"
- Added a switch to run all the action in unattended mode by passing -Run siwthc
	- example: .winutil.ps1 -Config "C:UsersasdfDesktop\111.json" -Run
	- This will run all the tweaks and install all the apps
- BUG: Known issue all the toggle switches are not set when the app loads. These toggle switches do not work if we want to use save to config file architecture, we should switches them to checkboxes, that way we will be able to load them from config and automate.

### DIFF
--- a/Compile.ps1
+++ b/Compile.ps1
@@ -15,7 +15,7 @@ Write-output '
 ################################################################################################################
 ' | Out-File ./$scriptname -Append -Encoding ascii
 
-(Get-Content .\scripts\start.ps1).replace('#{replaceme}',"$(get-date -format yy.MM.dd)") | Out-File ./$scriptname -Append -Encoding ascii
+(Get-Content .\scripts\start.ps1).replace('#{replaceme}',"$(Get-Date -Format yy.MM.dd)") | Out-File ./$scriptname -Append -Encoding ascii
 
 Get-ChildItem .\functions -Recurse -File | ForEach-Object {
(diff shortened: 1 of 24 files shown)


KonTy:
Later

ChrisTitusTech:
I'm going to add you to About section with all this work you've done on the gui and flow of the app. This is amazing work.

KonTy:
Haha thank you Chris, you don't have to, I'm not seeking glory :) just want to help the community

KonTy:
@ChrisTitusTech would it be possible to ask your community to test -Config -Run flags? There are known issues with Toggle buttons for theme changes etc and also uninstalling multiple apps is not working correctly but besides that hope people could test it, see if there is anything else. After that I want to fold it into microwin.iso that would be so awesome. Can't wait to get this going, reinstalling the os will be so easy....

PR #1398, opened 2024-01-15
Import/Export is now global, Settings menu added and many more improvements

KonTy:
- Fix version inefficiency the version is already stored in sync, no need to slow down loading by one extra replace.
- Created custom dialog and About message
- Create a menu with Import/Export values
	- press on teh Settings cog in the right upper corner and save all the checkboxes
	- then you can either load them or load and run automatically
image

- Made Import Export load across the whole app
- Optimized the way checkbox controls are looked up, it is 20% faster now
- Added a switch to load all the checkboxes from a config file
	- example: ```.winutil.ps1 -Config "C:UsersasdfDesktop\111.json"```
- Added a switch to run all the actions from the config file by passing -Run switch
	- example: ```.winutil.ps1 -Config "C:UsersasdfDesktop\111.json" -Run```
	- This will run all the tweaks and install all the apps
- general optimizations making app load faster
- fixed specilize issue with Win10 microwin images

**Next**
- tweaks toggles are not loading from the config file and not being run
- there are a couple of approaches to do this but it needs more thought
        - for example turn all toggles into checkboxes and store them as options, this will certainly eliminate all the issues with toggles
        - move config loading and running code to a different section where is might work better, for example on line ~main.ps1:137
```
$sync.keys | ForEach-Object {
    if($sync.$psitem){
```
- next thing is to put winutil into firstrun.ps1 section for microwin, this will automate the whole installation process when you load your iso come back and all the apps are installed and all tweaks are applied.

### DIFF
--- a/Compile.ps1
+++ b/Compile.ps1
@@ -15,7 +15,7 @@ Write-output '
 ################################################################################################################
 ' | Out-File ./$scriptname -Append -Encoding ascii
 
-(Get-Content .\scripts\start.ps1).replace('#{replaceme}',"$(get-date -format yy.MM.dd)") | Out-File ./$scriptname -Append -Encoding ascii
+(Get-Content .\scripts\start.ps1).replace('#{replaceme}',"$(Get-Date -Format yy.MM.dd)") | Out-File ./$scriptname -Append -Encoding ascii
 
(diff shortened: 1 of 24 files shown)


ChrisTitusTech (approved):
Holy crap... You are amazing!

PR #1400, opened 2024-01-15
Remove Cider Music Player

hamburgerghini1:
Cider has a new version that is not FOSS anymore and it's a paid program.
This is (or will soon be) deprecated.


ChrisTitusTech (approved):
Thanks for the update and following this!

PR #1403, opened 2024-01-15
launch command fix

antunnitraj:


ChrisTitusTech:
That used to be what was their but too many people block insecure connections and why we changed it to have the https:// prefix. Cuts down on issues.

PR #1415, opened 2024-01-18
Reworked all the toggles to be checkboxes

KonTy:
Reworked all the toggles to be checkboxes, to comply with the rest of the tweaks, 

This allows us to use it from the config ``winutil -Config <path> -Run``, now we can automate this and even tweaks, 
UNDO also works and Get Installed also works

@ChrisTitusTech Feel free to not take this PR is you think toggles as they were are more important than ability to automate everything.

NEXT: If PR is taken, fold this into Microwin.iso and run it after install is done


ChrisTitusTech:
Yeah I think having toggles and being more user friendly is far more important than automating it. Most of the user base these days are more normal users than power users.

KonTy:
@ChrisTitusTech sounds good makes total sense.  :-) In this case I would also recommend to back out my other commits that add Global Config Import/Export functionality. See, importing settings doesn't work well with things real time knobs that change the environment in real time when you press a button. And because of that there is a bug right now, if you export let's say a white theme and you have a dark theme set the knobs will act in reverse. (there is an issue reported on that) Which settings you need a positive "hand off" like you load them and click apply button. In that regard I really like how the tweaks are done, it's almost all data driven which is awesome. So I made the real time knobs behave like settings, basically you load them and then click apply so no more real-time environment changes.

Hope you don't mind me splitting the project, I will probably rename it to like WinUtilEx and continue working on this path, I really want to be able to call winutil at the end of the setup in microwin. The point that I want to get to, is create microwin image that will allow and call winutil at the end of setup process to setup all the apps and apply all the tweaks to the system, in order to do that I need this import/export functinality to work properly and for that we can't have real time knobs. :)

PR #1421, opened 2024-01-19
Update README.md

DeveloperDurp:
Added this since I have seen some functions get created without this naming standard. Things won't get loaded into a runspace if it doesn't follow the standard and could cause potential bugs if something in a runspace is trying to call said function.

This is where the script loads in the functions to the runspace if you want to modify it.
https://github.com/ChrisTitusTech/winutil/blob/3b133e704e1d393161ca957f2ce40b111a2355fa/scripts/main.ps1#L13

Not sure if there is a better place to put this.


ChrisTitusTech (approved):
Thanks for clarifying this!

PR #1434, opened 2024-01-21
Added new toggle tweak button

vukilis:
- Added new tweak **"Snap Assist Flyout"** that I think is useful for many people who do not like new feature that **MS** added when hover over **maximize** button.
- When **Disable/Enable** tweak, it will restart **explorer.exe** to take effect and set registry value **0/1** to **EnableSnapAssistFlyout**  
- <i>This will disable **ONLY** snap preview when hover **Maximize** button **NOT** snap preview when is **window** dragged to the edge of the screen.<i>


ChrisTitusTech (approved):
image

SnapAssist Toggle Approved

PR #1447, opened 2024-01-24
fix for bug with selecting presets and toggle switches

DeveloperDurp:
When selecting a preset it will also turn off all the sliders in the tweaks tab. Testing a fix for this issue.

### DIFF
--- a/functions/public/Invoke-WPFPresets.ps1
+++ b/functions/public/Invoke-WPFPresets.ps1
@@ -27,7 +27,7 @@ function Invoke-WPFPresets {
         $CheckBoxesToCheck = $sync.configs.preset.$preset
     }
 
-    $CheckBoxes = $sync.GetEnumerator() | Where-Object { $_.Value -is [System.Windows.Controls.CheckBox] -and $_.Name -like "WPFToggle*" }
+    $CheckBoxes = $sync.GetEnumerator() | Where-Object { $_.Value -is [System.Windows.Controls.CheckBox] -and $_.Name -notlike "WPFToggle*" }
     Write-Debug "Getting checkboxes to set $($CheckBoxes.Count)"
 
     $CheckBoxesToCheck | ForEach-Object {
--- a/functions/public/Invoke-WPFPresets.ps1
+++ b/functions/public/Invoke-WPFPresets.ps1
@@ -27,7 +27,7 @@ function Invoke-WPFPresets {
         $CheckBoxesToCheck = $sync.configs.preset.$preset
     }
 
-    $CheckBoxes = $sync.GetEnumerator() | Where-Object { $_.Value -is [System.Windows.Controls.CheckBox] }
+    $CheckBoxes = $sync.GetEnumerator() | Where-Object { $_.Value -is [System.Windows.Controls.CheckBox] -and $_.Name -like "WPFToggle*" }
     Write-Debug "Getting checkboxes to set $($CheckBoxes.Count)"
 
     $CheckBoxesToCheck | ForEach-Object {


KonTy:
@DeveloperDurp this is great, FIY I'm working on ripping out global config, import/export functionality Chris said that he prefers "user friendly" UI to be able to automate things. Basically the issue is that these real time toggle controls don't work nicely with storing things in the config. For example, if you store White Theme in the config file, but you have Dark Theme loaded and the toggle is set the to Dark theme. When you load the config file do you then flip the toggle immediently to Dark? But if you do they user might not want that be changed in real time, they might want to preg config  for later but not necessarily change it in real time. So import/export config functionality doesn't work nicely with real time controls like these toggles. What I proposed in thie PR https://github.com/ChrisTitusTech/winutil/pull/1415#issuecomment-1901044106 is to change them to check boxes so you can still import/export them but if you want to "do an action" you have to press "Run Selected" button. This way it also makes the UI consistent between all of the functionality accross winutil. So basically I made all the toggles work like the rest of the tweaks, you can "Run Selected", Clear, "Get Installed" etc. check it out here 

``https://raw.githubusercontent.com/KonTy/winutil/test-2024-01-15/winutil.ps1``

I think I'm going to fork the project, It is very important for me to be able to automate things. What I want to accomplish in the end is be able to save the config, with all the apps, tweaks, themes etc. Embedd it into Microwin, make an ISO and as a last step apply all the tweaks and install all the apps so when you come back everything will be done and installed.

KonTy:
@DeveloperDurp could you also ping me, I would like to talk to you about running UI on a different thread. I didn't bother much with all the work I added. And I didn't bother learning how you did this since you are the expert in this area and I didn't want to waste my time with that. But I think I need your advise and also would like to bounce a couple of ideas off of you. For instance, I think when things are being run in the background, I think it would be better to through some modal dialog box with a spinning wheel or something on the UI that way people would start another action. I've noticed when the tool is doing things and you start a new action, a lot of times it would crash etc.

DeveloperDurp:
> @DeveloperDurp this is great, FYI I'm working on ripping out global config, import/export functionality Chris said that he prefers "user friendly" UI as oppose to having the ability to automate things. Basically the issue is that these real time toggle controls don't work nicely with storing things in the config. For example, if you store White Theme in the config file, but you have Dark Theme loaded and the toggle is set the to Dark theme. When you load the config file do you then flip the toggle immediently to Dark? But if you do they user might not want that be changed in real time, they might want to preg config for later but not necessarily change it in real time. So import/export config functionality doesn't work nicely with real time controls like these toggles. What I proposed in thie PR [#1415 (comment)](https://github.com/ChrisTitusTech/winutil/pull/1415#issuecomment-1901044106) is to change them to check boxes so you can still import/export them but if you want to "do an action" you have to press "Run Selected" button. This way it also makes the UI consistent between all of the functionality accross winutil. So basically I made all the toggles work like the rest of the tweaks, you can "Run Selected", Clear, "Get Installed" etc. check it out here

@KonTy We had that originally some time ago but the problem was how do you communicate to the user when you want to do a "toggle" IE enable dark mode, if checked it will enable it but how do you communicate to disable it? a second check box? I hear where you are coming from and that's the reason I tried to make everything as modular the best I could. However as Chris has mentioned winutil is meant to be a user friendly ui. I have setup some frameworks for other tools to piggyback off the main tool however. 

> 
> `https://raw.githubusercontent.com/KonTy/winutil/test-2024-01-15/winutil.ps1`
> 
> I think I'm going to fork the project, It is very important for me to be able to automate things. What I want to accomplish in the end is be able to save the config, with all the apps, tweaks, themes etc. Embedd it into Microwin, make an ISO and as a last step apply all the tweaks and install all the apps so when you kick off in ISO install and come back after it is done everything will be done and installed.

All of the tweaks live inside the config files so you wouldn't really need to use winutil to achieve this. This is where the modular structure comes in. There are several functions in the private section that handle the tweaks through the use of parameters, if memory serves me it should be the exact same parameters that get generated through the import/export process. 

> @DeveloperDurp could you also ping me, I would like to talk to you about running UI on a different thread. I didn't bother much with all the work I added. And I didn't bother learning how you did this since you are the expert in this area and I didn't want to waste my time with that. But I think I need your advise and also would like to bounce a couple of ideas off of you. For instance, I think when things are being run in the background, I think it would be better to through some modal dialog box with a spinning wheel or something on the UI that way people would start another action. I've noticed when the tool is doing things and you start a new action, a lot of times it would crash etc.

I am not sure which thread you would want to continue that discussion but yes, I did notice some places that the runspace structure would benefit the additions you have made. As for running the gui in its own runspace, it's possible but introduces other complexities. Right now because it is on the main thread it can do quick actions itself or spawn threads to handle longer running tasks. However if the gui was in a runspace you now have to interact with the main thread to spawn new tasks and adds additional complexity that I didn't see the benefit of having. If the task is using a runspace it shouldn't be crashing when a user attempts to do another action as everything I have setup prevents you from doing more than 1 action. If you find it crashing that might be something that needs to be ran in a runspace. 

I also agree with you, we should have better feedback when things are running. Right now it's all in the console and I have had ideas as well to improve this but I have ended up stepping away for some time to work as other projects have taken priority.

DeveloperDurp:
Gotcha, I see what you mean. I will spend some time this weekend to see if I can think of a solution however I am in agreement with Chris on trying to keep this easier on the end user over automation so having text saying "uncheck for light mode check for dark mode" is a no go for me. Toggle switches have been a pain in the past but there are ways to work around it. 

if anything it would be part of the import/export process and when you import the config it will also handle the toggles as well. You should be able to piggy back from there.

also going to set this back to drafts, I should mirror what I have here

https://github.com/ChrisTitusTech/winutil/blob/3b133e704e1d393161ca957f2ce40b111a2355fa/functions/private/Get-WinUtilCheckBoxes.ps1#L36-L38

KonTy:
@DeveloperDurp sounds good let me know what you come up with. I'm not going to make any more changes untl I hear from you. Please take a look at my global import/export changes, it is buggy because of toggles. 

As far an automation goes I don't agree with prioritizing toggles vs automation. Chris himself said the reason why he started the script is because he wanted to automate tweaks at work, and going back on that vision it really strange. Creating an automation tool that ends up, in the end, needing a lot of clicks :-)

If you wanna see what I'm tlaking about look at my changes in main where you can pass ``-Config <file> -Run`` -- that would run automation that is loaded from the config. Anyway let me know what you decide.

KonTy:
@DeveloperDurp thinking about it some more, we could turn those into drop downs, like Windows. You can chose your tweaks from the drop-down and then click "Run Tweaks". 
image

This way we can store it in the config and get around this real time toggling issue. We have to do this for hte DNS drop down list, I didn't do, but if we get DNS working we can have more drop-down lists

DeveloperDurp:
> @DeveloperDurp thinking about it some more, we could turn those into drop downs, like Windows. You can chose your tweaks from the drop-down and then click "Run Tweaks". image
> 
> This way we can store it in the config and get around this real time toggling issue. We have to do this for hte DNS drop down list, I didn't do, but if we get DNS working we can have more drop-down lists. 

Could be an option, Through we can treat the toggles exactly as all the other checkboxes, if you pass the name of the checkbox assume enabled, if missing assume disabled. Then the export function can pull the names of enabled toggles and pass it to the json. Though I guess that could get a little weird because the default windows install will have some of those options enabled.

I think there might also be a miscommunication when we say end user > automation. I am all for automation and should strive for that, however the interface needs to be easy to understand because not everyone is a power user. IE the checkbox for dark mode situation would confuse a lot of people, however your drop down idea might be a possible solution to that issue. Tagging @ChrisTitusTech to get his opinion.

ChrisTitusTech (approved):

ChrisTitusTech:
I'm still working on the user friendly toggles for the tweaks section. I need to rework the wording of the tweaks on the run and undo. Maybe switch them to Enable / Disable or something that is a bit more clear.

KonTy:
@ChrisTitusTech I will have a POC sometime today for this. I think Derp and I have an idea I just need to finish it. Let me send you a message sometime today.

KonTy:
@ChrisTitusTech @DeveloperDurp hey guys take a look at my experimental toggle design. So basically it is not real time, you have to toggle things and click "Run Tweaks", you can also click "Get installed" and that will load all the tweaks that are applied to the system. This is still not finished, if you untoggle things and "run tweaks" it will just get ignored. I didn't spend any time on that logic, just worked on the UI design, besically lifted it from Windows 11

``irm https://raw.githubusercontent.com/KonTy/winutil/test-2024-01-15/winutil.ps1 | iex``

PR #1465, opened 2024-01-26
Update winutil.ps1

KayleighOwO:
Fixed some minor spelling/grammar issues. Changed hover description for the Zones option as it was just copied from the File Extension setting


OmriOn (approved):

munkk01:
These modifications are directly to `winutil.ps1` which will be overwritten at the next compile time. These changes should be made to `inputXML.xaml` instead.

Additionally, PR's should be made to whatever `test` branch is available rather than `main`.

In any case, thanks for the contributions!

PR #1467, opened 2024-01-26
Lots of changes to Invoke-WPFFixesUpdate

Carterpersall:
Made a lot of changes to Invoke-WPFFixesUpdate:
- Added a description that details exactly what the function does
- Added 'Aggressive' parameter
  - Put some actions behind it that take a long time, have a risk of causing problems, or are usually unnecessary
  - Could be a toggle within the UI, but I don't know how to add one
- Added progress bars
  - This adds progress bars to most actions, giving a more user-friendly indicator of progress
  - This part took the longest to add by far, mostly due to my attempts to redirect SFC's output
    - When redirected, SFC only seems to output when the stdout buffer is full, making it output in a few large chunks
      - Seems to be a bug with SFC and not with PowerShell or .NET
    - For whatever reason, SFC outputs in Unicode instead of UTF-8, which causes problems when PowerShell attempts to parse it
    - It also uses /CR/CR/LF instead of the /CR/LF or /LF that most other programs use, seems to be yet another bug on SFC's side, albeit a really weird one
- Added system corruption scan
  - Aggressive only due to the significant amount of time it can take
- Fixed Windows Update discovery
  - `wuauclt /detectnow` has been deprecated, but still works, albeit not how one would expect
    - See more about that [here](https://www.ajtek.ca/wsus/does-wuauclt-exe-detectnow-reportnow-work/)
- Removed the previous step 10
  - Only applies to Windows 8
- Lots of small code and formatting tweaks

Been working on this on and off for a few months now, mostly due to the SFC redirection issue. System repairs are a deep rabbit hole that I could dive into for several months more, but I think I have this at a good point.

Git decided to do that thing again where it gives the pull request a bunch of commits I didn't do. So there are a few extra commits in here that I had to undo to get things back to normal.


ChrisTitusTech (commented):
Looks great Carter will put this through tests and merge later this week.

ChrisTitusTech (approved):

PR #1474, opened 2024-01-26
Add applications

Yuuh15:
Added sonarr, radarr, prowlarr and bazarr


ChrisTitusTech (commented):
Fantastic programs, but these really should be used in docker container or on a server. Windows desktop isn't where I'd put these. Closing this out.

PR #1497, opened 2024-01-29
Calling Invoke-WPFRunspace with argument fix

ogfrm:
$DebugPreference is not sent as different argument


ChrisTitusTech (approved):
Good Catch

PR #1510, opened 2024-01-31
Added Mega Sync

itz-Me-Pj:
Added [mega sync](https://mega.io/) to applications utilities section

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -1133,6 +1133,15 @@
 		"panel": "0",
 		"winget": "Element.Element"
 	},
+	"WPFInstallMEGAsync": {
+		"category": "Utilities",
+		"choco": "megasync",
+		"content": "Mega Sync",
+		"description": "Easy automated synchronisation between your computer and your MEGA cloud.",
+		"link": "https://mega.io/",
+		"panel": "4",
+		"winget": "Mega.MEGASync"
+	},
 	"WPFInstallmeld": {
 		"category": "Utilities",
 		"choco": "meld",


KonTy:
@ChrisTitusTech we have so many apps in winutil now, it is getting slow when loading. Do we want paid apps/services in the tool? IMHO we should have strickter acceptance criteria for apps? I do have some new code that I'm working on that will give us more space on the Install page but still WinUtil is not an app store.

itz-Me-Pj:
@KonTy the utility also has a free version as well with enough storage required. And the utility is added because of it's zero knowledge encryption policy much better than it's alternative like google drive and one drive that are constantly sneaking up on your data.

KonTy:
Strange I didn't see any free tiers I only so a tier that starts at $109. Where are the free tiers?

Yuuh15:
> @ChrisTitusTech we have so many apps in winutil now, it is getting slow when loading. Do we want paid apps/services in the tool? IMHO we should have strickter acceptance criteria for apps? I do have some new code that I'm working on that will give us more space on the Install page but still WinUtil is not an app store.

Yeah, paid apps shouldn't be on the utility. And is there anything that can be done about the slow loading?

KonTy:
> Yeah, paid apps shouldn't be on the utility. And is there anything that can be done about the slow loading?

@Yuuh15 you are right, see because this is a script and not a compiled app, everything we do has to be done in real time, and the more apps we have the slower it is going to get. My point is winutil is not an app store we should only have Open Source projects that are free for everyone to use and have source posted on the web. 

PS as another consideration, I'm thinking about building a plugin infrastructure where you get core apps quickly but then you could click a button and load more apps from the list on github, same for tweaks.

ChrisTitusTech:
Yeah we might need to cull some apps. I'm going to close this for now. We need to update application json anyways to remove panel with a new pr I added

PR #1519, opened 2024-02-01
Addition of ImageMagick

andrewcollier-com-au:
convert over 100 file types.


ChrisTitusTech:
Changed the way applications are added and panel parameter is no longer needed

PR #1526, opened 2024-02-01
Choosing Alternate Scartch Path and Busy Messages

LeeDowA:
I think i finally got it. 🥇 
If its still making problem feel free to implement this idea yourself :)

### DIFF
--- a/functions/public/Invoke-WPFGetIso.ps1
+++ b/functions/public/Invoke-WPFGetIso.ps1
@@ -12,6 +12,10 @@ function Invoke-WPFGetIso {
         return
     }
 
+  $sync.BusyMessage.Visibility="Visible"
+    $sync.BusyText.Text="N Busy"
+
+
     Write-Host "         _                     __    __  _         "
 	Write-Host "  /\/\  (_)  ___  _ __   ___  / / /\ \ \(_) _ __   "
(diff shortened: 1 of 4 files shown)


LeeDowA:
A Side Note:
I found a bug in the compile.ps1 
when they more then one file with the same function it is causing havoc,

ChrisTitusTech (commented):
Great PR! ;)

functions/public/Invoke-WPFGetIso.ps1
```diff
@@ -138,7 +173,7 @@ function Invoke-WPFGetIso {
             $imageName = $_.ImageName
             $sync.MicrowinWindowsFlavors.Items.Add("$imageIdx : $imageName")
         }
-        $sync.MicrowinWindowsFlavors.SelectedIndex = 0
+        $sync.MicrowinWindowsFlavors.SelectedIndex = 5
```
CodingWonders, line 176: What if the Windows image selected doesn't contain 11 indexes? What if the Pro edition isn't the sixth index?

This can fail by throwing an exception. We shouldn't do it this way.

LeeDowA:
I don't know from where this change was made. :o
 i didn't made that change about the index and it not related to my contribution. /:

PR #1530, opened 2024-02-01
Add Adobe Creative Cloud Cleaner tool

mrkenhoo:
This PR adds the **Adobe Creative Cloud Cleaner** tool.

I have tested it many times and it works.

Hope the commits are as clear as it must be as well.


ChrisTitusTech (changes requested):
Can you pull down the new changes from the test branch and use feature.json for adding this instead of using the inputXAML as this was the old method. Thank you!

Issue #1496, opened 2024-01-29
Windows Features from Config tab not geting installed


zzarac:
**Describe the bug**
Clean installation of Windows 11 ("world" setup) OS build 22631.3007 (January 2024), installed drivers and all updates.
Uninstalled most remaining bloat before starting winutil.
In winutil, first fixed winget, then ran "Desktop" recommended tweaks which all finished succesfully.
But when selecting any of the Features from "Config" tab and clicking "Install Features", I get this text in powershell window and nothing happens. Everything worked fine in previous OS build (December 2023).

===================================
---   Features are Installed    ---
---  A Reboot may be required   ---
===================================

**To Reproduce**
1. Install WIndows 11 OS build 22631.3007 (January 2024)
2. Start winutil
3. Click "Config" tab
4. Select any of the features
5. Click "Install Features"

**Expected behavior**
Selected windows features install and the world is fine.

**Screenshots**
![image](https://github.com/ChrisTitusTech/winutil/assets/6088246/088d9c0e-aa26-4755-b657-d6ea8fc21946)

**Additional context**
When I try clicking "Install Features" again, the info text "features being installed, a reboot may be required" is printed again, as the screenshot shows.

zzarac:
Maybe I should have added that I used OS.click website (https://os.click/en) to download the latest Windows build ISO (currently 22631.3007).

DansNewLegs:
Same issue here, downloaded windows 11 from official site and used microwim too on it.

zzarac:
I just tried installing some windws features with winutil on a separate machine that is still on Windows 11 21H2, and it didn't work there either.

I guess there must have been some pull request that misbehaved because I used winutil in the past on that machine with no problems, and windows there has not been updated for ages.

Yuuh15:
After some testing the issue seems to have started at the  commit to main branch made by Chris in 2024-01-12. The logic of checking if the checkboxes are selected changed a lot in that commit, and it seems to be causing it.

Garou7:
Same here on Windows 11 Pro 22H2 (official iso not MicroWin):
![fe](https://github.com/ChrisTitusTech/winutil/assets/113905482/c038e85b-b364-4f0c-beea-b511aeaa8901)

zzarac:
Fixed in last update.

Issue #1457, opened 2024-01-26
scope=machine flag breaks some package installs


lakub-muravlov:
**Bug description:** 
after latest release there was --scope=machine added to winget install script, which leads to some packages failing to get installed

**To Reproduce**
1. Go to app
2. Select application(s) - can be reproduced with discord, onlyoffice, rust, teams, viber
3. Install selection
4. Install fails with error "No applicable installer found; see logs for more details."

**Expected behavior**
Package should be installed

**Screenshots**
![Screenshot 2024-01-26 024203](https://github.com/ChrisTitusTech/winutil/assets/85365426/d130b234-bc34-43b3-8f28-f7d26077a6dd)

**Additional context**
Removing this flag resolves the issue
![Example](https://github.com/ChrisTitusTech/winutil/assets/85365426/f7ff94b2-8f38-402a-99a2-499cb7129352)

Nathan-Shea:
I'm also having this problem for pretty much every app. Removing the flag resolved the issue for me as well.

https://github.com/ChrisTitusTech/winutil/issues/1494#issuecomment-1916145351

KonTy:
I think this is the issue if you already have apps installed as a local scope and try to install it again as a machine scope? I just tried installing all these apps without any problems with the machine scope. 

I wonder if there is a way in winget to automatically detect installation and thier scope and do the appropriate thing.

KonTy:
Hmm so strange I'm still getting this error on load, it then loads just fine. We should probably ask more people to test it. I wonder why you are not seeing this can you run this command? This seems to be busted:

`` irm https://raw.githubusercontent.com/ogfrm/winutil/test-2024-01-30/winutil.ps1 | iex``




``` powershell

convertfrom-json : Invalid JSON primitive: .
At line:7876 char:6
+ }' | convertfrom-json                                                                                                                              +      ~~~~~~~~~~~~~~~~                                                                                                                                  + CategoryInfo          : NotSpecified: (:) [ConvertFrom-Json], ArgumentException                                                                    + FullyQualifiedErrorId : System.ArgumentException,Microsoft.PowerShell.Commands.ConvertFromJsonCommand
```

Issue #1494, opened 2024-01-28
No apps are getting installed


GamingNerdLeith:
**Describe the bug**
Whenever you try to install applications an error "No applicable installer found; see logs for more details." appears. This so far has appeared with the following apps:

**"OBSProject.OBSStudio", "KeePassXCTeam.KeePassXC", "JanDeDobbeleer.OhMyPosh", "OpenJS.NodeJS", "Notepad++.Notepad++",
"VSCodium.VSCodium", "LizardByte.Sunshine", "GOG.Galaxy",  "Microsoft.PowerShell", "TorProject.TorBrowser", "SomePythonThings.WingetUIStore", "Nvidia.GeForceNow", "Nilesoft.Shell", "7zip.7zip", "Unity.UnityHub", "Discord.Discord",
"EpicGames.EpicGamesLauncher", "RARLab.WinRAR", "BlenderFoundation.Blender", "Microsoft.Teams"**

**To Reproduce**
1. Go to 'Start'
2. Type "Terminal"
3. Type "iwr -useb https://christitus.com/win | iex" to open WinUtil
4. Press Enter
5. Click on the apps listed above 
6. Click on 'Get Installed'
7. Click on 'Installed Selected'
8.  See Error

**Expected behavior**
The applications will be installed into Windows 11 without error. 

**Screenshots**
![image](https://github.com/ChrisTitusTech/winutil/assets/67237948/ac2fa2a4-0b5b-4b43-a3bd-dd64ca0428d7)

**Additional context**
Processor	Intel(R) Core(TM) i5-10300H CPU @ 2.50GHz   2.50 GHz
Installed RAM	16.0 GB (15.9 GB usable)
System type	64-bit operating system, x64-based processor
Edition	Windows 11 Pro
Version	23H2
Installed on	‎28/‎01/‎2024
OS build	22631.3085
Experience	Windows Feature Experience Pack 1000.22684.1000.0

Tried to reinstall winget with the issue not being resolved.

RocGod:
I have this issue as well, but it is not applied to all the applications. Some apps can still be installed.

TheFlano23:
Same issue here. Tried following fixes/workarounds on other issues but none have worked so far. On Windows 10 22H2

tegaidogun:
Same issue here, it seems to be from the `scope=machine` stuff in the scripts.

This thing is really annoying me.

Nathan-Shea:
[...]

PR #1533, opened 2024-02-02
Fix issue #1496 and #1457

Yuuh15:
- Revert some changes related to the checkboxes made some weeks ago.
- Remove "--scope=machine" argument from winget install, as it blocks some applications from installing.

### DIFF
--- a/config/helperscript.ps1
+++ b/config/helperscript.ps1
@@ -210,7 +210,7 @@ $jsonfile | ConvertTo-Json | Out-File ./config/preset.json
         ],
         "InvokeScript": [
             "Script to make modifications not possible with the above types
-            Special care needs to be taken here as converting from json to a scriptblock
+            Special care needs to be taken here as converting from json to a ScriptBlock
             can cause weird issues. Please look at the example below to get an idea of how things should work"
         ],
         "UndoScript": [
--- a/config/preset.json
+++ b/config/preset.json
@@ -10,7 +10,7 @@
     "WPFTweaksStorage",
     "WPFTweaksTele",
     "WPFTweaksWifi",
-    "WPFMiscTweaksPower"
+    "WPFTweaksPower"
   ],
   "laptop": [
     "WPFTweaksAH",
@@ -22,7 +22,7 @@
     "WPFTweaksStorage",
     "WPFTweaksTele",
     "WPFTweaksWifi",
-    "WPFMiscTweaksLapPower"
+    "WPFTweaksLapPower"
   ],
   "minimal": [
     "WPFTweaksHome",
--- a/functions/private/Get-WinUtilCheckBoxes.ps1
+++ b/functions/private/Get-WinUtilCheckBoxes.ps1
@@ -20,58 +20,56 @@ Function Get-WinUtilCheckBoxes {
     #>
 
     Param(
-        [boolean]$unCheck = $false
+        $Group,
+        [boolean]$unCheck = $true
     )
 
-    $Output = @{
-        Install      = @()
-        WPFTweaks     = @()
-        WPFFeature    = @()
-        WPFInstall    = @()
+
+    $Output = New-Object System.Collections.Generic.List[System.Object]
+
+    if($Group -eq "WPFInstall"){
+        $filter = Get-WinUtilVariables -Type Checkbox | Where-Object {$psitem -like "WPFInstall*"}
+        $CheckBoxes = $sync.GetEnumerator() | Where-Object {$psitem.Key -in $filter}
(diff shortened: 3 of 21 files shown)


KonTy:
removing --scope=machine creates all sorts of other problems, we can't just remove it. For example, if you install python or any developer environment it will not put it in the right place and you will not be able to use python properly. Including pip abd other utilities. If we want to fix this properly we need to leave scope as machine by deault and have some kind of flag for installs that breaks.

Can you give me example of installs that get blocked by this scope=machine?

KonTy (commented):
Did you test import/export and -Config -run functinality with this change? It relies on Install being in the hashtable WPFInstall and Install and 2 distict code pathways, and both are needed. Make sure you didn't break other functinality

KonTy:
Also I see that you changed look up seach wildcard patterns in the code and the actual controls in inputxaml were not changed back. So you might have broken other things as well with this change. I suggest doing a more surgical Did you test import/export and -Config -run functinality with this change? It relies on Install being in the hashtable WPFInstall and Install and 2 distict code pathways

Yuuh15:
> removing --scope=machine creates all sorts of other problems, we can't just remove it. For example, if you install python or any developer environment it will not put it in the right place and you will not be able to use python properly. Including pip abd other utilities. If we want to fix this properly we need to leave scope as machine by deault and have some kind of flag for installs that breaks.
> 
> Can you give me example of installs that get blocked by this scope=machine?

Issue #1494 states some of the applications that don't install with the argument.

Yuuh15:
There is a bit of work to be done here, I admit, this should have been a draft, but the main thing that I was trying to check out was the root cause of the features not installing.

KonTy:
> There is a bit of work to be done here, I admit, this should have been a draft, but the main thing that I was trying to check out was the root cause of the features not installing.

Good job keep up the good work. What I recommend is please try to actually fix the bugs instead of reverting changes. There is other work around the module **and** planned work, that is in the works, that relies on these changes. Usually it is not a good idea to revert things :-) because this could blow up other features and stall work in progress. But in any case, thank you jumping in and proactively fixing issues. 

PS in the nutshell Install and WPFInstall are two sections that are "work in progress" for automating installs and tweaks. WPFInstall secion just sets the Checkboxes on the UI where is Install section will allow us to install apps directly without even having to load the UI. You might argue that you don't really need "Install" section and that might be true because in theory you could lookup all of the commands from the json file but that is a lot more work. Let's finish that feature and think how to optimize it?? :) But at the moment if you nuke all these changes it will break lots of forward looking features.

PS What surprises me is those errors that people report. It must have been a bad merge because we tested it a lot and it was working fine. Chris even had a video where it was working fine on screen. I'm not sure what happened here.

KonTy:
> Issue #1494 states some of the applications that don't install with the argument.

agreed. maybe we revert back to local installs and introduce scope variable to the json? I think though I would rather have machine scope by default, because local scope also breaks a lot of apps because winget puts all the apps int overy obscure directory and people can't find it. 

``C:\Users\CollinSteward\AppData\Local\Microsoft\WinGet\``

I would rather have machine scope and have winget put files into ``c:\program files`` by default. btw are you sure it is the scope issue? In my branch I'm still setting scope and machine and all of these apps listed in the issue https://github.com/ChrisTitusTech/winutil/issues/1494 install just fine

Yuuh15:
Thanks for explaining all of that to me konty, I have learned a lot since I started looking at the code. Closing this PR

KonTy:
> Thanks for explaining all of that to me konty, I have learned a lot since I started looking at the code. Closing this PR

Thank you for your enthusiasm and thank you jumping in and fixing issues. I think looking into it some more, do you want to remove scope = machine? and maybe if you have time add scope to applications.json? You can either write a macro or a little app to add scope to each node in the json?

Here is a small powershell snippet code  that you can use as a starting point to add scope to all nodes in the applications.json

``` Powershell

# Convert JSON to PowerShell object
$jsonObject = $jsonContent | ConvertFrom-Json

# Add "scope": "local" to each node
$jsonObject | ForEach-Object {
    $_.Value.Add("scope", "local")
}

# Convert back to JSON
$jsonWithScope = $jsonObject | ConvertTo-Json -Depth 10

# Output the modified JSON
Write-Output $jsonWithScope
```

Yuuh15:
I may try it, but it seems like a band-aid solution

KonTy:
@Yuuh15 do you think a proposal to do it some other way? I'm just thinking local scope definitely breaks some apps and if machine scope breaks others being able to specify scope seems like a logical solution, no? :)

Yuuh15:
It does seem like the best way :v

PR #1534, opened 2024-02-02
application are evenly distributed to 5 columns

ogfrm:
As KonTy suggested, I modified the application tab so that applications are evenly distributed to 5 columns. That way we don't need panel property in the application.json file.


KonTy:
Thank you for doing this. Hope Chris likes this as this will buy us even more time before we will stop accepting new apps :-)

PR #1537, opened 2024-02-02
Fix features not being installed and winget reinstall

Yuuh15:
- Remove a typo that was causing the features from config tab to not being installed (No nuking the codebase this time hehe).
- Add "--force" argument to winget reinstall.


ChrisTitusTech (approved):
That looks clean. Thanks for catching this.

PR #1539, opened 2024-02-02
Improve detection for Pro editions of Windows

CodingWonders:
This PR significantly improves the automatic Professional edition detection created by @LeeDowA. The following was done:

- Modified the code that assumed that the Pro edition was always in 6th place (bad programming action) and replaced it with image detection

Using image detection, we can get the Pro edition of Windows much more reliably by checking if the `EditionId` property of a Windows image is "Professional". However, this slows down the process of specifying the ISO, but this is for the better as it can detect any index changes in the Windows image.

Here is a demonstration video showing this functionality:

https://github.com/ChrisTitusTech/winutil/assets/101426328/483b7dfb-2296-41ce-b300-0f959d49c8bc

Any (future) optimization to this function is welcome.


KonTy:
@CodingWonders good job, love it!

ChrisTitusTech (approved):

ChrisTitusTech:
Perfect, exactly what I wanted when I modified that PR yesterday to just do the static 5 index.

PR #1554, opened 2024-02-03
JSON config fix

Zeunig:
Config tab works now with these changes


ChrisTitusTech:
wrong file and branch.

PR #1569, opened 2024-02-04
Update winutil.ps1 to not use scope=machine, as it breakes many packages installation

lakub-muravlov:


Yuuh15:
The file will be overwritten on compilation, and these changes would create other kinds of problems

ChrisTitusTech:
This is already been fix on test branch. A couple packages from some custom Discord apps and a few other apps failed from hash mismatches. I added some checks to fix this.

Issue #1120, opened 2023-10-21
Change region to World


jstnas:
**Is your feature request related to a problem? Please describe.**
"The Perfect Windows 11 Install" guide mentions changing the region to `World` to bypass bloatware, though it seems that `winget` will not be installed in the process, so changing the region temporarily is required.
However, Windows will prevent you from restoring the region back to `World` through the `Country or region` setting.
Manually changing the region is possible through the registry at `Computer\HKEY_CURRENT_USER\Control Panel\International\Geo`.

**Describe the solution you'd like**
Adding an option to set the region to `World` probably in the Advanced Tweaks section.

**Additional context**
The data that would need to be changed in the registry is as follows:
`Name` to `World`
`Nation` to `39070`

PR #1579, opened 2024-02-06
Add advanced tweak "set region to world"

Yuuh15:
Added this tweak based on the issue #1120


ChrisTitusTech:
The only purpose of a World toggle is to change this to your region. Setting it to World would cause the MSStore to stop functioning. Not something I would want normal users to do.

PR #1589, opened 2024-02-08
remove tweak toggle functions and add them to tweaks.json

ogfrm:
- Move the xaml file autogeneration function to compile. Startup time may improve
- Remove tweaks/toggle functions and add them to tweak.json file
- Invoke-WinUtilCurrentSystem function is separated to handle tweak portion. This way it will not continue to check other parameters if it is false. It will scan faster
- removes check Get-WinUtilToggleStatus function. Since toggle functions are in json file and Invoke-WinUtilCurrentSystemTweak function can handle individual tweak we can use that


Compile.ps1
```diff
@@ -1,3 +1,113 @@
...
+                        $blockXml +="`n</StackPanel>`n</Border>`n"
+                        $blockXml += "<Border Grid.Row=""1"" Grid.Column=""$panelcount"">`n<StackPanel Background=""{MainBackgroundColor}"" SnapsToDevicePixels=""True"">`n"
+                        $panelcount++
+                    }
+                }
+                $appInfo = $organizedData[$panel][$category][$appName]
+                if ("Toggle" -eq $appInfo.Type) {
+                    $blockXml += "<StackPanel Orientation=`"Horizontal`" Margin=`"0,10,0,0`">`n<Label Content=`"$($appInfo.Content)`" Style=`"{StaticResource labelfortweaks}`" ToolTip=`"$($appInfo.Description)`" />`n"
+                    $blockXml += "<CheckBox Name=`"$($appInfo.Name)`" Style=`"{StaticResource ColorfulToggleSwitchStyle}`" Margin=`"2.5,0`"/>`n</StackPanel>`n"
+                } elseif ("Combobox" -eq $appInfo.Type) {
```
KonTy, line 79: I wonder how after all these changes we will be changing UI. There is a separation of business logic and UI on purpose, that's why we have XAML and code logic separately. I wonder how we are going to change UI after these changes. Do you expect to jump around beween code and XAML? It will be extremly hard to adjust UI. Imagine you want to change the layout of these pages. 

another problem with code like that in addition to breaking principle of MVC, it is impossible to read with allthe loops, regex replacments in place. etc. IMHO this is truly a nightmare code, that I wouldn't want to maitain personally :-)

But you know what they say: "What works for one may not work for another" I'm just expressing my opinion Chris has to make the call :)

ogfrm: > I wonder how after all these changes we will be changing UI. There is a separation of business logic and UI on purpose, that's why we have XAML and code logic separately. I wonder how we are going to change UI after these changes. Do you expect to jump around beween code and XAML? It will be extremly hard to adjust UI. Imagine you want to change the layout of these pages.
> 
> another problem with code like that in addition to breaking principle of MVC, it is impossible to read with allthe loops, regex replacments in place. etc. IMHO this is truly a nightmare code, that I wouldn't want to maitain personally :-)
> 
> But you know what they say: "What works for one may not work for another" I'm just expressing my opinion Chris has to make the call :)

I haven't changed the function that changes xml file. I just move it compile file because it doesn't have to run on runtime.  I agree that function is little complicated. The way the function works was in the code before I changed it. It was organizing applications. I just added tweaks and features tabs to that function. I wouldn't have write the function like that but I did not want to delete and rewrite function so I adjust it. I don't understand the problem with jumping between xaml to code part. Before my changes to add tweak or feature you have to change json file and xaml file. Now you just have to change json file. To change layout you have to change just xml file like before.

ogfrm: I also do not recall adding any regex or any extra loop. I was planning to delete the order and panel properties from tweaks and features json file and make the order in json file make it little cleaner.

PR #1593, opened 2024-02-08
Clean up white-spaces

veniware:
Remove unnecessary white-spaces and replaced tabs with 4 spaces for consistency.


RocGod:
@veniware I think you need to raise the PR against test branch rather than main branch.

PR #1595, opened 2024-02-09
Improve unattended answer file creation for MicroWin and add automatic conversion of ESD files

CodingWonders:
I've modified the `Test-CompatibleImage` procedure to now detect if the image version is greater than or equal to a "desired version", rather than detecting whether or not it is Windows 8 or earlier.

The contents of the base unattended answer file now contain `<#REPLACEME#>`, which is replaced by either the Specialize pass (on Windows 11 images) or an empty string (on Windows 10).

I've tested this function on both a Windows 10 and Windows 11 image, and this procedure works. I've also opened the unattended answer files in Windows SIM and they contain the correct information (after converting the unattended answer files to UTF-8 with BOM).

imgmgr_4RFYyGnq4S

imgmgr_1lC1ioeonq

However, I **haven't** tested them with Windows Setup, which is what determines if the answer files work, and some problems may arise.

This PR also converts the `install.esd` file automatically to a WIM file if the installation media provided has said file. No conversion will occur with images that contain `install.wim` files


ChrisTitusTech:
Like this, noticed some issues with the unattended xml on Microwin. Going to try and review this tomorrow.

CodingWonders:
> Like this, noticed some issues with the unattended xml on Microwin. Going to try and review this tomorrow.

OK, take your time.

ChrisTitusTech (approved):

CodingWonders:
I've recently noticed an issue with the modified unattended answer file, which causes the installation image to never install properly due to Setup being unable to deserialize the file:

Windows 11 (MicroWin)-2024-03-22-22-54-55

I guess that Setup is failing to parse the unattended answer file because it uses UTF-16 LE encoding, whereas the installer expects UTF-8 with BOM:

Windows 11 (MicroWin)-2024-03-22-22-51-47

The workaround is that, after having applied the image, you boot to the installation medium again and delete the answer files from both `Panther` and `sysprep`, like this:

Windows 11 (MicroWin)-2024-03-22-22-57-10

The installation should then finish successfully. I'll investigate this shortly and create a PR containing the appropriate fixes

PR #1596, opened 2024-02-09
tweaks and features json files are cleaned and toggle functions are integrated to json file

ogfrm:
- the order, category and panel fields are removed from tweaks and features json files.  the order is same as order in json file.
- tweak section in Invoke-WinUtilCurrentSystem function is separated to different Invoke-WinUtilCurrentSystemtweak function. This way it will not continue to check other parameters if it is false. It will scan faster


ChrisTitusTech:
Really liking what you are doing with this PR. I'm looking at different refactors to make the code more efficient. Right now it is growing to fast and I like a lot of your ideas here.

PR #1609, opened 2024-02-12
cleanup and more changes to inputXML ,some buttons functions moved to…

ogfrm:
- MicroWin tab is untouched and stays in inputXML file 
- tweak section in Invoke-WinUtilCurrentSystem function is separated to different Invoke-WinUtilCurrentSystemtweak function. This way it will not continue to check other parameters if it is false. It will scan faster
- Updates,Control Panel,Fixes buttons are removed from WPFButtons function and handled at json file


ChrisTitusTech (commented):
This is going to take a while to review... Damn man. Some big changes.

ChrisTitusTech:
I love that over 1000 lines of code were removed during your refactor as well. One of the big reasons I paused more application additions was because I was seeing the slowdown on launch.

ogfrm:
I found small bug. That's why i closed the PR. There should not be much slowdown on launch. I thought about it and I even implemented that generation of inputXml file to compile level but I wasn't sure if you would want that. It should improve launch time. If you want I can split this PR to smaller PR's.

PR #1619, opened 2024-02-15
change neofetch link to correct project

shrimp332:
the neofetch-win package is from the [https://github.com/nepnep39/neofetch-win](https://github.com/nepnep39/neofetch-win) project and is separate to the neofetch used on linux


CamoCatX:
That's not nearly neofetch, I would advise trying to put that as a separate package, but taking a quick look at that project, it seems like it is written in C++; opposed to shell.

ChrisTitusTech (approved):

PR #1621, opened 2024-02-15
cleanup main.ps1, inputXML generation improved. Some functions revoved and handled from json file

ogfrm:
MicroWin tab is untouched and stays in inputXML file
- tweak section in Invoke-WinUtilCurrentSystem function is separated to different Invoke-WinUtilCurrentSystemtweak function. - This way it will not continue to check other parameters if it is false. It will scan faster
- Updates,Control Panel,Fixes,tweak buttons are removed from WPFButtons function and handled at json file


KonTy (changes requested):
@ogfrm I love your enthusiasm and you have a lot of great ideas. I loved how you thought about moving injection code into compile script etc. I think what would be even more cool in addition to the changes that you are making think of minimizing XAML in your powershell code.  Honestly the for loops here are really hard to read and as a result hard to maintain. 

My suggestion would be to simplify think of a stretegy of having minimal necessary amount of code in powershell and moving everything into XAML, for example margins and other style code can be done as a XAML style. This way if you need to move, rearrage etc it can still be done in XAML without having to dive in, into powershell.

Another big thing to consider is, now that you are moving some of the logic to compile.ps1 so loading speeds are non issue, **and** you know that we have 5 panels, for example, in the apps section, you can now rethink and leave even mode XAML tags in XAML, tags like StackPanel etc. 

Basically my suggestion have as few XAML code as possible in powershell for example in Apps you could probably get away with <label><textblock><taxtblock> for label, action and help link

Same thing for tweaks it has to be possible to minimize amount of XAML in the code.

ogfrm:
I am sorry that I couldn't make the code more readable. Since I have to loop every json file and then every item in the json file I had to use loops. 

What I was thinking,  moving xaml codes to powershell is to standardize the apps, tweaks... tabs. make margins and styles of items like checkbox, buttons... same at every tab. That way you don't have to edit xaml or powershell. If you want to add item to the tab you just add it to json file. It will be added like any other item. Since there are different types of items I don't see a way to clean powershell site of margins and  styles. I may be wrong. The way I do it here is If you want to add tab like tweaks you just add it to tabs.json file and create json file for that tab items. You do not change ps1 files or xaml file. But if you want a different style tab like microWin tab, you do it the old way. edit xaml file and create functions like you normally do. 

I may be making things harder to manage. If I understand you correctly about moving some part of xaml back to xaml file, I don't see a way with my way of automation. If that is the case, PR can be rejected and can be made differently. 

Thanks for the input

KonTy:
@ogfrm you have some great ideas :) thank you for all the contributions. I think, there are a couple of things that can be changed here that will make it great. I wish we could discuss it and have a high bandwidth conversation as oppose to writing back and forth. :)

ChrisTitusTech:
Love a lot of this work, but I am currently refactoring the entire project. The main issue I have with moving more into JSON is it slows the script down because of the dynamic XAML generation. Check out my last two commits where I take out your previous additions in main.ps1 and move all the dynamic generation to the compile.ps1 file so it no longer taxes the user launching the script.

PR #1625, opened 2024-02-16
Add 'Run as administrator' to 'Invoke-WPFShortcut' Function

og-mrk:


og-mrk:
## Quick Note
For anyone interested as to why it works this way (changing bits in a specific Byte of the .lnk file), it's because there isn't any COM interfaces to change this value, and when I look it up I've found [this solution from Stack overflow](https://stackoverflow.com/a/29002207) which I borrowed to make this feature possible.

## Screenshots showcasing the difference before (on the left) & after (on the right)
Screenshot

og-mrk:
so far... I don't think the Byte Position & value has changed since windows 7 (I think even before that, maybe Windows XP or 2000!), but I can't tell for sure other than this trick is known and used since at least 2015 (The Stack Overflow question year post).

Anyhow, if there's anything out of the ordinary or suggests you may have, feel free to point them out and I'll update my commit as soon as possible 👍

og-mrk:
Also just a quick question, Should I push the compiled `winutil.ps1` script with my current commit changes?
I'm kind of new to this project, and I've used & read the source code a bit, but don't know the exact workflow for committing to this repo, so I'm asking just in case.

Thanks in advance

ChrisTitusTech:
No it gets auto compiled on commit through a github action. I'm currently in the middle of a big refactor so I haven't accepted any PRs. I like this addition fixing the shortcut though.

ChrisTitusTech (approved):

PR #1630, opened 2024-02-18
Add "take ownership" context menu

Yuuh15:
This is not working, and I don't have enough experience to troubleshoot, so I would like some help


Yuuh15:
Made use of reg command for it to work. I would appreciate if anyone could explain so I can avoid the same issue in the future

ChrisTitusTech:
This should have been done with a Powershell command via Set-Property instead of invoke script using reg add, even then this is a bit of an obscure tweak that I'd prefer people to use Shell program to add it instead of a direct edit like this.

Issue #1441, opened 2024-01-22
【Optimization】The Project is too resource intensive on low powered PC.


blusewill:
**Is your feature request related to a problem? Please describe.**
So I am installing Windows 10 LTSC into a QEMU VM which is 2 cores 2 thread (Very low powered PC) and found out that Toolbox in the dark mode have a very intensive resource usage (Might because of glowing effect). Makes CPU goes up to 80% makes the entire PC slow down.

**Describe the solution you'd like**
Maybe what you can do is making the glowing color not resource intensive? or disable it completely in some low-powered machine when detected. Or back to classic using CLI interface.

**Additional context**
Image of Showcasing VM is being resource intensive.
![image](https://github.com/ChrisTitusTech/winutil/assets/62820836/6fa14e7f-50b8-4d70-bf5d-1fe20401ecc9)

martin-braun:
Same problem. I try to setup a Windows VM with sane defaults and this tool is quite a hog. I assume the cool looking design is at fault. Last time I checked this project, it looked like a Windows XP app, but it was much faster. Sadly unusable right now.

Please revert those design changes or allow to provide a flag to launch it without fancy UI.

berlintay:
Could possibly pass a config.json file option with the util.

example

```
pwsh.exe -Command "irm https://christitus.com/win | iex" -ConfigFile ~\config.json -NonInteractive"

```


i dont think it will work just like that you might have to create a script but it could help.

martin-braun:
Just because the issue is inactive doesn't mean the issue doesn't exist.

v-Kaefer:
I'm so glad Chris set up a bot to close issues, finally. @martin-braun feel free to open another issue, but i'm pretty sure, it's because of the big list being loaded, right from the start. I'm working on sub-lists, that may end up solving that aswell.

blusewill:
Alright this issue is closed due to inactivity it's an optimization issue that still haven't done yet

PR #1643, opened 2024-02-22
Change the Theme Configuration for low spec.

blusewill:
I modified the Theme in Dark Mode since is Resource intensive
It makes the problem in #1441 and #1423 better by disabling the glow color when detected only on board graphics
Which is the File  `Invoke-WinUtilGPU.ps1` does.
Checking for the GPU is detected or not
If not it will fallback to the Lite Dark Theme which is specific in the `theme.json`
There is no point to change the Classic Theme due to is already lite enough that won't uses many process.


KonTy:
This is not a PowerShell issue, this is literally low spec issue :) What I did is buried an easter egg in the code where the glow was pulsating, I wanted to remove it but forgot to do it. Since on my machine I don't notice any difference. @blusewill there is a much easier way to disable the shadow pulse. In the theme.json there is a variable called
``
"ShadowPulse": "0:0:3"
``

if you change it "Forever" it will stop pulsing and taxing your machine.

blusewill:
@KonTy Well Yes if you check the code I make another Theme called Dark. 
But I don't want to disable the glow color so I decided to make a detection of the Graphic driver is done or not.
And the other configuration in the theme.json is copied from classic theme due to classic theme is light enough.

ChrisTitusTech (approved):

bdebates:
This is an issue! I have a quadro rtx 4000 and it is using 70% of my GPU.  When I tried searching for GPU related issues I came across issue 1423. Someone in that issue had their 6950XT was sitting at almost 50%. Another person had their 3070 mobile sitting at 34%. This issue was closed due to inactivity. Now that i finally was able to track this down, I see there are a couple of issues related to this. 

This was all because _one_ person (@KonTy ) wanted pulsating text and snuck a huge performance bug into the code as an "Easter Egg"? 

This is ridiculous and needs to either be completely removed or toggle able, because our measly GPUs pass your "check" when they aren't good enough for your pulsating text.

blusewill:
In the future update might have a toggle for the Dark Mode or Matrix.
But for now you can Download Winutil and Modify the Code it self also would work.

PR #1649, opened 2024-02-23
inputXaml generation moved to compile.ps1 and improved. some function…

ogfrm:
- I am confused about your changes because I think I would break you one file run system. I changed my PR to move inputXAML generation to Compile.ps1. It just creates winutil.ps1. I kept the other changes. I know you are factoring but It would have been harder for me to work on old functions.
  MicroWin tab is untouched and stays in inputXML file
- tweak section in Invoke-WinUtilCurrentSystem function is separated to different Invoke-WinUtilCurrentSystemtweak function. - This way it will not continue to check other parameters if it is false. It will scan faster
- Updates,Control Panel,Fixes,tweak buttons are removed from WPFButtons function and handled at json file


ChrisTitusTech:
This PR is just doing way too much for me to accept. Moving things from XAML to JSON just makes it harder to read and manage. The reason for my last two commits were so the user doesn't have any dynamic XML generation on launch as it slows it down. The idea is to simplify and consolidate the code with the refactor. I can't make heads or tails of what this PR is doing having to look at JSON files that get brought into the code.

PR #1665, opened 2024-02-27
Fix & Add Apps

v-Kaefer:
- Fix Vivaldi winget
- Add KiCad
- Add Cryptomator


blusewill:
This will be closed right now due to Chris is refactoring the project. No apps should be added.

But will be nice to change the Vavaldi Package

Issue #1644, opened 2024-02-22
Winget Woes


PatrickJnr:
Winget works fine on my machine, but when trying to use it using this tool, I get

![image](https://github.com/ChrisTitusTech/winutil/assets/3267523/46e90861-8a1d-4af2-b21b-77907f04dbd0)

When trying to do a reinstall of Winget  via the tool I get this
![image](https://github.com/ChrisTitusTech/winutil/assets/3267523/1136e882-3700-4d9a-bf27-dc47142fd237)

This is a fresh installation of Windows 11 with the latest patches.

winget -v shows v1.7.3481-preview

I manually updated to the latest via here: https://github.com/microsoft/winget-cli/releases

now it shows v1.7.10514 and winutil seems to see it now.

og-mrk:
This issue is related to chocolatey package manager, rather than a WinUtil issue.
Feel free to post an issue over their community packages repo on Github, to see what can be done to fix this error.

**Note**: Please share the log file for choco (Not Winutil Log File) so people over there can actually help out, it'll be a huge help for them on Diagnosing the source of them issue.

MohamedMehrath:
I got the same problem.

PR #1682, opened 2024-03-06
Handle preview versions of winget

Marterich:
Handle "Winget not detected" message if a preview version of winget is installed that meets the version requirenments. 

Currently you cannot install any apps with winutil if you have a preview version of winget installed. The PR addresses the "-preview" suffix of the version number and prints a yellow message to the console, warning the user that a preview version is detected and unexpected problems may occur.

image

Reference to an example winget preview version:
- https://github.com/microsoft/winget-cli/releases/tag/v1.8.532-preview

Edit:
This PR should also fix the problem described in Issue #1644 Winget Woes where winget is not detected due to the preview version being installed on a fresh W11 install


functions/private/Test-WinUtilPackageManager.ps1
```diff
@@ -20,11 +20,17 @@ function Test-WinUtilPackageManager {
     # Install Winget if not detected
     $wingetExists = Get-Command -Name winget -ErrorAction SilentlyContinue
     if ($wingetExists) {
-        $wingetVersion = [System.Version]::Parse((winget --version).Trim('v'))
+        $wingetversionfull = (winget --version)
```
og-mrk, line 23:
```suggestion
        $wingetversionfull = Get-WinGetVersion
```

You can use the PowerShell Cmdlet that WinGet Provides, which of course from Microsoft themselves (Source of Cmdlet(s) is `Microsoft.WinGet.Client`), may be using Cmdlet is better than saving output of command into a variable, don't really know if that's true or not, it could be personal preference in cases like these.

anyhow, good commit @Marterich, Thanks for your contribution 😄

Marterich: TLDR: I dont think the possible problems that this change could induce justify the means to use the Cmdlet over the raw application call.

First of all, Thanks for the Idea, 
I would nevertheless keep the version call raw, without using the Cmdlet. I see the point in using the Microsoft Provided functions (in case they change the inner workings of Winget later on), but it poses some difficulty as well to use it here. 
If you use the "raw" binary argument without the cmdlet, suffices if the binary is in your $PATH and can be executed by the user. 
If you however want to use the Cmdlet "Get-WinGetVersion" you have to make sure that the user has the  Module `Microsoft.WinGet.Client` installed as well before checking the version, which isn't necessarily needed for running Winget in itself. In case the Module is not installed, yet, you would then need to handle the installation, which would possibly require setting up a NuGet Provider, allowing PSAppGallery to download the module and then installing it.
So, this would require quite a lot more changes to the inner workings and error handling of winutil which seems to me a bit much to query the version. 

Nevertheless, the use of the Module `Microsoft.WinGet.Client` could be taken into consideration for for examle Listing the installed packages and so on, because it returns actual CustomPSObjects for Name, Id, Version etc. which would remidiate the need for the error prone string manipulation that is used at the moment to parse the output of the binary and possibly other tasks as well. 
However i think this would definetly need to be handled in another PR or a Discussion Thread to check if it is worth it to integrate the Module as a dependency into the project.

ChrisTitusTech (approved):
Very cool love the PR. I'm about to rewrite the winget install on the new 23H2.

PR #1722, opened 2024-03-22
Added "Betterbird" to Communication section

Today20092:
[Betterbird ](https://www.betterbird.eu/)is a fork of [Thunderbird](https://www.thunderbird.net/en-US/)

> Betterbird is a fine-tuned version of [Mozilla Thunderbird](https://www.thunderbird.net/), Thunderbird on steroids, if you will.

> Betterbird is better than Thunderbird in three ways: It contains new features exclusive to Betterbird, it contains bug fixes exclusive to Betterbird and it contains fixes that Thunderbird may ship at a later stage. Please refer to this [feature table](https://www.betterbird.eu/#featuretable) for examples. This should give you an impression of where the project is headed.

> Easy adoption: You can install Betterbird at the same time as Thunderbird and run them on the same profile. That means that you can try out Betterbird with zero hassle, and go back to Thunderbird if you don't like it - which is unlikely. Please read the fine print on our [support page](https://www.betterbird.eu/support/index.html#switch-tb-bb)

why use betterbird

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -127,6 +127,14 @@
 		"link": "https://github.com/sharkdp/bat",
 		"winget": "sharkdp.bat"
 	},
+	"WPFInstallbetterbird": {
+		"category": "Communications",
+		"choco": "betterbird",
+		"content": "Betterbird",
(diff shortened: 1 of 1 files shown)


og-mrk (approved):
Looks good 👍
The Syntax and everything else is correct, good commit @Today20092

PR #1724, opened 2024-03-23
Fix unattended answer file woes for MicroWin

CodingWonders:
This is a fix to the issue in PR #1595, where Windows would never complete installation due to an encoding mismatch of the XML file, as it was being saved with UTF-16 LE encoding.

Windows 11 (MicroWin)-2024-03-22-22-54-55

Windows 11 (MicroWin)-2024-03-22-22-51-47

This change forces the target file to be saved with UTF-8 encoding, and now it's imported correctly on Windows SIM. The same thing should happen with Windows Setup.


ChrisTitusTech (approved):

ChrisTitusTech:
Thanks @CodingWonders I'll get this merged asap

PR #1384, opened 2024-01-12
Test 2024 01 03

ChrisTitusTech:
rework of unit tests... after 1billion commits lol. 
Program install details by @KonTy 
more microwin fixes by @CodingWonders

PR #1727, opened 2024-03-26
Fixed Ventoy drive copy typo and began conversion of DISM commands into DISM cmdlets

CodingWonders:
This PR fixes a typo in the Ventoy drive copy function spotted by @JoKopii, on PR #1384 (!); and also changes some DISM commands with appropriate cmdlets from the DISM module.

The cmdlets provide more output to the user, showing progress information. The following was updated:

- Image mount operation
- Image unmount operation
- Image export operation

The cmdlets work the same way as the DISM commands.


NewCodingAI (approved):

og-mrk (changes requested):
The changes in this commit looks fine, but the tabs & spaces are mixed, fix in a new commit, or change it and force-push the same commit changes but with the tabs/spaces issue resolved.

Overall good commits @CodingWonders, _and IK what I've previously stated is somewhat nitpicking_, but it's better to point these stuff out if, for example, you haven't noticed it while writing these changes.

CodingWonders:
> The changes in this commit looks fine, but the tabs & spaces are mixed, fix in a new commit, or change it and force-push the same commit changes but with the tabs/spaces issue resolved.
> 
> Overall good commits @CodingWonders, _and IK what I've previously stated is somewhat nitpicking_, but it's better to point these stuff out if, for example, you haven't noticed it while writing these changes.

The PS ISE causes the indentation issue - it never gets it right, and I can't just use VSCode because then the compilation script fails on me (stating that `winutil.ps1` is in use).

At least PowerShell is not like Python, where indentation matters

og-mrk:
> > The changes in this commit looks fine, but the tabs & spaces are mixed, fix in a new commit, or change it and force-push the same commit changes but with the tabs/spaces issue resolved.
> > Overall good commits @CodingWonders, _and IK what I've previously stated is somewhat nitpicking_, but it's better to point these stuff out if, for example, you haven't noticed it while writing these changes.
> 
> The PS ISE causes the indentation issue - it never gets it right, and I can't just use VSCode because then the compilation script fails on me (stating that `winutil.ps1` is in use).
> 
> At least PowerShell is not like Python, where indentation matters

Interesting problem you got there... well, good thing PowerShell doesn't depend on Indentation for scoping. Hope you find a solution for it.

ChrisTitusTech (approved):

Issue #1181, opened 2023-11-15
"Upgrade All" function is not working


proteuteu:
**Describe the bug**
Package agreements were not agreed to. Operation cancelled.

**To Reproduce**
1. Go to 'Install'
2. Click on 'Upgrade All'
3. Scroll down to bottom

**Expected behavior**
All apps get updated one by one

**Screenshots**
![image](https://github.com/ChrisTitusTech/winutil/assets/40414056/9ffb0c64-afc3-4bf3-9e55-01f58ef2db4d)

**Additional context**
I looked over #1030 and followed both OP and anonyao instructions + manually download and installed winget from [github](https://github.com/microsoft/winget-cli) but to no avail.

PR #1730, opened 2024-03-26
Fixes Update.

blusewill:
This fixes the Update issues in #1181 
Which is some function needs to be changed in the `Update-WinUtilProgramWinget.ps1`


ruxunderscore:
I know that there is an issue I ran into with my changes in my Pull Request in regards to installing Winget. Non-UWP apps refuse to install when scope is defined as machine (i.e. `--scope=machine`).

ChrisTitusTech (approved):
Thanks Blue!

PR #1732, opened 2024-03-26
Change order for easier launch in README and fix outdated info

ModernTTY:
By default the “easy way” in README is the long `iwr-useb`.
I changed the order so that the `irm` version is first.

Another thing I fixed is that in README.md in the Overview section there was left pretty old, outdated info.


CodingWonders (approved):
Even though both commands essentially do the same thing, the `irm` one can be easier for users to understand

CodingWonders (approved):

og-mrk:
Looks good to me, using `irm` instead of `iwr -useb` is easier to remember like @CodingWonders mentioned.
The updated docs are pretty accurate as well, good commit(s) @ModernTTY 👍

og-mrk (approved):

ChrisTitusTech (approved):
Agreed irm is easier and I use it all the time.

PR #1734, opened 2024-03-26
Remove Auto-Generated WinUtil file from script, and fix it to work with gitignore.

CoryBorek:
As mentioned in the stream, it would be nice to have the autogenerated script to work with the gitignore. The file was added to the repo before the gitignore, so it would still work. This should allow it to be ignored.


ChrisTitusTech:
Can't actually merge this, but I get what you are saying. I'll merge this today.

Issue #374, opened 2022-11-11
Program Additions for next update


ChrisTitusTech:
- [x] SuperF4 - https://stefansundin.github.io/superf4/
- [x] Choco
- [x] Scoop
- [x] Ventoy
- [x] dotnet 7
- [x] dotnet 4.8
- [x] directx
- [x] winget install --id=KDE.Okular -e
- [x] winget install --id=KDE.Kate -e
- [x] winget install --id=KDE.Kdenlive -e
- [x] winget install --id=KDE.KDEConnect -e
- [ ] winget install --id=KDE.digikam -e
- [ ] winget install --id=KDE.Krita -e
- [ ] winget install -e --id Docker.DockerDesktop
- [x] winget install HeroicGamesLauncher.HeroicGamesLauncher
- [ ] yt-dlp.yt-dlp
- [ ] FxSoundLLC.FxSound
- [ ] Automattic.Simplenote
- [ ] winget install Flow-Launcher.Flow-Launcher
- [ ] winget install Kitware.Cmake
- [ ] winget install gerardog.gsudo
- [ ] winget install MuseScore.MuseScore
- [ ] JPEGView 
- [x] MusicBee
- [x] ModernFlyouts 
- [ ] Oracle.VirtualBox
- [x] Safing.Portmaster
- [ ] Giorgiotani.Peazip
- [ ] valinet.ExplorerPatcher
- [ ] CoreyButler.NVMforWindows Node version manager
- [ ] scrcpy

Expanding Tool to include other CLI installers. Winget is being unpredictable in 22H2 and I want to start diversifying the installs, while expanding its capabilities.

ChrisTitusTech:
Request per a comment from - Stefan W:
Lockhunter: https://lockhunter.com/
Rename Bulk Utility: https://www.bulkrenameutility.co.uk/ (PowerToys has a similar, harder-to-use tool)
Philips Hue Sync app: https://www.philips-hue.com/en-sg/explore-hue/propositions/entertainment/hue-sync
Start10/Start11 (but that's a paid program), don't know if you'd want that.
AutoHideMouseCursor: https://www.softwareok.com/?seite=Microsoft/AutoHideMouseCursor
Disable Nvidia Telemetry (don't know if this is what your toolbox already does, but this is a program to disable, but I have also used PowerShell to remove the nvidia one, as it takes up a lot of space.): https://github.com/NateShoffner/Disable-Nvidia-Telemetry
[...]

PR #1747, opened 2024-03-28
Add “a few ”apps

ModernTTY:
I added most of the apps mentioned in issue #374.

All other apps that I didn’t add aren’t available in WinGet so obviously cannot be added.

List of apps:

- FxSound
- SimpleNote
- Flow launcher
- CMake
- MuseScore
- JPEGView
- Explorer patcher
- Lockhunter
- Bulk Rename Utility (BRU)
- SpaceSniffer
- Quick Look
- Harmonoid
- Revolt
- Link shell extensions
- WizFile
- HWMonitor
- XnView
- Tabby
- SignalRGB
- Autoruns

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -1735,6 +1735,14 @@
 		"link": "https://www.spacedrive.com/",
 		"winget": "spacedrive.Spacedrive"
 	},
+	"WPFInstallspacesniffer": {
+		"category": "Utilities",
+		"choco": "spacesniffer",
+		"content": "SpaceSniffer",
+		"description": "A tool application that lets you understand how folders and files are structured on your disks",
+		"link": "http://www.uderzo.it/main_products/space_sniffer/",
+		"winget": "UderzoSoftware.SpaceSniffer"
+	},
 	"WPFInstallstarship": {
 		"category": "Development",
 		"choco": "starship",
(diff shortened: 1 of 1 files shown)


og-mrk:
> All other apps that I didn’t add aren’t available in WinGet so obviously cannot be added.

Hello @ModernTTY
May I ask what are these apps you haven't found in WinGet Package Repo, I can try and add them into WinGet?

adding programs is similar to WinUtil, done in this repo, but with YAML formatting instead of json, and _a bit more_ stricter rules and tests for every single app you're trying to add, due to the fact no one wants to have Malware in their Package Manager, right? 😆

ModernTTY:
@og-mrk In winget.run is was not able to find these apps:

- VistaSwitcher
- MPV
- Uplay
- PDFGear
- Windows Firewall Control (maybe it is I’m not sure because the winget.run search is terrible.)
- Unigram(it is available as a MS store app, but that will cause issues with people who don’t use a Microsoft account/the Microsoft store. Should be as a native.)
- Tixati
- MPO GPU fix (https://github.com/RedDot-3ND7355/MPO-GPU-FIX)
- OrcaSlicer
- CapFrameX
- ZenTimings

Issue #1736, opened 2024-03-27
Get-WmiObject not recognized in Powershell 7. 


ruxunderscore:
**Describe the bug**
I was testing the script in a Windows 10 22H2 Testing VM. I had been using Powershell 5.1 (default to Windows 10 22H2) and decided to test in Powershell 7 to see if I got the same result, and got this error that I hadn't seen in Powershell 5.1.

**To Reproduce**
1. Run the winutil.ps1
2. Error appears for $gpuInfo

**Expected behavior**
Expected no error.

**Screenshots**
![image](https://github.com/ChrisTitusTech/winutil/assets/16318553/bf682cd9-fe33-4f17-8455-0c8e818794df)

**Additional context**
Hitting enter after the error allows the script to continue, seemly unharmed. This was while testing in a Windows VM. 
- Windows 10 (22H2) Fresh Install
- Been working on a couple changes in the Winget Installer side of things, when I came across this bug.
- I believe I saw the $gpuInfo part added as a pull request by Chris recently. 
- If I try to call Get-WmiObject outside of the script, it isn't recognized in Powershell 7.

ruxunderscore:
After a little research, I may suggest using Get-CimInstance as a fallback.  `Get-CimInstance -query "SELECT * from Win32_VideoController"` will return similar information.

ruxunderscore:
The Current commit wasn't a full fix, it references the incorrect class. @YusufKhalifadev is working on a fix, then will close after merged.

YusufKhalifadev:
made a pr about this

PR #1752, opened 2024-03-28
Test 2024 03 21

YusufKhalifadev:
Misspelled Win32_OperatingSystem instead of Win32_VideoController


ruxunderscore:
Awesome. Adding this as a reference to #1736.

ruxunderscore:
This is being changed in the winget.ps1 file, not the source script. I will work on a fix.

ruxunderscore:
#1753 Updated to the correct file.

PR #1757, opened 2024-03-29
Update of Test-WinUtilPackageManager and Install-WinUtilWinget

ruxunderscore:
## Update of Test-WinUtilPackageManager and Install-WinUtilWinget. 
I created this Pull Request to change the default behavior of Winget's installation method. This was so that by default it would install using the GitHub method unless failing—failing over to using Chocolatey. 

Also changed Test-WinUtilPackageManager to return more verbose responses, then implemented it's use in various Functions that were previously using `Get-Command`. This moves multiple uses of the same code to one location. Added Color to beautify the terminal output.

Split into two commits.

Tested in Windows 10 (22H2)

### Commit 1
- Updated Test-WinUtilPackageManager to better handle the -Winget param and return a more verbose status.
- Moved many of the "is installed"/"is not installed" outputs to the Test-WinUtilPackageManager function.
- Changed Install-WinUtilWinget to use the GitHub install method as the primary method, and on error use the Chocolatey install method.
- Updated various functions to use the Test-WinUtilPackageManager function.
### Commit 2
- Changed handling of Test-WinUtilPackageManager in Install-WinUtilWinget, to prevent Test-WinUtilPackageManager from printing out to terminal twice.

### Finishing Thoughts
I'll stop contributing for a bit and get some sleep now. If anyone has some suggestions, or can think of better ways to handle some of these things, let me know.

### DIFF
--- a/functions/private/Install-WinUtilChoco.ps1
+++ b/functions/private/Install-WinUtilChoco.ps1
@@ -10,20 +10,19 @@ function Install-WinUtilChoco {
     try {
         Write-Host "Checking if Chocolatey is Installed..."
 
-        if((Get-Command -Name choco -ErrorAction Ignore)) {
-            Write-Host "Chocolatey Already Installed"
+        if((Test-WinUtilPackageManager -choco) -eq "installed") {
             return
         }
 
-        Write-Host "Seems Chocolatey is not installed, installing now"
+        Write-Host "Seems Chocolatey is not installed, installing now."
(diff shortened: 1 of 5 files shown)


ChrisTitusTech (approved):
Dig the changes the subtle color change from not installed to installed is nice. Thank you Rux.

ruxunderscore:
> Dig the changes the subtle color change from not installed to installed is nice. Thank you Rux.

Love adding a little color to spice things up.

PR #1760, opened 2024-03-29
Add Simple Feature to keep the Service Startup upon Applying Service Tweaks, but not when Undoing it

### LINKED ISSUE #1726: Don't re-enable disabled services
**Is your feature request related to a problem? Please describe.**
Whenever I disable some services using NTLite, this script (or windows itself, not sure) sets some services from Disabled to Manual which is not desired behaviour.

**Describe the solution you'd like**
An option to not touch disabled services

**Describe alternatives you've considered**

**Additional context**

og-mrk:
### Commit Message Body
Added a new parameter that gives freedom of control on whether to disable this feature or not, and of course the simple feature in question.

The way it works is by Getting the service using its name, and see if the Startup Value of this service is equal to the default type that Windows comes with it, if not (The User has changed it in the past), then WinUtil won't change it by default (The KeepServiceStartup is true by default), this is a more desirable behaviour when compared to how it previously worked.

These changes were tested by the Author of this commit, Please read the commit patches for exact details on the changes.

### Issues this PR tries to Resolve
Resolves #1726

### DIFF
--- a/functions/private/Invoke-WinUtilTweaks.ps1
+++ b/functions/private/Invoke-WinUtilTweaks.ps1
@@ -10,11 +10,15 @@ function Invoke-WinUtilTweaks {
     .PARAMETER undo
         Indicates whether to undo the operation contained in the checkbox
 
+    .PARAMETER KeepServiceStartup
+        Indicates whether to override the startup of a service with the one given from WinUtil,
+        or to keep the startup of said service, if it was changed by the user, or another program, from its default value.
     #>
 
     param(
         $CheckBox,
-        $undo = $false
+        $undo = $false,
+        $KeepServiceStartup = $true
     )
 
     Write-Debug "Tweaks: $($CheckBox)"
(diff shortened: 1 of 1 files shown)


og-mrk:
### Notes About this PR & Last Force-Pushed Commit
Added a warning in the last force-pushed commit, about why I've added `!($undo)`, which will make this feature only work when Applying Service Tweaks, but not when Undoing it, in other words, Clicking `Undo Selected Tweaks` when you have `Set Services to Manual` Selected, will set the startup type to whatever the ones specified in `tweaks.json` file.

ChrisTitusTech:
Also make sure you are submitting to test branch not main

og-mrk:
> Might get too talkative in prompt, but I'm going to test it some more and see. 

I do agree on the over wording in debug messages, _probably because I was half asleep when I wrote this_, did this to make it easier on me when testing.

Shortening (if this's even a word) the debug message will be a good idea. 👍

PR #1766, opened 2024-03-30
Sacrifice to the AV Gods test 24-03-28 branch

ChrisTitusTech:
- Removal of Self Elevation
- Removal of Disable UAC
- Check if service setting has already been adjusted @og-mrk 
- Update Choco fallback and Winutil fixes @ruxunderscore 
- Program fixes description and incorrect program labels @YusufKhalifadev 
- Cleanup free space on fresh MicroWin run @CodingWonders


blusewill:
So Microsoft Does track the hashes and update to the Database

CodingWonders:
> So Microsoft Does track the hashes and update to the Database

Allowing the "risk" works for now, until Microsoft decides to flag the script with a different threat

blusewill:
The Sacrifice does work I guess?
Only one AV having False Positive 
image

ruxunderscore:
I figured it was a hash based detection, those are usually the ones that get the false positives. Heuristics can as well, but only if you are doing something that generally shouldn't be done, or is a known mechanism used in malware. (I am probably missing some details, but that's about what I recall while doing my Firewalls course in college.)

PR #1783, opened 2024-03-31
Re-Opening of PR #1747

og-mrk:
Credits for the commits goes to @ModernTTY
and I've made sure that ModernTTY is the Author of every commit he made, and the Date stays the same as well.

Only did a small contribution by removing a Trailing Tab in commit "Add autoruns"


og-mrk:
Will close this PR, noticed something odd in the commits, will make a new PR for the same purpose.

PR #1801, opened 2024-04-04
Update Github Actions Auto Close Issues

og-mrk:
### PR Changes & Why Making them in The First Place
The changes in this PR will get all Labels of every Open Issue, and checks if Label Name _**equals**_ a name that we're expecting, I've put it as "Keep Issue Open", but it can changed easily.

Of course the Repo needs a new Label with the Same Name as the one above, which can only be done by the Admins and/or Maintainers who have the Right Privileges to do so.

The why part is to make sure one or more Issues stay open without letting The Auto Closing Workflow close the issue after a period of inactivity have passed.


ChrisTitusTech (approved):
Thanks OG, I'll get this merged.

PR #1826, opened 2024-04-12
Added OperaGX to browsers 

oliverpatrick:


ModernTTY:
@oliverpatrick Chris usually does not allow any version of Opera to be in Winutil, PR will probably be denied. Also it’s going to the wrong branch.(should go to test branch)

brrock:
> @oliverpatrick Chris usually does not allow any version of Opera to be in Winutil, PR will probably be denied. Also it’s going to the wrong branch.(should go to test branch)

I agree never will be accepted

ChrisTitusTech:
Yeah, I will NEVER accept any PR with Opera or an Opera based browser.

PR #1827, opened 2024-04-12
Add Tweak to disable Microsoft Account Login

lCTdE:
A suggestion to disable Microsoft Account login for Windows Accounts. The idea behind this is to stop Windows from showing those blocking screens at login and pop-us in the UI that push local account users to switch to a Microsoft Login Account.
At work this is considered a security risk, so normally I disable it using group policy, but for this tweak I decided to use a registry method so those poor souls using Windows 10/11 Home can benefit too.

This is the source I used: https://www.elevenforum.com/t/enable-or-disable-microsoft-accounts-in-windows-11.23799/

In the tweak I used value "1" instead of "3" to avoid possible locks if a Microsoft Account is already in use and because this allows other users to use Microsoft Accounts before applying the tweak.
In the undo tweak part, as I didn't found a way in the tool to delete a registry key like the article suggests, I opted instead to use the value "0".  I tested this on a VM and it seems to work.

Comments on the name, order and description of the tweak are welcomed, as I lack a bit of imagination and this is my first time contributing to this tool. I this pull request is denied, I will completely understand it.

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2655,5 +2655,21 @@
     "panel": "2",
     "Order": "a082_",
     "Type": "300"
+  },
+  "WPFTweaksDisableMicrosoftAccountLogin": {
+    "Content": "Disable Microsoft Account Login",
+    "Description": "Disables new logins into Microsoft Accounts, so you can use local accounts in peace.",
+    "category": "z__Advanced Tweaks - CAUTION",
+    "panel": "1",
+    "Order": "a023_",
+    "registry": [
+      {
+        "Path": "HKLM:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Policies\\System",
+        "Name": "NoConnectedUser",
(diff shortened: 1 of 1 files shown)


SkylerWallace:
I really dislike using group policies as a way to disable features, it feels like a nuclear option.

There's a better option available and there's even an article on ElevenForums for this too:
https://www.elevenforum.com/t/enable-or-disable-lets-finish-setting-up-your-device-in-windows-11.5205/

lCTdE:
@SkylerWallace  Nice idea, thanks! I will think about it and make a commit with a more mature idea. Right know I'm debating if I should implement only one of them or both and how.

lCTdE:
I decided to add the suggested approach as part of the Disable Telemetry Tweak in PR #1832 
For my original approach, I'm still not sure if it is still interesting to keep it as a Tweak, move it to Config or just close the PR.

Issue #1822, opened 2024-04-10
O&O ShutUp10++ shows error when together with Delete Temporary Files


DiegcD:
**Describe the bug**
When running O&O ShutUp10++ by itself no errors occur but the moment you select it with Delete Temporary Files an error appears on screen and I don't know if the tweaks of ShutUp10++ applies or not, I think it deletes the .cfg config before running O&O on quiet mode.

**To Reproduce**
1. Go to Tweaks Tab
2. Click on Run OO Shutup
3. Click on Delete Temporary Files
4. Click on Run Tweaks
5. Error appears

**Expected behavior**
An O&O ShutUp++ error should not appear.

**Screenshots**
Can't attach due a network error on my end

**Additional context**
Windows 11 updated to today, Spanish MX version

ModernTTY:
> an error appears on screen
Could you copy and paste the error that you got, because without it it is impossible to understand what is going wrong.

DiegcD:
![Captura de pantalla (49)](https://github.com/ChrisTitusTech/winutil/assets/102642718/517f21fa-2d36-404e-8f50-ebc0f8ed6331)

Marterich:
Looks like you encountered a race condition. 
Until now the execution for OO Shutup 10 looked as follows. 

Download EXE -> Download Config -> Run EXE with Config asynchronously
|---files stored in temp folder---------| 

Because the EXE was run asynchronously it technically could happen, that the temp file cleanup task deleted the config file before OO Shutup could load it properly. 

I revamped the OO Shutup logic anyway in #1828 and have also modified the code, so the winutil will now wait until OO Shutup is done applying its changes before continuing on.

Issue #1813, opened 2024-04-07
Re-enable location services


MrGamerDoesGames:
**Is your feature request related to a problem? Please describe.**
It'd be nice to have a button or toggle to reenable the location services. There are legitimate uses for this kind of telemetry, such as having night mode go based on your current location, or having the system clock work based on your location.

**Describe the solution you'd like**
Put a checkbox or button in the GUI, maybe next to the one that disables location tracking, to enable it.

Marterich:
In PR #1828 I revamped how OO Shutup is applied, updated the config file to reflect the recommendations by oo-software, and thereby removed the settings that disable location services completely by default. 
I also added the functionality to completely remove any changes by OO Shutup and added a button to start the GUI to modify the policy changes

MrGamerDoesGames:
> In PR #1828 I revamped how OO Shutup is applied, updated the config file to reflect the recommendations by oo-software, and thereby removed the settings that disable location services completely by default. 
> 
> I also added the functionality to completely remove any changes by OO Shutup and added a button to start the GUI to modify the policy changes  

Wow, perfect! Thank you! Has this already been merged or can I not do it now?

Marterich:
The PR hast not been merged yet. 
If you want to apply the changes manually, please follow the steps i outlined in issue #1810
The Procedere ist identical

Issue #1812, opened 2024-04-06
Running OO Shutup doesn't work


ZongMan:
**Describe the bug**
Running OO shutup doesnt work. 

**To Reproduce**
1. Load CTT.exe, go to tweek tab, check :Run OO shutup, Click run tweeks.
2. CMD prompt for CTT states its running script for the OO but it never loads a window for selecting what options to use. 
3. CMD then says tweaks are finished.

**Expected behavior**
Am i not supposed to get a windows to select what options to choose? does this part of the tweak do everything on its own?

**Screenshots**

**Additional context**

tchjeff:
The script checks if the necessary files to run OO Shutup are already downloaded. If not, it downloads them automatically. It then runs the program in a silent mode using a configuration file. This applies the tweaks in the background without any user interaction.

https://github.com/ChrisTitusTech/winutil/blob/4a7c8a35bf2e874579504fb81d5b3589823da7a0/config/tweaks.json#L2237

Marterich:
In #1828 I overhauled the logic of what policies are applied (or reverted) by OO Shutup and also added a Button under the advanced tweaks section to open the OOSU GUI and modify the options manually

PR #1828, opened 2024-04-13
OO Shutup 10 Tweaks revamped

Marterich:
**!!! IMPORTANT !!! Because the PR relies on downloading files directly from the Git Repo (Main), any tests will fail until the config files are moved there or the download links in "Invoke-WPFOOSU" are modified for testing**
- Created a config file that contains the recommended changes by the newest version of OO Shutup 10
- Created a config file that contains the factory settings
- Changed the location of the config files to reside in the config directory
- Added functionality to undo the tweaks by OO Shutup
- Added a button in the Advanced  Tweaks section to open the OO Shutup GUI to modify the settings manually #1812 
- Moved all of the logic for downloading and applying the tweaks from the tweaks.json to a public function "Invoke-WPFOOSU"
- Added -Wait to the execution parameters to avoid race condition when run together with temp file removal (Should fix: #1822 )

The motivation for this PR was, that I kept reading issues #1813 from people who eg. couldn't enable location services anymore after running the toolbox. Further digging resulted in a few realizations:
1. The old config file disabled location services completely via policies
2. No functionality was built into the script to reset or modify the behavior of OO Shutup 10 
3. The old configuration file was created using an older version of OOSU so not all new and shiny options were present inside.

Furthermore, I decided, that because OOSU and the config file are kind of a black box to most users, it would probably be best to rely on the default, unaltered recommendations by oo-software that do not break the functionality of apps (like disabling location services completely) and rather provide a simple way of starting the program directly if a user should decide that they want to apply some of the more intrusive optimizations.


ChrisTitusTech:
I like the changes, but at this point we probably should remove O&O from the utility as most of the changes it makes are already in the tweaks itself. Downloading the exe and config seems too cumbersome.

ChrisTitusTech:
> > I like the changes, but at this point we probably should remove O&O from the utility as most of the changes it makes are already in the tweaks itself. Downloading the exe and config seems too cumbersome.
> 
> Allow me to doubt that statement. When you say "most of the changes is makes" it implies you went through each and every setting it has both for local machine and current user recommended settings, which takes up a lot of time and effort and dedication. But that's not enough, let's say you did went through each and every setting - who said your commands cover O&O commands one-to-one? Example of what I mean by that - O&O would disable Bing search by 2 commands, one for local machine and one for current user, now how do you know what O&O did on the background in terms of which were the right commands to disable Bing search in terms of which reg files it modified? Also you have to take into consideration that O&O updates one in a while and could change it commands and/or features, someone has to keep an eye out for updates and implement them into winutil.
> 
> Look, all I'm saying is if you want to remove O&O that's fine, but make sure to implement 100 percent of it and not 60 percent and do it accurately in terms of which reg files to modify and commands to use, and most importantly - check for O&O updates and implement any new setting.

I am sure I missed something in the O&O Tweaks that they do cover. The main thought is running this on top of the tweaks is a bit overkill and I am more worried about O&O changing the recommended to something I disagree with. It hasn't happened in the past couple of years, but being less dependent on them would be an overall net win even if a few tweaks O&O makes are missed.

Marterich:
It could also be an option to remove the OO Tweak with the preconfigured config from the simple tweaks, but keep the Button in the advanced section that just simply starts the Tool. This way, it would not be neccessary to maintain the OO config files, and keep them up to date, but it would still allow the end user to start the application by the simple click of a button instead of needing to download and run the tool manually.

ChrisTitusTech:
Thanks for humoring me, I love all those suggestions. On the next update I'm going to pick these out and get it implemented. I will leave this PR open for now until I get ready to merge to main.

Issue #1824, opened 2024-04-12
Trying to choose a install like VScode and get errors


Apisit65:
By choosing an app from the list I tried with VScode and Brave to install. After clicking install these error came up.

Version: v1.3.2691
    - Winget is a release version.
    - Winget is Up to Date

Winget is already installed.
Failed in attempting to update the source: winget
Failed when searching source: winget
An unexpected error occurred while executing the command:
0x8a15000f : Data required by the source is missing

No packages were found among the working sources.
===========================================
--      Installs have finished          ---
===========================================
===========================================
--         Winget is installed          ---
===========================================
Version: v1.3.2691
    - Winget is a release version.
    - Winget is Up to Date
Getting Installed Programs...
Done...
===========================================
--         Winget is installed          ---
===========================================
Version: v1.3.2691
    - Winget is a release version.
    - Winget is Up to Date

Winget is already installed.
Failed in attempting to update the source: winget
Failed when searching source: winget
An unexpected error occurred while executing the command:
0x8a15000f : Data required by the source is missing

No packages were found among the working sources.
Failed in attempting to update the source: winget
Failed when searching source: winget
An unexpected error occurred while executing the command:
0x8a15000f : Data required by the source is missing

No packages were found among the working sources.
===========================================
--      Installs have finished          ---
===========================================

og-mrk:
Hello @Apisit65 
your WinGet Package Manager (The program WinUtil Uses, which's used to download and install programs), seems to be outdated. _If I'm not mistaken, which I might be, I think this particular version has a known bug, and it might be the version which ships with windows 10_, anyone more experience can correct me if I'm wrong.

Anyhow, Please refer to [this comment](https://github.com/ChrisTitusTech/winutil/issues/1693#issuecomment-1995937545), which explains how to re-install WinGet to the latest version. Afterward, if the issue is gone, Great! If not, then please provide more information so we can help out.

ruxunderscore:
Interesting I have a check that should have caught this as outdated. I'll have to troubleshoot to see why this happened.

Apisit65:
Ok sounds good!

Ari Gerrits

ruxunderscore:
[...]

PR #1829, opened 2024-04-13
Winget Version Check Fix

ruxunderscore:
- Changed version check for Winget to keep Winget updated to latest version and a couple formatting fixes.
- Should mitigate issues, such as, #1824

### DIFF
--- a/functions/private/Test-WinUtilPackageManager.ps1
+++ b/functions/private/Test-WinUtilPackageManager.ps1
@@ -38,10 +38,12 @@ function Test-WinUtilPackageManager {
 
             # Check if Winget's Version is too old.
             $wingetCurrentVersion = [System.Version]::Parse($wingetVersion.Trim('v'))
-            $wingetBadVersion = [System.Version]::Parse("1.2.10691") # Windows 11 (22H2) comes with v1.2.10691, which is bugged.
-            $wingetOutdated = $wingetCurrentVersion -le $wingetBadVersion
+            # Grabs the latest release of Winget from the Github API for version check process.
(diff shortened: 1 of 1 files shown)


Marterich (changes requested):
One change I would welcome would be to keep `$wingetBadVersion = [System.Version]::Parse("1.2.10691") # Windows 11 (22H2) comes with v1.2.10691, which is bugged.`.
Then you could change the output shown to the user to be like a Yellow warning message saying something like "Winget is not the newest Version available" if there is an update avaliable and a Red Message "Winget is Out of Date" when the installed Version is older then the bugged version. This way it would be clearer to the user if an update is strictly necessary or only advisable.
In my opinion there still should be a clear distinction between not having the newest version and a broken one, but I'm interested if anyone has another take on this.

Otherwise nice work :)

ruxunderscore:
> One change I would welcome would be to keep `$wingetBadVersion = [System.Version]::Parse("1.2.10691") # Windows 11 (22H2) comes with v1.2.10691, which is bugged.`.
> Then you could change the output shown to the user to be like a Yellow warning message saying something like "Winget is not the newest Version available" if there is an update avaliable and a Red Message "Winget is Out of Date" when the installed Version is older then the bugged version. This way it would be clearer to the user if an update is strictly necessary or only advisable.
> In my opinion there still should be a clear distinction between not having the newest version and a broken one, but I'm interested if anyone has another take on this.
> 
> Otherwise nice work :)

So I've seen multiple cases where WinGet doesn't work for various reasons, not just the bugged version in 22H2, which is why I did this change. However I can see the reasoning for letting the user know they have a bugged version.

ChrisTitusTech (approved):

PR #1836, opened 2024-04-16
Add ability to remove MS Copilot

ModernTTY:
I added the option to remove Copilot AI, this was pretty simple since **_yet_** there aren’t any special roadblocks (can remove Copilot with simple `Remove-AppxPackage`). Tested on latest 23H2 build.

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2300,8 +2300,7 @@
     ],
     "UndoScript": [
       "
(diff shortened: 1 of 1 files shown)


ModernTTY:
Seems that Microsoft has pushed out an update that patches this. Closing for now will search for new solutions.

PR #1841, opened 2024-04-17
Add option to remove Microsoft Copilot AI

ModernTTY:
I added the function to Remove the Microsoft Copilot AI. Besides some small problematic updates it was straightforward. Works on 23H2 build 22631.3447.

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2300,8 +2300,7 @@
     ],
     "UndoScript": [
       "
-      Write-Host \"Install Microsoft Edge\"
-      Start-Process -FilePath winget -ArgumentList \"install -e --accept-source-agreements --accept-package-agreements --silent Microsoft.Edge \" -NoNewWindow -Wait
+      
       "
     ]
   },
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2285,6 +2285,26 @@
       "
     ]
   },
+  "WPFTweaksRemoveCopilot": {
+    "Content": "Remove Microsoft Copilot",
+    "Description": "Removes MS Copilot AI built into Windows since 23H2.",
+    "category": "z__Advanced Tweaks - CAUTION",
+    "panel": "1",
+    "Order": "a025_",
+    "InvokeScript": [
+        "
+        Remove-AppxPackage \"Microsoft.Windows.Ai.Copilot.Provider\"
+        Remove-AppxPackage \"MicrosoftWindows.Client.CoPilot\"
+        Set-ItemProperty -Path \"HKCU:\\Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\Advanced\" -Name \"ShowCopilotButton\" -Type \"DWord\" -Value \"0\" 
+        "
+    ],
+    "UndoScript": [
+      "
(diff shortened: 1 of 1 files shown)


og-mrk (approved):
Good commits, the ability to Undo this action would be great, will research a way to do so and open new PR for it in your Fork. Otherwise, well done @ModernTTY 👍

**EDIT**: just wondering about your approach of removing Copilot, which's why remove it entirely? looking online and I've found a Registry Hack to turn Copilot off, so removing Copilot entirely using `Remove-AppxPackage` is rather an interesting approach.

Please note that **_if_** your intensions is to remove it entirely (Probably due to Privacy reasons, etc...), then I guess your approach is more reasonable than a Simple Registry Hack (You never know what Microsoft might be doing in the background 👀)

ModernTTY:
> **EDIT**: just wondering about your approach of removing Copilot, which's why remove it entirely? looking online and I've found a Registry Hack to turn Copilot off, so removing Copilot entirely using `Remove-AppxPackage` is rather an interesting approach.

My intention was to do a **complete** removal for privacy but I do plan adding an option for a disable in addition to removing later on.

ModernTTY:
Have experienced problems, converting to draft

ChrisTitusTech:
You might also reference this reg edit via Set-ItemProperty to disable copilot so windows won't refer to something that isn't there:

```
Windows Registry Editor Version 5.00

[HKEY_CURRENT_USER\Software\Policies\Microsoft\Windows\WindowsCopilot]
"TurnOffWindowsCopilot"=-

[HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\Windows\WindowsCopilot]
"TurnOffWindowsCopilot"=-
```

ChrisTitusTech:
> Does the following code do anything?
> 
> ```
> if((Test-Path -LiteralPath "HKCU:\SOFTWARE\Policies\Microsoft\Windows\WindowsCopilot") -ne $true) {  New-Item "HKCU:\SOFTWARE\Policies\Microsoft\Windows\WindowsCopilot" -force -ea SilentlyContinue };
> New-ItemProperty -LiteralPath 'HKCU:\SOFTWARE\Policies\Microsoft\Windows\WindowsCopilot' -Name 'TurnOffWindowsCopilot' -Value 1 -PropertyType DWord -Force -ea SilentlyContinue;
> ```

I'd use this code AND also add HKLM as well so it disables user and machine level.

ModernTTY:
So this is what happened:
After running tweaks in my testing and manually turning on Copilot in settings the icon didn’t appear so I made the PR. Then after opening PR I ran a security update and rebooted my computer after that is saw the Copilot icon and somehow when I pressed it it opened, the weird thing is that it still doesn’t show up in the settings app list, and doesn’t show up when running `Get-AppxPackage` I have a few theories on what happend.

1. The security update partially reinstalled Copilot but didn’t install the metadata for it to show up in settings.
2. There actually are lots of protections for Copilot specifically Windows uninstalls Copilot in RAM but when I rebooted and Windows tried to actually write the deletions it failed doing that.
3. The Copilot files are stored somewhere in cache/temporary file but the actual files are gone.

PR #1846, opened 2024-04-18
Update Get-TabXaml function

ogfrm:
- Order of items are handled from orders in feature  and tweaks files.  
- remove category, order, panel fields from feature and tweaks json files

### DIFF
--- a/config/feature.json
+++ b/config/feature.json
@@ -1,24 +1,20 @@
 {
+  "category0": {
+    "category": "Features"
+  },
   "WPFFeaturesdotnet": {
     "Content": "All .Net Framework (2,3,4)",
     "Description": ".NET and .NET Framework is a developer platform made up of tools, programming languages, and libraries for building many different types of applications.",
-    "category": "Features",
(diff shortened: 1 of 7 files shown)


ogfrm:
> Reviewing this will take some time. That is one chunky PR.
> 
> What was the main purpose of moving the order? I had it organized into sections and now all the tweaks are reorganized alphabetically so I can't verify if anything has changed.

Purpose is to remove extra order,category,panel fields from json file. Output files remains same. It does not improve anything at  runtime. It is just my personal preferance.

ChrisTitusTech:
I can't accept the PR because the functions have been moved around. I have no way of verifying if values are changed. 

Example...

Original code
WSL Function
Tweak Function
NFS Function

Your code 
Tweak Function
NFS Function
WSL Function

Trying to debug or see if things are changed are impossible when this happens and why I can't take the PR. This should be several PRs and never reorder functions as it will show up as a sea of changed for the maintainer.

PR #1845, opened 2024-04-18
Update inputXML.xaml

ruxunderscore:
- Added UseLayoutRounding="True" (suggested by @SkylerWallace)
- This may help fix some cases where text may look blurry on screen for some people. #1823
- Tested in Win10 and Win11

PR #1852, opened 2024-04-20
Test 2024 03 30

ChrisTitusTech:
- Multiple winget fixes by @ruxunderscore 
- Fix DISM conversion for ESD files @CodingWonders 
- MASSIVE fixes and Issue resolved by @og-mrk 
  - toggle widgets
  - fix special characters in json files
  - multiple program additions in `applications.json`
  - restarting explorer to fix classic right click fix
  - github actions fix for leaving an open issue
  - fix disable storage sense
  - robocopy fix for onedrive uninstall for users with space in username
- Independent Edge Removal Fixes by @DaEpicR 
- Disable SCOOBE @ICTdE
- O&O Tweaks update by @Marterich 
- Adding catch for Windows Restore Point creation.


og-mrk:
If I have to give my favorite change, it's PR #1845 by @ruxunderscore , with @SkylerWallace 's Suggestions. Man was I feeling like an Old man trying to read that blurry text 😆

Thanks for the change, and to everyone who contributed thus far into the Project, It made my life easier, _and to me at least_, finding a somewhat small, but a very nice community of Intelligent PowerShell Scripters.. which helped me a lot in improving myself, not only in PowerShell, but with in my own self..

So Thank you Chris.. I really do mean it 🤍

ruxunderscore:
> If I have to give my favorite change, it's PR #1845 by @ruxunderscore , with @SkylerWallace 's Suggestions. Man was I feeling like an Old man trying to read that blurry text 😆
> 
> Thanks for the change, and to everyone who contributed thus far into the Project, It made my life easier, _and to me at least_, finding a somewhat small, but a very nice community of Intelligent PowerShell Scripters.. which helped me a lot in improving myself, not only in PowerShell, but with in my own self..
> 
> So Thank you Chris.. I really do mean it 🤍
> 
> 
> **EDIT**: Will try and continue working on this project, but I can't give promises, as Life can get in the way sometimes. see you all in the Next Monthly Release Cycle (I really don't know what to call it, but it's a test & release/merge cycle of some sort, so it's accurate enough 😆)

I'm glad I decided to start contributing to the project, I never thought myself good enough to contribute until I saw that WinGet functionality wasn't great in winutil. It's been a lot of fun, even if my methods might not be the most efficient (as I am sure there is someone out there that will be able to improve upon my work.).

ChrisTitusTech:
Thanks @ruxunderscore - I ended up rewriting the process for winget install, but the entire idea was yours and lead to fixing the program installation issue I've had for years. Like Spotify failing because it wants unelevated prompts.

I ended up doing a Start-Process for launching directly instead of calling a function and then had it work through a machine scope with admin  then user scope with admin and if all that fails it does a runas user to unelevated and install that way. So now any program no matter how the devs want it installed, should work.

ruxunderscore:
> Thanks @ruxunderscore - I ended up rewriting the process for winget install, but the entire idea was yours and lead to fixing the program installation issue I've had for years. Like Spotify failing because it wants unelevated prompts.
> 
> I ended up doing a Start-Process for launching directly instead of calling a function and then had it work through a machine scope with admin then user scope with admin and if all that fails it does a runas user to unelevated and install that way. So now any program no matter how the devs want it installed, should work.

Just looked at the changes. I like that better, at the end of the day. At the end of it all, it just turns out that every installer is built differently for Windows. This, means that some installers don't honor machine-wide installs or user installs, but even more annoyingly, some installers, don't technically do either one. So, while Chrome might handle user and machine-wide installations, Paint.NET does not and fails for both. Paint.NET also doesn't honor the silent mode flag, so it will still open a separate window for the installation process no matter what flags you give it. 

With all that in mind, from what I can tell is that if you send winget the flag for machine-wide install to something like Paint.NET, it fails because the winget catalog doesn't have a "machine-wide" installer for Paint.NET (By the way, if I recall correctly, Paint.NET also didn't have a "user installer," either, so you have to use the "Any" flag, when I was using the Modules.). Winget is great, but it's things like that, which can make it unreliable without having all the work arounds.

ChrisTitusTech:
Yeah, it's just one of those "Microsoft things" haha.

PR #1857, opened 2024-04-21
Added ability to import drivers

CodingWonders:
This PR adds the ability to import drivers from the active installation to the MicroWin image:

<p align="center">
    
</p>

This is the introductory bit, focusing on the core technology of this feature. More additions will come in the future.

### DIFF
--- a/functions/public/Invoke-WPFMicrowin.ps1
+++ b/functions/public/Invoke-WPFMicrowin.ps1
@@ -53,6 +53,7 @@ public class PowerManagement {
 	$keepEdge = $sync.WPFMicrowinKeepEdge.IsChecked
 	$copyToUSB = $sync.WPFMicrowinCopyToUsb.IsChecked
 	$injectDrivers = $sync.MicrowinInjectDrivers.IsChecked
+	$importDrivers = $sync.MicrowinImportDrivers.IsChecked
 
     $mountDir = $sync.MicrowinMountDir.Text
     $scratchDir = $sync.MicrowinScratchDir.Text
@@ -117,11 +118,35 @@ public class PowerManagement {
 			if (Test-Path $driverPath)
 			{
 				Write-Host "Adding Windows Drivers image($scratchDir) drivers($driverPath) "
-				Add-WindowsDriver -Path "$scratchDir" -Recurse -Driver "$driverPath"
+				dism /English /image:$scratchDir /add-driver /driver:$driverPath /recurse | Out-Host
 			}
 			else 
 			{
 				Write-Host "Path to drivers is invalid continuing without driver injection"
+			}
+		}
+
+		if ($importDrivers)
+		{
+			Write-Host "Exporting drivers from active installation..."
+			if (Test-Path "$env:TEMP\DRV_EXPORT")
+			{
+				Remove-Item "$env:TEMP\DRV_EXPORT" -Recurse -Force
+			}
+			New-Item -Path "$env:TEMP\DRV_EXPORT" -ItemType Directory -Force
+			dism /english /online /export-driver /destination="$env:TEMP\DRV_EXPORT" | Out-Host
+			if ($?)
+			{
+				Write-Host "Adding exported drivers..."
+				dism /english /image="$scratchDir" /add-driver /driver="$env:TEMP\DRV_EXPORT" /recurse | Out-Host
+			}
+			else
+			{
+				Write-Host "Failed to export drivers. Continuing without importing them..."
+			}
+			if (Test-Path "$env:TEMP\DRV_EXPORT")
+			{
+				Remove-Item "$env:TEMP\DRV_EXPORT" -Recurse -Force
 			}
 		}
 
@@ -335,7 +360,7 @@ public class PowerManagement {
 			if (Test-Path $driverPath)
 			{
 				Write-Host "Adding Windows Drivers image($scratchDir) drivers($driverPath) "
-				Add-WindowsDriver -Path "$scratchDir" -Driver "$driverPath" -Recurse
+				dism /English /image:$scratchDir /add-driver /driver:$driverPath /recurse | Out-Host
 			}
 			else 
 			{
--- a/functions/public/Invoke-WPFMicrowin.ps1
+++ b/functions/public/Invoke-WPFMicrowin.ps1
@@ -112,6 +112,47 @@ public class PowerManagement {
             return
         }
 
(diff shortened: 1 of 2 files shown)


KonTy (commented):
I'm so glad you are using dism directly without using applets. I saw that people started to convert dism calls to applets, I think it is a very bad idea, applets are wrappers on dism anyway so they introduce latency. On top of that applets for dism are not always installed I had a win10 machie where MicroWin didn't work after some people converted some calls to applet. I think we should convert all the calls back to dism for compatibility and stability

functions/public/Invoke-WPFMicrowin.ps1
```diff
@@ -117,14 +118,38 @@ public class PowerManagement {
...
+				Write-Host "Adding exported drivers..."
+				dism /english /image="$scratchDir" /add-driver /driver="$env:TEMP\DRV_EXPORT" /recurse | Out-Host
+			}
+			else
+			{
+				Write-Host "Failed to export drivers. Continuing without importing them..."
+			}
+			if (Test-Path "$env:TEMP\DRV_EXPORT")
+			{
+				Remove-Item "$env:TEMP\DRV_EXPORT" -Recurse -Force
```
KonTy, line 149: I think we should have an option to just export drivers for later use, I've had that on my plate for the longest time.  Maybe you could add a path to directory to export to and not delete it. This is important when you do prepping on a different machine, and maybe you want to trim some drivers and leave only the very necessary drivers.

functions/public/Invoke-WPFMicrowin.ps1
```diff
@@ -117,14 +118,38 @@ public class PowerManagement {
...
+			}
+		}
```
KonTy, line 151: Perhaps, you can import drivers to the same directory where Inject Drivers is pointing to?  These are related activities anyway, you need to inject after you import...

functions/public/Invoke-WPFMicrowin.ps1
```diff
@@ -117,14 +118,38 @@ public class PowerManagement {
 			if (Test-Path $driverPath)
 			{
 				Write-Host "Adding Windows Drivers image($scratchDir) drivers($driverPath) "
-				Add-WindowsDriver -Path "$scratchDir" -Recurse -Driver "$driverPath"
+				dism /English /image:$scratchDir /add-driver /driver:$driverPath /recurse | Out-Host
```
KonTy, line 121: Good job on this feature I've had it on my plate for months, you got it done :)

functions/public/Invoke-WPFMicrowin.ps1
```diff
@@ -117,14 +118,38 @@ public class PowerManagement {
...
 
+		if ($importDrivers)
+		{
+			Write-Host "Exporting drivers from active installation..."
+			if (Test-Path "$env:TEMP\DRV_EXPORT")
+			{
+				Remove-Item "$env:TEMP\DRV_EXPORT" -Recurse -Force
+			}
+			New-Item -Path "$env:TEMP\DRV_EXPORT" -ItemType Directory -Force
+			dism /english /online /export-driver /destination="$env:TEMP\DRV_EXPORT" | Out-Host
```
KonTy, line 137: Perhaps, export drivers to "inject drivers" directory? That way they will all be in once place and if you check the inject drivers checkbox they will get injected immediately?

functions/public/Invoke-WPFMicrowin.ps1
CodingWonders: The thing is that this doesn't depend on the directory chosen in that option. I can work on that but leave `$env:TEMP\DRV_EXPORT` as a fallback if that option is enabled and a directory hasn't been specified

CodingWonders:
> I'm so glad you are using dism directly without using applets. I saw that people started to convert dism calls to applets, I think it is a very bad idea, applets are wrappers on dism anyway so they introduce latency. On top of that applets for dism are not always installed I had a win10 machie where MicroWin didn't work after some people converted some calls to applet. I think we should convert all the calls back to dism for compatibility and stability

I've reverted driver addition commands to use DISM simply because I wasn't receiving an output of the cmdlets and, while I didn't experience any problems with the DISM module, that can still happen.

Could it be resolved by importing the DISM module?

CodingWonders:
After doing the same tasks with my utility and applying those to a different image, no component store corruption is detected:

DT_DrvImport

It seems as though the previous tests were a fluke. Can you check this to see if it corrupts the images?

KonTy:
@CodingWonders no I don't see any corruption on my side. I don't see how/why it got corrupted you are using standard dism commands. If it did get corrupted it must have been a fluke.

CodingWonders:
That means this change is ready to be pulled. I'll work on some of your proposed changes before making this PR ready for review though

CodingWonders:
> @CodingWonders no I don't see any corruption on my side. I don't see how/why it got corrupted you are using standard dism commands. If it did get corrupted it must have been a fluke.

It was a fluke all this time:

cmd_u1nehInsHm

Testing this procedure again now, and it's continuing with no issues

functions/public/Invoke-WPFMicrowin.ps1
CodingWonders: The only problem is that I'm not good at modifying XAML code, as I'm more used to WinForms

functions/public/Invoke-WPFMicrowin.ps1
CodingWonders: I've modified this function to take advantage of the injection directory if the option is enabled and the directory exists. Otherwise, it'll keep using `$env:TEMP\DRV_EXPORT`.

powershell_ez0YJ1KGsv

powershell_XVm4rjQeZU

powershell_1tGzyRJJs1

The advantage is that, if a driver injection directory were to be specified, this function won't remove it afterward. I've tried this option but haven't determined if the default option still works. It should, but it's best to test it

functions/public/Invoke-WPFMicrowin.ps1
CodingWonders: This is already implemented by the change mentioned in your last review

functions/public/Invoke-WPFMicrowin.ps1
CodingWonders: Already implemented to some extent by the change below

ChrisTitusTech (approved):

PR #1787, opened 2024-04-01
Add apps from #250

ModernTTY:
List of added apps:

- Microsoft Azure Data Studio
- Java 11, 17, 18, 19 runtimes
- Microsoft SQL Management Studio

PR #1884, opened 2024-04-25
Added java 8 JRE and java 17 JRE

BenignPigeon:
Added java 8 JRE and java 17 JRE

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -903,6 +903,14 @@
 		"link": "https://jami.net/",
 		"winget": "SFLinux.Jami"
 	},
+	"WPFInstalljava8": {
+		"category": "Development",
+		"choco": "temurin8jre",
+		"content": "OpenJDK Java 8",
+		"description": "OpenJDK Java 8 is the latest version of the open-source Java development kit.",
+		"link": "https://adoptopenjdk.net/",
+		"winget": "AdoptOpenJDK.OpenJDK.8"
+	},
 	"WPFInstalljava16": {
 		"category": "Development",
 		"choco": "temurin16jre",
@@ -910,6 +918,14 @@
 		"description": "OpenJDK Java 16 is the latest version of the open-source Java development kit.",
 		"link": "https://adoptopenjdk.net/",
 		"winget": "AdoptOpenJDK.OpenJDK.16"
+	},
+	"WPFInstalljava17": {
+		"category": "Development",
+		"choco": "temurin17jre",
+		"content": "OpenJDK Java 17",
+		"description": "OpenJDK Java 17 is the latest version of the open-source Java development kit.",
+		"link": "https://adoptopenjdk.net/",
+		"winget": "EclipseAdoptium.Temurin.17.JRE"
 	},
 	"WPFInstalljava18": {
 		"category": "Development",


KonTy:
Guys do we really need all these different versions of tools especially Java? I always hesitate to add specific versions of anything. Imagin 6  months from now this will be outdated and catching up and keeping this updated is going to be nightmare. I think, this will just pollute winutil for such a small corner case for maybe one or two people in the world?

ModernTTY:
@BenignPigeon Java 17 runtime already is in Winutil I added it in PR #1787.

og-mrk:
I do see your concerns @KonTy, but may I add that these are different packages under-the-hood, meaning we can't simply add one entry and call it a day.

Now a concern of mine is that, WinUtil will eventually.. be very bloated in terms of it's Apps and/or Tweaks, as more and more apps and features are added, it'll be more difficult to maintain and even use in the first place (Imagine seeing over 200 entries in the Programs Tab.. my goodness will it be too many to view and navigate 😆), so at some point a new solution to organize these entries will be in place.. for now, we should accept Valid Additions PRs, because adding more apps _is generally_ better than denying said PRs.

KonTy:
@og-mrk @ChrisTitusTech  I disaagree with the approach just adding apps blindly, I think we are already at the point where winutil is blotaed, PowerShell is an interpreted language and the more apps we add the slower it gets. We had to do a lot of optimizations to get it to the point it's at, and the performance is still not ideal. It takes 17 seconds to load on my laptop.

If we are going to accept all apps, by this logic, let's just accept all apps from https://winget.run/ repository? I think, we should only accept good quality, free, open source, apps and not pollute the tool with commercial products.

It is not harder and in many cases simpler now to just run ```winget install someapp ``` 
then run winutil -> find an app in the list -> select the checkbox -> click install.. 

We always have to keep in mind what is the value we are providing with WinUtil? Searching apps on https://winget.run/ is as simple as seaching it in winutil.

ChrisTitusTech:
@KonTy @og-mrk My views on apps are this.

Big FOSS apps with lots of users should be added
niche apps with little to no users should not be added
trial apps paid apps should NOT be added

We have a few apps already in the utility that fit niche and trial apps and I will go through and remove those. StartAllBack is on the chopping block because it is trialware. As far as adding old Java versions, that doesn't make sense. There is no reason to add something like that as most will want the most recent version. 

The idea behind the toolbox is that the everyday person can use it. Adding a bunch of garbage like old versions and apps that no one will use makes it harder.

og-mrk:
@KonTy @ChrisTitusTech
Well.. this's a bit concerning 😆 (See screenshot below)

More than **HALF** the PRs open at the time of writing, are just Program Additions (it'll be 6 PRs if we include this one).. we should try to find a solution for the "Too Many App Entries" issue as soon as we can, as I can see this getting out of hand really quick.

> [!NOTE]
> If we consider adding only 10 % of Open Source Software which we see as "highly demanded" by the community, or is worth adding... it'll still be too much to render, and the User Experience will be awful, _and my humble guess_ is that WinUtil will be somewhat annoying/confusing to use, due to how many app entries to navigate, which we obviously don't want.

### Screenshots

image

Issue #1881, opened 2024-04-25
Winget  Failing


ceramicv:
**Describe the bug**
On older Windows 10 laptop trying to install software throws a winget error

**To Reproduce**
Check apps you want to install/upgrade, then click on that button. In the shell windows can see this output:

Program 'winget.exe' failed to run: The file cannot be accessed by the systemAt line:2640 char:35
+             $wingetVersionFull = (winget --version) # Full Version wi ...
+                                   ~~~~~~~~~~~~~~~~.
At line:2640 char:13
+             $wingetVersionFull = (winget --version) # Full Version wi ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ResourceUnavailable: (:) [], ApplicationFailedException
    + FullyQualifiedErrorId : NativeCommandFailed

You cannot call a method on a null-valued expression.
At line:2643 char:17
+             if ($wingetVersionFull.Contains("-preview")) {
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull

You cannot call a method on a null-valued expression.
At line:2652 char:13
+             $wingetCurrentVersion = [System.Version]::Parse($wingetVe ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull

===========================================
---        Winget is installed          ---
===========================================
Version:
    - Winget is a release version.
    - Winget is Out of Date
Getting Installed Programs...
Done...
===========================================
Error: Program 'winget.exe' failed to run: The file cannot be accessed by the systemAt line:28 char:35
+             $wingetVersionFull = (winget --version) # Full Version wi ...
+                                   ~~~~~~~~~~~~~~~~.
===========================================

**Expected behavior**
Applications to be installed/updated

**Screenshots**

**Additional context**

blusewill:
Sorry. Can you upload the text file to here?
Because of GitHub's Markdown Function, it's very hard to tell the issues

Marterich:
[...]

PR #1887, opened 2024-04-26
Fix Winget Detection by doing Proper Error Handling using a Try-Catch

og-mrk:
Should fix issue https://github.com/ChrisTitusTech/winutil/issues/1881
Thanks to @Marterich for suggesting a fix by using Proper Error-Handling 👍

### DIFF
--- a/functions/private/Test-WinUtilPackageManager.ps1
+++ b/functions/private/Test-WinUtilPackageManager.ps1
@@ -21,7 +21,6 @@ function Test-WinUtilPackageManager {
 
     if ($winget) {
         # Get Winget Version and Install it if not detected
-        $wingetVersionFull = ""
         $wingetExists = $true
         try {
             $wingetVersionFull = winget --version
--- a/functions/private/Test-WinUtilPackageManager.ps1
+++ b/functions/private/Test-WinUtilPackageManager.ps1
@@ -20,7 +20,7 @@ function Test-WinUtilPackageManager {
     $status = "not-installed"
 
     if ($winget) {
-        # Get Winget Version and Install it if not detected
+        # Check if Winget is available while getting it's Version if it's available
         $wingetExists = $true
         try {
             $wingetVersionFull = winget --version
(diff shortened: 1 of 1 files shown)


functions/private/Test-WinUtilPackageManager.ps1
```diff
@@ -20,13 +20,16 @@ function Test-WinUtilPackageManager {
     $status = "not-installed"
 
     if ($winget) {
-        # Install Winget if not detected
-        $wingetExists = Get-Command -Name winget -ErrorAction SilentlyContinue
+        # Get Winget Version and Install it if not detected
+        $wingetVersionFull = ""
```
Marterich, line 24:
```suggestion
```
You don't really need to initialize the variable as an empty string. 
Powershell will happily create the variable on first use in the try statement if it does not exist already and if you run into the catch statement, the variable won't be used either because the whole version check is skipped if winget is not installed. 
I would simply remove the line to keep the code as clean as possible :)

Marterich (approved):
Nice work :)
I would only remove the one not necessary variable assignment of $wingetVersionFull

functions/private/Test-WinUtilPackageManager.ps1
og-mrk: > You don't really need to initialize the variable as an empty string.

After a bit of testing, it turns out PowerShell doesn't limit a Variable Initialization into a Try-Catch Scope like I thought it would, or in other words, I thought the value of `wingetVersionFull` after the Try-Catch Code Block(s), would be like accessing an uninitialized variable, and it'll null or something (Was thinking way too low level, not gonna lie 😆)

and it indeed works without that Variable Initialization on line 24, thanks for the catching my silly mistake @Marterich 😄, will commit your suggestion in a new commit 👍

### Screenshots

image

functions/private/Test-WinUtilPackageManager.ps1
```diff
@@ -20,13 +20,15 @@ function Test-WinUtilPackageManager {
     $status = "not-installed"
 
     if ($winget) {
-        # Install Winget if not detected
-        $wingetExists = Get-Command -Name winget -ErrorAction SilentlyContinue
+        # Get Winget's Version and Install Winget if it isn't detected
```
Marterich, line 23:
```suggestion
        # Get Winget's Version
```
Maybe I am way too pedantic, but, I think the comment was not entirely correct from the get-go. 
The function `Test-WinUtilPackageManager` only checks if winget/chocolatey are installed and prints / returns the result. The logic to install winget is handled outside of this function so the commend isn't really accurate of what is happening in the script.

Marterich: Actually felt a bit bad writing the comment because it seemed so inconsequential to fix a single line that didn't even change anything on how the code behaved ^^

PR #1898, opened 2024-04-28
Add option to remove Microsoft Copilot AI

ModernTTY:
After the last PR I made I had a bit more time to spend creating Copilot removal.

Essentially it works in the following way:

1. It uninstalls some Copilot packages
2. Unpins Copilot from taskbar
3. Disables the ability to pin Copilot back to the taskbar

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2333,7 +2333,7 @@
         "
         Get-AppxPackage *copilot* | Remove-AppxPackage
         Set-ItemProperty -Path \"HKCU:\\Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\Advanced\" -Name \"ShowCopilotButton\" -Type \"DWord\" -Value \"0\" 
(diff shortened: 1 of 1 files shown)


og-mrk (approved):
Other than the Undo Script being missing (Looked into it in the past, haven't found any useful resource about re-installing Copilot 😞), this's a Really Cool Commit you've made @ModernTTY 😎

CodingWonders (approved):
I'll only use AI when I want, not when Windows wants. Good job!

CodingWonders:
@og-mrk, I've tested a MicroWin image based on Windows 10 once and it installed a Copilot AppX package.

Unfortunately I don't have the screenshot available, but I'll edit this message when I find it.

UUP Dump also installs Copilot when building the most recent Windows versions. Maybe we could take a look at the apps list to see how we can restore it. That's my guess.

ChrisTitusTech (commented):
Changed from a removal to a disable

ChrisTitusTech (approved):

PR #1931, opened 2024-05-03
Optimize winget download speed

Marterich:
Microsoft in its endless wisdom, created Invoke-Webrequest in such a way that the progress bar, which is shown during a download, drastically reduces the download speed that can be achieved. This is especially painful when downloading larger files (like winget ~200+ MB) 
The simplest workaround is, to disable the progress bar for the time of the download and print a message like "Downloading..." manually.


ruxunderscore:
Ah... Didn't realize that was the issue! Thanks for the solution @Marterich!

ChrisTitusTech (approved):

Issue #1921, opened 2024-05-02
ooshutup10_recommended


Obegg:
Why would you store the `ooshutup10_recommended.cfg` in a separate external file
https://github.com/ChrisTitusTech/winutil/blob/main/config/ooshutup10_recommended.cfg

When you can create it in script:
```powershell
Invoke-WebRequest -Uri https://dl5.oo-software.com/files/ooshutup10/OOSU10.exe -OutFile $ENV:temp\OOSU10.exe
$OOSU10 = '############################################################################
# This file was created with O&O ShutUp10++ V1.9.1436
# and can be imported onto another computer. 
#
# Download the application at https://www.oo-software.com/shutup10
# You can then import the file from within the program. 
#
# Alternatively you can import it automatically over a command line.
# Simply use the following parameter: 
# OOSU10.exe <path to file>
# 
# Selecting the Option /quiet ends the app right after the import and the
# user does not get any feedback about the import.
#
# We are always happy to answer any questions you may have!
# © 2015-2023 O&O Software GmbH, Berlin. All rights reserved.
# https://www.oo-software.com/
############################################################################

P001	+
P002	+
P003	+
P004	+
P005	+
P006	+
P008	+
P026	+
P027	+
P028	+
P064	+
P065	+
P066	+
P067	+
P070	+
P069	+
P009	-
P010	-
P015	-
P068	-
P016	-
A001	+
A002	+
A003	+
A004	+
A006	+
A005	+
P007	+
P036	+
P025	+
P033	+
P023	+
P056	+
P057	-
P012	-
P034	-
P013	-
P035	-
P062	-
P063	-
P081	-
P047	-
P019	-
P048	-
P049	-
P020	-
P037	-
P011	-
P038	-
P050	-
P051	-
P018	-
P039	-
P021	-
P040	-
P022	-
P041	-
P014	-
P042	-
P052	-
P053	-
P054	-
P055	-
P029	-
P043	-
P030	-
P044	-
P031	-
P045	-
P032	-
P046	-
P058	-
P059	-
P060	-
P061	-
P024	-
S001	+
S002	+
S003	+
S008	-
E001	+
E002	+
E003	+
E008	+
E007	+
E010	+
E011	+
E012	+
E009	-
E004	-
E005	-
E013	-
E014	-
E006	-
Y001	+
Y002	+
Y003	+
Y004	+
Y005	+
Y006	+
Y007	+
C012	+
C002	+
C013	+
C007	+
C008	+
[...]

PR #1932, opened 2024-05-03
Compile OOSU Config into winutil.ps1

Marterich:
Changed the compile script so that the OOSU Config files are Compiled right into winutil.ps1 and do not need to be downloaded afterward. 
This removes the need to have a static download link referencing the main branch of the repo.

The idea came from a discussion in #1921


ruxunderscore:
Nice! Anyway to get rid of hard links is a good commit to me.

ChrisTitusTech (approved):

Issue #1907, opened 2024-05-01
"Disable GameDVR" tweak causes Color Management to not get applied in Exclusive Fullscreen


Luadih:
**Describe the bug**
Disabling GameDVR basically makes ICC Profiles or any Color Management done through Windows to not get applied while in Exclusive Fullscreen in games.

This can be kind of confusing with some games because when you start them it seems to get applied but if you alt tab and go inside the game again it doesn't.

**To Reproduce**
1. Have some kind of Color Management applied through Windows (Different Brightness or Gamma is more noticeable)
2. Disable GameDVR
3. Open a Game
4. Switch between window/borderless window and fullscreen modes
5. Compare

Note: Alt+Tab also works but comparison its easier by switching display modes

**Expected behavior**
I won't say that disabling GameDVR should keep the colors applied while in Fullscreen (since I think it may be something with the way GameDVR does what it does), but I'm reporting it in case someone goes through the same pain of trying to troubleshoot why is it happening after using Winutil and not knowing that GameDVR is the cause.

For Winutil I would just say to add a note that it may cause this, or if it's just one of the registry keys that causes it to not change it if possible.

Luadih:
I can try it yeah, but I will only be able to notice the color problem

Luadih:
Restoring **"GameDVR_DXGIHonorFSEWindowsCompatible"** to its original value after the tweak is applied seems to fix the color management issue, but I believe I cannot test if it causes more problems when trying to open overlays

I tested with Counter Strike 2 (with and without vulkan just in case) and Rocket League. I think this is worth saying since there is "GameDVR_HonorUserFSEBehaviorMode" which has a really close name? though it may be for something completely different

Something I missed to note in the original issue, is that this doesn't apply to just color management but it also affects to how quickly alt+tab and changing display modes switch, which is really noticeable by me switching from borderless to exclusive fullscreen in Rocket League

Luadih:
> So what solves the alt+tab issue?

Sorry, I didn't make it really that clear. Both the alt+tab and Color Management problems are fixed with **"GameDVR_DXGIHonorFSEWindowsCompatible"** set to 0.

While setting **"GameDVR_DXGIHonorFSEWindowsCompatible"** to 0, fullscreen has the correct colors and the alt+tab speed is normal, it is possible to break it again just by going to each game and in its compatibility tab simply "disabling fullscreen optimizations" there.

So:
[...]

PR #1934, opened 2024-05-04
Made disabling Fullscreen optimizations its own tweak

Luadih:
This is to fix the confusion talked about in #1907 

* Removed "GameDVR_DXGIHonorFSEWindowsCompatible" modification from "Disable GameDVR" tweak

* Created advanced tweak "Disable fullscreen optimizations"

Disables fullscreen optimizations in all applications. (Which can cause problems with Color Management in Exclusive Fullscreen)

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2573,13 +2573,6 @@
       },
       {
         "Path": "HKCU:\\System\\GameConfigStore",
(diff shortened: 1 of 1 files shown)


ChrisTitusTech (approved):

ChrisTitusTech:
I like it, I need to break apart more of these changes as it can conflict with peoples configurations.

PR #1941, opened 2024-05-06
Main window handle

BKlaasWerkman:
Condition is comparing $proc.Id with [System.IntPtr]::Zero. The Id property is an integer, representing the process ID, and it's not directly related to window handles. Instead, you should compare the MainWindowHandle property against [System.IntPtr]::Zero

### DIFF
--- a/scripts/main.ps1
+++ b/scripts/main.ps1
@@ -271,10 +271,14 @@ Add-Type @"
 "@
     }
 
-    foreach ($proc in (Get-Process | Where-Object { $_.MainWindowTitle -and $_.MainWindowTitle -like "*titus*" })) {
-        if ($proc.Id -ne [System.IntPtr]::Zero) {
+   foreach ($proc in (Get-Process | Where-Object { $_.MainWindowTitle -and $_.MainWindowTitle -like "*titus*" })) {
+        # Check if the process's MainWindowHandle is valid
+    	if ($proc.MainWindowHandle -ne [System.IntPtr]::Zero) {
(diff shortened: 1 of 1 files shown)


og-mrk (changes requested):
@BKlaasWerkman
If your proposed changes is only concerned with `main.ps1` File under `scripts` folder, then Why are you Submitting Change(s) made in `applications.json` File as well?

Either you cut this change from this PR, and make a New PR for it, Or update This PR's Title & Description to be more accurate/descriptive of your Changes.

**EDIT**: Noticed that you've made a separate PR for adding Microsoft Office App Entry to `applications.json` File. (here's [a link](https://github.com/ChrisTitusTech/winutil/pull/1946) to it) Now the best course of action is to not have this App Addition, by simply removing the commit titled `Add Microsoft Office apps`.

BKlaasWerkman:
@og-mrk Thanks for this. I've now removed the commits.

og-mrk (approved):
Looks good to me. Thanks for your contribution 😄

Some references:
> Process.MainWindowHandle: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.mainwindowhandle?view=net-8.0
> IntPtr Example showcasing similar usecase of `IntPtr.Zero` but in C# (see the Second Example): https://learn.microsoft.com/en-us/dotnet/api/system.intptr.zero?view=net-8.0#remarks

ChrisTitusTech (approved):

PR #1886, opened 2024-04-25
Test 2024 04 20 2 | 7 additions to applications.json

misspent:
Adds:

[NetLimiter](https://www.netlimiter.com) 
[OneCommander](https://www.onecommander.com)
[Borderless Gaming](https://github.com/Codeusa/Borderless-Gaming)
[Equalizer APO](https://sourceforge.net/projects/equalizerapo)
[FreeFileSync](https://freefilesync.org)
[Elgato Stream Deck](https://www.elgato.com/uk/en/s/downloads)
[Stardock Curtains](https://www.stardock.com/products/curtains)

I couldn't test it myself because I'm new and not entirely sure what to do, but I am willing to learn and apologies if I made a mistake. Keep up the astounding work.

PR #1943, opened 2024-05-06
Only install choco if needed

Marterich:
This PR changes the behavior of the script to only install chocolatey if needed, and not as per default on the startup of winutil. 

Changed the main script to check/install if winutil is installed instead of choco.

Added checks in the places where choco is called to make sure that everything keeps working as expected.

### DIFF
--- a/functions/private/Install-WinUtilWinget.ps1
+++ b/functions/private/Install-WinUtilWinget.ps1
@@ -53,6 +53,8 @@ function Install-WinUtilWinget {
         Write-Host "Failure detected while installing via GitHub method. Continuing with Chocolatey method as fallback." -ForegroundColor Red
         # In case install fails via GitHub method.
         Try {
+        # Install Choco if not already present
+        Install-WinUtilChoco
         Start-Process -Verb runas -FilePath powershell.exe -ArgumentList "choco install winget-cli"
         Write-Host "Winget Installed" -ForegroundColor Green
         Write-Output "Refreshing Environment Variables...`n"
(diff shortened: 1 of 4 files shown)


IDisposable:
Excellent!

ruxunderscore:
Nice! The Winget Fix button that is in the utility still tries to use choco for the install. I am sure I can cook up a fix for that tonight to coincide with this, though.

Marterich:
> Edit: nvm
> An issue I can still see is—What do we do if winget doesn't have an installer for an application (i.e. #1886)? Do we need to add a check for na and failover to choco for the install? 

Yes, i was wondering about that as well, because I couldn't find any logic that worked in that way, when I created the PR. 
I'm not quite sure if I missed something or if the fallback was removed in a previous change of winutil. 

I sadly don't really have time to validate this or implement the change myself in the next two weeks so feel free to open a PR or an issue to make sure choco can be used as a fallback. Should probably be a simple fix

ruxunderscore:
> > Edit: nvm
> > An issue I can still see is—What do we do if winget doesn't have an installer for an application (i.e. #1886)? Do we need to add a check for na and failover to choco for the install? 
> 
> Yes, i was wondering about that as well, because I couldn't find any logic that worked in that way, when I created the PR. 
> I'm not quite sure if I missed something or if the fallback was removed in a previous change of winutil. 
> 
> I sadly don't really have time to validate this or implement the change myself in the next two weeks so feel free to open a PR or an issue to make sure choco can be used as a fallback. Should probably be a simple fix

When I made my changes, there wasn't a fallback to choco at that point, so we might need to create a new fallback mechanism.

ChrisTitusTech (approved):

ChrisTitusTech:
Just added a choco fallback will merge this and add another spot to check for choco install.

PR #1954, opened 2024-05-09
Adobe Network Blocking / Adobe Debloater / Tweak Fixes & more

MyDrift-user:
Disclaimer:
- You can run/test my fork using `irm mdiana.win/util | iex`
- Everything was tested on two devices running Win11 23H2
   - It seems to work for me :)

Fixes:
- Fixed "WPFTweaksRestorePoint" to always run as first tweak, which it does at the time but only because it starts with "C" & the ordering of the tweaks is done alphabetically.
- Fixed Shell Names in "Right-Click Method" under "Usage" in README to accurate ones

Modifications:
- Moved Checkboxes a bit to the right so it looks more organized with the rest
- Redesigned Preferences
  - Put Toggles to the left instead of the right to be more consistent in the design
  - Removed "Enable" & "Disable" from the toggles because they define the action (which would be the other way around). 
     but actually define the state. Adding another d would make the toggle uneccessary long so i removed it completely
- Renamed "Remote Desktop Connection Manager RDCMan" to simply "RDCMan" because it is a commonly used shortform (if you actually know what it is). even on the website (https://learn.microsoft.com/en-us/sysinternals/downloads/rdcman)

Additions:
- Added Preference: Snap Window (Toggle entire Snap Windows Feature)
  Settings > System > Multitasking > Snap Windows
- Added Preference: Snap Assist Suggestion (Toggles Suggestions to snap other Windows on the rest of the screen you just snapped one.
  Settings > System > Multitasking > Snap Windows >
- Block Adobe Networking (As of your suggestion of last week i added this tweak to add redirects of Adobe domains/IPs to a local domain (https://github.com/Ruddernation-Designs/Adobe-URL-Block-List)
  - I am not sure if that should be in the same tweak as Debloat Adobe, so i 
    created a new one.
- Added Application: "AFFiNE", an open source notion/obsidian alternative with many features & syncable using selfhosted instance
  https://affine.pro/
- Added Application: "Wazuh", the agent of a super cool open source security monitoring platform.
  https://wazuh.com/

Suggestion:
- I am working on an Import mechanism that lets the user import Applications/Tweaks/Features using a Source file as Path or link. It would be a GUI accessable from the settings where the user can currently import/export selections. I would like to know you opinion about it to know if i should take my time to finish that. Test it using `irm mdiana.win/util/smgmt | iex`
- I have not got it to work correctly with the WPF Window but i am a big fan of the "PoshTaskbarItem" Module. 
  - Demo of all features: 
     - Install-Module -Name PoshTaskbarItem -Scope CurrentUser
     -  `irm mdiana.win/demo/poshtaskbaritem | iex`
  - There would be so many cool possibilities with it like:
    - custom Taskbar icon
    - JumpTask to WinUtil / Actions / Tabs
    - Counter on taskbar counting how many left there are of
      - applications to install/upgrade/uninstall
      - Tweaks to be applied
    - Taskbar Progressbar to see progress on apps/tweaks or MicroWin
    - Flash Taskbar item when something has finished

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Winutil must be run in Admin mode because it performs system-wide tweaks. To ach
 
 1. **Right-Click Method:**
    - Right-click on the start menu.
-   - Choose "PowerShell As Admin" (for Windows 10) or "Windows Terminal As Admin" (for Windows 11).
+   - Choose "PowerShell (Admin)" (for Windows 10) or "Terminal (Admin)" (for Windows 11).
 
 2. **Search and Launch Method:**
    - Press the Windows key.
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Winutil must be run in Admin mode because it performs system-wide tweaks. To ach
 
 1. **Right-Click Method:**
    - Right-click on the start menu.
-   - Choose "PowerShell (Admin)" (for Windows 10) or "Terminal (Admin)" (for Windows 11).
+   - Choose "Windows PowerShell (Admin)" (for Windows 10) or "Terminal (Admin)" (for Windows 11).
 
 2. **Search and Launch Method:**
    - Press the Windows key.
--- a/config/applications.json
+++ b/config/applications.json
@@ -114,7 +114,7 @@
 	"WPFInstallrdcman": {
 		"category": "Microsoft Tools",
 		"choco": "rdcman",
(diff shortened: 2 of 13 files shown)


MyDrift-user:
undone compilation but still nagging, damn

ChrisTitusTech:
Like a lot of these fixes.

PR #1956, opened 2024-05-09
Test 1715290913 fix ugly ifs

ty802:


ruxunderscore:
Thank you! I wasn't thinking properly when I wrote these nested ifs. T^T

PR #1891, opened 2024-04-26
add GlazeWM to applications.json

FedeP14:
First contribution so please tell me if I did something wrong

As the commit message I added [GlazeWM](https://github.com/glzr-io/glazewm) to the applications.json

PR #1958, opened 2024-05-10
Fix Typo in the App Entry 'GlazeWM' in 'applications.json' File

og-mrk:
### Problem that arise from such a Typo

<div align="center">

<p><i>The App Entry not order correctly, and the fact the Install Function(s) can't pick the App Name due to the Typo Position affecting the Underlying Logic.</i></p>
</div>

### How to avoid such problem(s) in the future

Beside better/stricter PR review process, _even though in this particular case, the typo was caught waayyy back (approx. 2 weeks ago) in https://github.com/ChrisTitusTech/winutil/pull/1891 PR_, there should be Automatic Checks for these kind of issue, which it can be in the `Compile.ps1` Script, or as a Separate Unit Test all together.

This solution above should reduce the possibility of this ever happening again, because the Unit Test(s) will catch it as soon as possible.. _at least, that's in theory_.

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -2655,7 +2655,7 @@
 		"link": "https://github.com/magic-wormhole/magic-wormhole",
 		"winget": "magic-wormhole.magic-wormhole"
 	},
-	"WPFInstalglazewm": {
+	"WPFInstallglazewm": {
 		"category": "Utilities",
 		"choco": "na",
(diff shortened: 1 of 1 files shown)


ChrisTitusTech (approved):

og-mrk:
Fixed merge conflicts, all should be good 👍

PR #1960, opened 2024-05-10
 Install flow rework

ty802:
Sperated Winget and Chocolatey setps when installing software add skip them when not needed


ty802:
Closing will open new pr on non test branch

PR #1965, opened 2024-05-13
Fix typo in filename

ModernTTY:
In the Wiki directory of Winutil one file ends with ,md and so it doesn’t have a file extension. Changed to .md.


og-mrk (approved):
Looks good, you got one sharp eye there @ModernTTY 😃

Issue #1515, opened 2024-01-31
Vencord not installing


Tman564wsDEV:
**Describe the bug**
Vencord is not installing

**To Reproduce**
1. Go to 'Install'
2. Click on 'Vencord'
3. Wait for it to try to install

**Expected behavior**
I expect Vencord to install

**Screenshots**
![image](https://github.com/ChrisTitusTech/winutil/assets/133183000/d1a952b1-968c-4d31-97a6-6994574104c4)
![image](https://github.com/ChrisTitusTech/winutil/assets/133183000/739e006d-8b24-4ab4-b7ae-723af1adc54f)

**Additional context**

inmymum:
I had to manually download vencord's executable from vencord.dev.

Yuuh15:
The error is caused by the `--scope=machine` argument in winget. It is causing a lot of problems for a lot of installers that do not support it. But without the arg the following error occur: 
![powershell_nuf9k3ZaAg](https://github.com/ChrisTitusTech/winutil/assets/97139700/71744584-708e-473f-9513-188a61e4cd72)
The installation is simply borked through winget.

Garou7:
Same error when installing Starship shell.
![2024-01-31 21_19_31-irm https___christitus com_win _ iex(Admin)](https://github.com/ChrisTitusTech/winutil/assets/113905482/e4e69bc1-6c49-431c-84d0-1d4925842ce1)

ChrisTitusTech:
These issues should be resolved in the latest commit 7df98c20c3e4bcc5354eea174e5715a10fd661b9

possumdoe:
Sorry @ChrisTitusTech but to me this issue seems related to #1580, #1515, #1502 and #1494

The fact that some packages doesn't specify in the manifest the **scope of the installer** cause the installation to fail and the error "No applicable installer found ..."

Seems confirmed even by this comment on `winget-cli` repo https://github.com/microsoft/winget-cli/issues/2176#issuecomment-1133353646

PR #1968, opened 2024-05-15
Removed broken winget from applications.json

Owen-3456:
The winget does not work for vencord. The installer downloads and then fails. (#1515)


Owen-3456:
additionally the 'Get Installed' button does not recognise vencord as being installed because of the broken package

ChrisTitusTech (approved):

PR #1974, opened 2024-05-16
Added open-source application QGIS

SnoepNFTs:
I added the FOSS application QGIS

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -1,2594 +1,2610 @@
 {
-	"WPFInstall1password": {
-		"category": "Utilities",
-		"choco": "1password",
-		"content": "1Password",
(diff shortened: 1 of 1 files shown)


og-mrk (changes requested):
This PR of yours will probably be rejected because the changes you've made are unclear as to **where there are (your _actual_ changes)** in your git commit, please change your editor's settings to match the style of this repo's JSON files, which's using Tabs instead of Spaces, so that git can show your intended changes better/properly.

I hope you can understand why the above reason is quite important in the review process, as we can't simply review every line of JSON when the JSON file in question is over ~2900 lines.

SnoepNFTs:
oops I didn't notice that. I'll send a new PR when I have the time

PR #1975, opened 2024-05-17
Update applications.json

Theproccy:
Fixed the link to the Jellyfin media player (removed the erroneous "f")


ModernTTY:
This is already fixed currently in test branch.

Theproccy:
In that case I am sorry for not digging deeper.

PR #1979, opened 2024-05-18
Optimize Display Behaviour of Category Labels 

Marterich:
Modified the display behavior of the Labels with the Category Description to be hidden on the application page if the category does not contain any applications (for example while using the search) 

### Before:
image

### After: 
image


ChrisTitusTech (approved):
Looks far better when searching! Thank you!

PR #1990, opened 2024-05-22
Added Everything 1.5a to Applications (Includes Dark mode)

sbrandsen:
This is the same publisher as the Everything Search already included, but has a dark theme (among numerous other improvements)

Don't get intimidated by the "Alpha" software release life cycle, it was released on 2021-03-13 and has been very stable.

For more details, you can refer to the Everything 1.5a announcement [here](https://www.voidtools.com/forum/viewtopic.php?t=9787) and the Winget package link [here](https://winget.run/pkg/voidtools/Everything.Alpha).

image


ChrisTitusTech:
I don't want to bloat things up with betas or alpha software.

PR #1992, opened 2024-05-22
Winget install return code handling & Window management

Marterich:
Fix return code handling : 
- If you had an application installed and clicked install/update again, winget would exit with return code "-1978335189" No applicable update found, which is != 0, and therefore the logic for the fallback installation methods would start.

Changed winget window management: 
- Modified the winget install to utilize the open powershell window rather than create new windows for each install. 

Personal opinion: I think the function Install-WinUtilProgramWinget is starting to get a bit out of hand in terms of repetition of code blocks, repetition of the same arguments for winget, and so on.  I think that we should rework the function (maybe after the next release when it's clear if the fallback works as expected) and maybe split it up into smaller logic blocks, or multiple functions to remove some of the repeated code and keep the quality of the code up to the standard of the rest of winutil.

### DIFF
--- a/functions/private/Install-WinUtilProgramWinget.ps1
+++ b/functions/private/Install-WinUtilProgramWinget.ps1
@@ -12,6 +12,7 @@ Function Install-WinUtilProgramWinget {
     
     .NOTES
     The triple quotes are required any time you need a " in a normal script block.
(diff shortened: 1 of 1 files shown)


ChrisTitusTech (approved):
Agreed, I was messing around with this for a couple hours trying to strike a good fallback on program installs.

PR #2046, opened 2024-06-06
Enhance DNS ComboBox

MyDrift-user:
- Replaced "Default" with "Custom"
- Set Default Dropdown Content to applied option or "Custom" if not in the list.
- saves custom IPs on multiple adapters during session in global variable
> If user sets DNS to egs. Cloudflare and runs the tweaks but wants to undo it before he closed the script he can select "custom" and run it to apply the saved addresses

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,9 @@ If you encounter any challenges or problems with the script, I kindly request th
 
 ## Contribute Code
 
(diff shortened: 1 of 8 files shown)


v-Kaefer:
Change the commit branch from 'main', to 'test-applications-2024-06'. 
Or literally any other branch man but, *don't commit to main*...

PR #2051, opened 2024-06-07
Tweaks

KonTy:
+ Added new **release actions**, every time we merge into main it will   create a release by extracting the version fromwinutil.ps1 and setting  all appropriate tags and versions
* Fixed .getignore issues, **it's it not enough to just add to gitgnore** like we had it before, if it was already tracked by git it will be continued to be tracked by git, so you have to 'git rm' it 
* Fixed ooshutup, there is no need to download configs, they are already compiled into winutil 
* Updated ooshutup config to include new version tweaks like remove copilot and a couple of others
+ Added a tweak to remove and disable Intel LMS, which have been   identifed as a potential backdoor into Windows.


functions/public/Invoke-WPFOOSU.ps1
Marterich, line 1: Nice, 
I think this broke while merging my last PR regarding OOSU because there were some weird File Conflicts that had to be manually merged by Chris

ChrisTitusTech (approved):

functions/public/Invoke-WPFOOSU.ps1
ChrisTitusTech: Yeah I'm moving to independant PRs instead of bundling them on test branch. So that won't happen again. @Marterich

ChrisTitusTech:
This actually has a bug that wipes out the Compile action. The Update release should have been a new action.

PR #2061, opened 2024-06-10
Rework Compile Script

Marterich:
This PR reworks the Compile script. 

Addressed Issues: 
- The compile sometimes would fail and show a "file in use" error when writing to winutil.ps1
-> This is addressed by assembling the whole script in memory and only writing to disk once (per created file)

Changes: 
- Add a Debug switch that only creates the auto-generated XAML files on disk if desired. If Debug isn't specified the files are removed from the disk (to avoid confusion with different versions of a newly compiled winutil script and old XAML files)
- Add a visual progress bar detailing the compile status using Write-Progress
- Add the header as a multiline string


og-mrk:
Great improvements overall, happy to see xaml files not being created whenever compiling! 😎

This change and the `.gitignore` fix in PR #2051 should remove the worry of accidently adding compiled files by new users, not to mention the other nice changes you've made.

I'll try testing it when I've some spare time, well done @Marterich 😄

Issue #2075, opened 2024-06-11
How to use WinUtil on Linux? 


foss-enjoyer:
Do I need windows in the first place to use WinUtil? Can I use it natively on Linux?

MagmaBro123:
This is for Windows only.

og-mrk:
@foss-enjoyer
WinUtil (short for `Windows Utilities`, also known as `Windows Tool Box`) is a PowerShell script designed and created for Windows only, as all of its Tweaks and Configuration is Windows related ones, the App Installation Section of WinUtil uses Package Managers (at the time of writing, only `Chocolatey` and `WinGet` Package Managers), which are designed for Windows.

Which concludes an answer to your question, that being.. No, WinUtil can not run without Windows, as WinUtil requires the system to be Windows 10 or 11, and WinUtil can't be ran natively on Linux.

PR #2078, opened 2024-06-12
Exit Script on non Windows Machines

MyDrift-user:
- Checks OS using $env:OS
- If OS does not equal to Windows NT it exits

This removes the endless errors if running on non Windows Machines.
It informs the user that the ToolBox is only for Windows, hopefully decreasing confusion like in Issue #2075


ChrisTitusTech:
Are people really trying to run the Windows Utility in other Operating systems? Seems like this adds more code bloat without solving any issue that someone has.

PR #2083, opened 2024-06-12
Remove force install of Winget + Small improvements

### LINKED ISSUE #2022: Winget Outdated...install takes FOREVER.
**Describe the bug**
Upon the first "install" item, the script detects that winget is out of date and tries to install an updated version. The problem is that the downloading of the new Winget util takes on average of an hour or more. Initially I thought this was a winget issue (can't download winget with winget though, am I right?) so I tweaked the winget `settings.json` file to use a different downloader, but that didn't fix the issue.

The problem is not bandwidth as I have a 200mb down/200mb up connection, and the util is only 155mb (per the [Microsoft App Store](https://apps.microsoft.com/detail/9nblggh4nns1?hl=en-us&gl=US#activetab=pivot:overviewtab)). 

**To Reproduce**
1. Run winutil
2. Click on 'Install'
3. Select item to install
4. Script detects that winget is out of date and tries to install the latest winget util.

**Expected behavior**
I honestly expect this to take a minute or two at best.

**Additional context:
I assume this is an issue with limitations around`Invoke-WebRequest` found in the `Get-WinUtilWingetLatest` function (located in `winutil.ps1`), but unfortunately don't have any suggestions on how to speed this up besides possibly incorporating the `-OutFile` parameter to download the files directly to the desired location without storing them in memory first.

### LINKED ISSUE #2039: Winget Auto Install
**Describe the bug**
Winget gets installed automatically. Its considered a bug not a feature as its impossible to remove/uninstall

**To Reproduce**
Run the latest version of winutil

**Expected behavior**
Only portable chocolaty gets installed. (It can be removed very easily by deleting the file). Ask user if they want to install winget (yes/no)

**Screenshots**
N/A

**Additional context**
Users should be made aware before software gets auto installed and should be able to choose whether they want to install or not. This tool gets used by a very diverse group of users with different needs and perhaps it shouldn't force install software with out user consent.

### LINKED ISSUE #2068: Remove winget and chocolately
I recently bought the .exe wrapper of this software because I thought that would remove the requirement of download both winget and chocolately but it still did install it. Can there please be a separate version without winget or chocolately, I don't use the download section, I only use the tweaks scripts. I already bought the program so I hope this can happen.

### LINKED ISSUE #2076: Not use winget but use chocolatey only
First run: Winget is not installed, winutil then installs it, then does the same thing for chocolatey
2nd run: same thing happens but both winget and chocolatey are installed, yet winutil still insists on downloading both again?

Winget takes absurdly long to install given how small it is, chocolatey installs at a duration that is reasonable

And I know that winutil can run with chocolatey as winutil uses chocolatey as a fallback if all else fails

Marterich:
- Removes the forced install of winget on startup. Suppose a user clicks install on the applications page. In that case, winget will be installed anyways, but this removes the delay on bootup for users with a bad internet connection and only want to use winget for the tweaks or micro win. (Closes #2076, Closes #2068, Closes #2039, Closes #2022, Closes #2065) (Maybe it's possible to install winget in a separate Runspace but I haven't tested this)

- Disables the unnecessary Progress bar on bootup which is normally shown when "Get-ComputerInfo" is run. The user doesn't care how long this takes, because the command runs in a separate Runspace anyway. 

- Adds a -Run Toggle in the Compile script so when developing and constantly recompiling and testing, it's easier to compile and run with a single command

### DIFF
--- a/Compile.ps1
+++ b/Compile.ps1
@@ -1,5 +1,6 @@
 param (
-    [switch]$Debug
+    [switch]$Debug,
+    [switch]$Run
 )
 $OFS = "`r`n"
 $scriptname = "winutil.ps1"
@@ -107,3 +108,7 @@ else {
 
 Set-Content -Path $scriptname -Value ($script_content -join "`r`n") -Encoding ascii
 Write-Progress -Activity "Compiling" -Completed
+
+if ($run){
+    Start-Process -FilePath "powershell" -ArgumentList ".\$scriptname"
+}
--- a/scripts/main.ps1
+++ b/scripts/main.ps1
@@ -133,9 +133,17 @@ $sync.keys | ForEach-Object {
 
 # Load computer information in the background
 Invoke-WPFRunspace -ScriptBlock {
-    $sync.ConfigLoaded = $False
-    $sync.ComputerInfo = Get-ComputerInfo
-    $sync.ConfigLoaded = $True
+    try{
+        $oldProgressPreference = $ProgressPreference
+        $ProgressPreference = "SilentlyContinue"
+        $sync.ConfigLoaded = $False
+        $sync.ComputerInfo = Get-ComputerInfo
+        $sync.ConfigLoaded = $True
+    }
+    finally{
+        $ProgressPreference = "Continue"
+    }
+    
 } | Out-Null
 
 #===========================================================================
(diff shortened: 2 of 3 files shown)


brrock (approved):
Great

ChrisTitusTech (approved):

Issue #2085, opened 2024-06-12
Feature Removal Discussion: Simple OO Shutup Tweak


Marterich:
Hey All,

I know there is a Discussion Tab but I think here, more People will see the Discussion and have the ability to chime in. 
Here it goes. 
### The Proposal
I'm considering if the time has come to remove the simple "Run OO Shutup" Tweak from the Essential Tweaks section of winutil and only keep the "Customize OO Shutup Tweaks" Button.

### The Reason
Over the last few months, OO-Software (the company behind OO-Shutup) has become way more lax regarding its "Recommended" Settings. Thereby I mean, they seem to have started to include tweaks as recommended which disable functionality of Windows that a bunch of people want to use. (For example The Clipboard History or RDP Access to the Computer) This year alone more than 15 issues were opened on how to reenable the clipboard history on this repo.

### Pro/Cons of having the "Essential Tweak" Checkbox
[+]It's nice to apply all the recommended tweaks with just one click (instead of 4 via the Customize Button)
[-] It's not clear to the user which settings are being changed
[-] It takes time to keep the config files up to date

I'm not sure if the "Ease of Use" argument is enough to justify the ongoing "support" for the feature considering the time of all the community members who are involved in answering and resolving issues (not only regarding this issue) 

I'm interested in hearing what you all think about this :)

stuckydev:
My opinion: since winutil should be used by more experienced windows-users anyway I think only having the "customize button" is a good decision.

Issue #2047, opened 2024-06-07
Big search box in task bar keeps appearing when running tweaks


jonesy-b-dev:
**Describe the bug**
I have the big search box in the task always set to always hide but when I run the tweaks (always run standard profile) the big search bar appears again and is really annoying. winutil shouldnt be changing this setting

**To Reproduce**
1. Run the OO Shutup tweak

**Expected behavior**
Taskbar search box setting should be preserved and kept on the same setting as it was.

4yman-el:
What version of Windows are you using?

p0ll3:
Just Experienced the same issue. 
Windows 10 Pro 22H2

jonesy-b-dev:
> What version of Windows are you using?

Windows 11 23h1 I belive , I'll get my exact version tomorrow

4yman-el:
I found a setting in OOSU10 called "Disable search box in task bar" turned OFF in the OOSU configuration file that the script uses. You can run OOSU yourself and change it.

4yman-el:
OOShutUp enables the search bar, not winutil itself.

p0ll3:
> > winutil shouldnt be changing this setting
> 
> OOShutUp enables the search bar, not winutil itself.

If I apply the default ooshutup settings with the ooshutup program, it does not enable the Searchbar. 

Running the ooshutup settings via winutil does.

4yman-el:
>If I apply the default ooshutup settings with the ooshutup program, it does not enable the Searchbar.

My copy of OOSU has "disable search box" set to "limited". if you configure it yourself without a config file, it will import settings from Windows but if you only apply recommended settings it enables the search box.

>Running the ooshutup settings via winutil does.

Yes, OOSU overwrites all settings when run with a config file.

jonesy-b-dev:
How do i edit the OOSU?

4yman-el:
It's inside [winutil.ps1](https://github.com/ChrisTitusTech/winutil/blob/2b80e14bf954e2cba07903e0c6822677fb97c3d9/winutil.ps1#L11927) and [ooshutup10_recommended.cfg](https://github.com/ChrisTitusTech/winutil/blob/main/config/ooshutup10_recommended.cfg)

4yman-el:
Here is a config file that include only recommended settings and remove the search bar: [ooshutup10.cfg](https://github.com/user-attachments/files/15791575/ooshutup10.cfg.txt)

[...]

Issue #2027, opened 2024-06-04
Issue with clipboard history


Chappycpuequalszero:
**Describe the bug**
When using tweaks to disable Essential Tweaks even if I don't select it, it sometimes just disables clipboard history and if I try to undo the tweak it doesn't do anything. Windows settings just says "This setting is managed by your organization" 

**To Reproduce**
1. Go to 'Tweaks'
2. Click on 'Disable activity history'
3. Click on 'Undo Selected tweaks'
4. Check if it did anything.

**Expected behavior**
I expect it to allow activity history but it doesn't seem to change any of the settings surrounding it especially not clipboard history.

**Screenshots**
<img width="796" alt="ErROR" src="https://github.com/ChrisTitusTech/winutil/assets/64427403/c95b02a2-8b52-474c-bc7a-e6360f81f438">

**Additional context**
I have no organization or other setting management software (Besides this one)

Marterich:
This has been answered in https://github.com/ChrisTitusTech/winutil/issues/1871#issuecomment-2074847855.

Issue #2012, opened 2024-05-28
unable to re-enable clipboard history 


freaked1234:
Hey!
First off, I really love your tool and it is great help setting up new installs.
On one of my Win11 pro installs I have an issue though I can't seem to fix myself:
On accident I disabled clipboard history and even after undoing the corresponding changes in the tool it won't work.
I tried adding the registry keys manually but no success either. When trying to enable it via windows settings it says "this setting is not available because of your organisations guidelines", even though this machine is not part of an Organisation.
Any help would be appreciated!

og-mrk:
@freaked1234 
This has been answered in [this comment](https://github.com/ChrisTitusTech/winutil/issues/1871#issuecomment-2074847855).

which is done by undoing the **OO Shutup tweak** and restarting, then going into the settings app to re-enable the Clipboard History feature (if it hasn't been enabled).

**_Or_** by clicking **OO Shutup advanced tweak** button, where WinUtil will launch the GUI of OO Shutup so you can enable Clipboard History for _both_ the current user and the machine (all users),  exit out of OO Shutup and WinUtil and restart your computer, then it should work just fine 👍

freaked1234:
@og-mrk 
ofc this is the typical answer of people who dont want to admit fault but I tried that before in all honesty and it did not work. 
Somehow I just applied what you suggested, rebooted and the setting did not apply. Doing it again finally worked. 
Thank you very much for taking the time!

og-mrk:
I'm glad this fix worked for you @freaked1234 😄

it's for now a temporary one. This issue shouldn't typically happen.. or at least inform the user of what to expect after applying OO Shutup tweak, so they (the user) don't get surprised by certain features of windows not working anymore.

@freaked1234, If you have any more questions in mind, feel free to post them here, and if you don't have any, then make sure to close this issue as completed, thanks in advance 😄

PR #2099, opened 2024-06-17
Remove oosu essential tweak

Marterich:
Resolves: #2085

As explained in the Linked Issue, I think the time has come to remove the OOSU Checkbox from the Essential Tweaks.
The Button under Advanced Tweaks remains, but the wording was changed so that it makes more sense even without the previous simple checkbox tweak. 

- It was unclear to the users what changes were applied by the function which generated lots of Issues regarding disabled functionality. 
- OO Software seems to have included more functionality that users actually want to use in their "recommendations"

For the full reasoning behind this change, please take a look at the Issue.  #2085

Prevents issues like this in the future
Closes: #2047
Closes: #2027
Closes: #2012


og-mrk:
The changes look great, and it's quite nice of you to propose the idea _first_ to get feedback from other users and contributers 👍

I guess that the `.cfg` files in `config` folder is not needed if this PR gets merged, may be removing those `.cfg` is worth doing? To remove unused files and drope support for them (still remember that O&O Shutup didn't work for a long time, up until the config files got updated 😅

It could be removed in a separate PR.. but I don't see why it wouldn't be possible to do it here.. but anyhow, good work as always @Marterich 👍

Marterich:
> The changes look great, and it's quite nice of you to propose the idea _first_ to get feedback from other users and contributers 👍
> 
> I guess that the `.cfg` files in `config` folder is not needed if this PR gets merged, may be removing those `.cfg` is worth doing? To remove unused files and drope support for them (still remember that O&O Shutup didn't work for a long time, up until the config files got updated 😅)
> 
> It could be removed in a separate PR.. but I don't see why it wouldn't be possible to do it here.. but anyhow, good work as always @Marterich 👍

> If you look real close at the changed files you can see that the files are removed as well 😜
The only thing I didn't remove was the cfg parsing bit in the compile script

og-mrk:
> > The changes look great, and it's quite nice of you to propose the idea _first_ to get feedback from other users and contributers 👍
> > 
> > I guess that the `.cfg` files in `config` folder is not needed if this PR gets merged, may be removing those `.cfg` is worth doing? To remove unused files and drope support for them (still remember that O&O Shutup didn't work for a long time, up until the config files got updated 😅)
> > 
> > It could be removed in a separate PR.. but I don't see why it wouldn't be possible to do it here.. but anyhow, good work as always @Marterich 👍
> 
> If you look real close at the changed files you can see that the files are removed as well 😜
> The only thing I didn't remove was the cfg parsing bit in the compile script

Yahh I just noticed that 😂 viewing Git patches in the middle of the night (inside a smartphone) is not the best thing to do 😅

Also the compile bit could be done quite easily (_famous last words_), I'll make a PR to your Fork, give me a minute while I open my laptop...

ChrisTitusTech (changes requested):
We had some changes to the compile script and I believe this one isn't up to date with that. Can you pull the changes and recommit?

ChrisTitusTech (approved):

4yman-el:
Niiiiice! :rocket:

Issue #2096, opened 2024-06-17
"Kill OneDrive process" falsly shown in console


stuckydev:
**Describe the bug**
"Kill OneDrive process" is shown in console, but wasn't chosen in winutil at all.

**To Reproduce**
1. Go to 'Tweaks'
2. Click on 'Essential Tweaks' and "Advanced Tweaks" as shown in screenshot attached.
3. See wrong entry in console "Kill OneDrive process"

**Expected behavior**
only do what is selected - not show any other logs.

**Screenshots**
![cO1052XMC0](https://github.com/ChrisTitusTech/winutil/assets/121982728/4b7a59a0-cf49-451b-a457-3113512a65b3)

stuckydev:
Thx @Marterich , yes this will solve this.

PR #2100, opened 2024-06-17
Remove wrong Log Message

Marterich:
The original creator of this tweak seems to have copied the OneDrive tweak to reuse the code (which is totally fair) but forgot to remove all Onedrive references,
Also the Undo Script didn't make any sense regarding undoing the change, so I replaced it with a simple message telling the user  that Undo isn't implemented for this tweak
 
Resolves: #2096

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2413,7 +2413,6 @@
     "Order": "a0015_",
     "InvokeScript": [
         "
-        Write-Host \"Kill OneDrive process\"
         $serviceName = \"LMS\"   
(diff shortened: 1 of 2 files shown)


og-mrk:
Thanks for addressing this @Marterich, noticed it (and was curious behind such a decision) in PR #2051, the Copy-Past of the tweak is a reasonable answer to suggest, as I can't see any other reason why it might be in the Undo Script section.

ChrisTitusTech:
I'm changing this in commit 4661bf31baeb5ed2bb19ee2ff28d91eb59fdbe75

PR #2102, opened 2024-06-17
Remove unused Functions

Marterich:
Removes Two Functions that are referenced nowhere and therefore only add clutter and bloat and serve no purpose


ChrisTitusTech (approved):
Thanks for spotting this!

PR #2111, opened 2024-06-18
Generate Panels at runtime / Reduce Filesize of winutil

Marterich:
This PR moves the generation of the Xaml Code to display the Panels from the Compile script to winutil itself to be built at runtime. 

This has a few advantages over the previous solution: 
+ The compiled winutil.ps1 file is shrunk by about 150 KB (Before 884KB, After 728KB) 
+ The line count is reduced by about 1200 Lines in the compiled script. (Before 15323, After 14090)
+ The amount of duplicate information in the compiled script is reduced which adds clarity.
+ Smaller File = Faster Download = Faster Start

Downsides: 
I measured the time it takes to build the panels at runtime which came out to add about 250ms. In my case, this wasn't noticeable though, because Powershell + WPFApps do take a moment to start eighter way, and taking into account the time saved downloading the script, this could very well still be faster on a slow internet connection.

IMHO:
The added time is a tradeoff worth taking, considering all the pros mentioned above.


ChrisTitusTech:
Yeah this is a hard decision to make. It can be beneficial, but on a potato machine with fast internet it would be slower. Since most of the internet falls in to this category I'm going to close this as it is too close to call.

PR #2120, opened 2024-06-19
Fix White space (bottom right of Application Tab)

MyDrift-user:
When I use WinUtil in dark mode I always get a small offcolored rectangle on the bottom right of the toolbox. I did following things to fix it:

- created rectangle, color it with backgroundcolor and place it at the blank space.
- added xName to Application ScrollViewer to get the state of it's scrollbars
- added style for rectangle that triggers it to show if both scrollbars are visible

Before:
image

After:
image

bc of the trigger the rectangle will not show if not needed so it won't brake anything else.

### DIFF
--- a/xaml/inputXML.xaml
+++ b/xaml/inputXML.xaml
@@ -571,6 +571,18 @@
                 </Setter.Value>
             </Setter>
         </Style>
+        <Style x:Key="ScrollVisibilityRectangle" TargetType="Rectangle">
+            <Setter Property="Visibility" Value="Collapsed"/>
+            <Style.Triggers>
+                <MultiDataTrigger>
+                    <MultiDataTrigger.Conditions>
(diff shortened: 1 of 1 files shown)


blusewill:
The file you just edited it's auto generated. You should be looking other files that contains the code

og-mrk:
> The file you just edited it's auto generated. You should be looking other files that contains the code

It actually isn't an auto generated file, you can look at [these lines of code](https://github.com/ChrisTitusTech/winutil/blob/main/Compile.ps1#L74-L93) inside the `Compile.ps1` script, which in the firs 2 lines of the ones I've referenced, you'll notice that `input.xaml` is actually used to determine how will the GUI will look/behave.

Hope this provides some clarity on this subject, and if you have any questions in mind, feel free to ask them here, or more preferably through GitHub's discussions feature.

blusewill:
This is called don't get too outdated in winutil moment XDDD

ChrisTitusTech (approved):

PR #2123, opened 2024-06-19
Update applications.json

Real-MullaC:
Added Nvidia Geforce Experience


blusewill:
Well this should be not worthy since we have NVCleaninstall.
And you can also install GFE in that program.

Real-MullaC:
> Well this should be not worthy since we have NVCleaninstall. And you can also install GFE in that program.

NVCleaninstall allows you remove things from the driver. Whereas GFE does the full driver. Someone did in a issue post about applications they want. Said they want GFE as a standalone on the winutil.

Issue #2127, opened 2024-06-20
Microsoft Edge removal Issue 


webdrakib:
**Describe the bug**

**To Reproduce**
1. Microsoft Edge doesn't remove completely like before 
2. Please check the error 

**Expected behavior**

**Screenshots**

**Additional context**

Real-MullaC:
@webdrakib If this issue is fixed please feel free to free to close this issue.

Issue #2119, opened 2024-06-19
Add kicad to the installer


Sergiogd112:
Kicad is an open-source EDA tool. It's a good starting point for those who want to do electrical design and is even used by professionals in the industry.

[https://www.kicad.org/](https://www.kicad.org/)

The project is on winget under the ID KiCad.KiCad.
![image](https://github.com/ChrisTitusTech/winutil/assets/34186354/ff1e8752-7ae5-45d9-97bb-5969efc0e67b)

Real-MullaC:
Will be fixed in PR #2131

PR #2129, opened 2024-06-20
Add Geforce Experience and Task View Tweak.

Real-MullaC:

PR #2128, opened 2024-06-20
Update tweaks.json

Real-MullaC:
Fixed issue with Edge not removing.

Fixes #2127

PR #2131, opened 2024-06-20
Fixed edge removal and added kicad as an app

Real-MullaC:
Edge removal syntax was wrong so failed compile.

Kicad app was requested in issue #2119

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -311,7 +311,7 @@
 		"link": "https://copyq.readthedocs.io/",
 		"winget": "hluk.CopyQ"
 	},
-	"WPFInstallditto": {
+	"WPFInstalldittoclipmanager": {
 		"category": "Utilities",
 		"choco": "ditto",
 		"content": "Ditto (Clipboard Manager)",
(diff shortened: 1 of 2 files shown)


og-mrk:
Thanks for these fixes and additions @Real-MullaC 😄

Although I would personally make two separate PRs, one for the `applications.json` file, and another for the Edge Removal Tweak Fix.. but this would generally be fine, as long as the maintainer (Chris Titus) accepts these two changes in this PR.

with that being said, I would like to humbly request that you close the follow PRs:
1. PRs https://github.com/ChrisTitusTech/winutil/pull/2128 & https://github.com/ChrisTitusTech/winutil/pull/2130 , as these two PRs tries to fix the Edge Removal Tweaks, which should be fixed in this PR.
   (Fixes issue https://github.com/ChrisTitusTech/winutil/issues/2127)
2. PR https://github.com/ChrisTitusTech/winutil/pull/2129 , which is an attempt to re-open PR https://github.com/ChrisTitusTech/winutil/pull/2123 but has faced some issues in git (happens to experienced Git Users as well, no need to worry about it 👍).
   I suggest that you make a new PR that only adds the Geforce Experience App into `applications.json` file, and see if the maintainer accepts it or not.

</br>

Also.. may I ask why is the first commit (and only commit at the time of writing) is authored by a GitHub User with a Username of [Callum Janes](https://github.com/hubster-bot).. sounds a bit odd IMO, If you could provide some clarification behind this.. rather weird authoring thing (at least to me), then it'll be much appreciated.

Issue #2116, opened 2024-06-19
Pop-Up notification when Tweaks are finished


LoopTJ:
**Is your feature request related to a problem? Please describe.**
I would prefer to have a pop-up window open when "Installs have Finished" and when "Tweaks are Finished" so that I know I can continue with the next process in the utility.

**Describe the solution you'd like**
Once the Installs and Tweaks are finished present a pop-up window letting me know they've completed. Similar to the pop-up when the "Updates are Finished".

**Describe alternatives you've considered**
Currently I have to switch back to the terminal window to see when the process is finished. I find the pop-up easier as it is a notification that the process completed, this is especially helpful when running the utility on a remote computer since I'm not actively using the computer, just running setup scripts.

**Additional context**
Since some people won't want pop-ups, if possible have a toggle to turn on / off pop-ups for Installs, Tweaks, and Updates.

Real-MullaC:
Will be fixed in PR #2133 and #2134

PR #2133, opened 2024-06-20
Added pop up messages for installs

Real-MullaC:
Requested in issue #2116

### DIFF
--- a/functions/public/Invoke-WPFInstall.ps1
+++ b/functions/public/Invoke-WPFInstall.ps1
@@ -49,6 +49,12 @@ function Invoke-WPFInstall {
                 Install-WinUtilChoco
                 Install-WinUtilProgramChoco -ProgramsToInstall $packagesChoco
             }
+            $ButtonType = [System.Windows.MessageBoxButton]::OK
+            $MessageboxTitle = "Install"
+            $Messageboxbody = ("Installs have finished!")
+            $MessageIcon = [System.Windows.MessageBoxImage]::Information
+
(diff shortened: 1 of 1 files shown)


Marterich:
How does this behave for users who want to automatically apply an exported winutil config? Will this _block_ the execution of the config import and therefore require user input before any further tweaks get applied? 
If I recall correctly when importing a config, the applications are installed first.
If so, this would break the "zero touch" workflow for importing config files

Real-MullaC:
> How does this behave for users who want to automatically apply an exported winutil config? Will this _block_ the execution of the config import and therefore require user input before any further tweaks get applied? If I recall correctly when importing a config, the applications are installed first. If so, this would break the "zero touch" workflow for importing config files

It shouldn't affect it as it was only added to say that installs are done at the end.

Marterich:
All right, I've taken another look at the code, and I have two improvements. 

1. Move the line `$sync.ProcessRunning = $False` in front of the Popup. 
-- The installs are started in a runspace so this would block execution until the button is pressed. (This doesn't become immediately clear, because the program installed are started last when importing a config)
2. Move the Console Output in front of the Popup as well. (Also relevant for #2134)
-- With the console output _after_ the Popup, the terminal message will only be printed once the user clicks on the popup, so your order of Log-Messages in the terminal can get messed up if you wait.

Real-MullaC:
@ChrisTitusTech I do not know how to re commit after changes to this PR.

ChrisTitusTech:
You can move it to draft form then bring it back. 

As far as the PR, I don't think a popup is a great solution and I'm struggling to find what would be. I think the best course from here is something that changes the GUI application directly that the user is staring at.

Issue #2125, opened 2024-06-20
OnDrive removal issues.


charlescgs:
**Describe the bug**
OneDrive is not being totally removed after running `Remove OneDrive` tweak.
Program still shows in app list, in `winget upgrade` and in the Explorer as not-removable pinned shortcut.
Also while running a tweak an error pops-up.

**To Reproduce**
1. Run `Remove OnDrive` tweak.
2. Check explorer, App List in Settings and winget list.
3. See the OneDrive there.

**Expected behavior**
Tweak should not error and after it OneDrive should be gone for good.

**Screenshots**
## Run it:
![image](https://github.com/ChrisTitusTech/winutil/assets/135472912/67397464-ac04-42ad-973f-b92f6af0d82a)

## See the error:
![image](https://github.com/ChrisTitusTech/winutil/assets/135472912/1cc11abf-fd4b-40b9-a151-dee7899bc7e9)

## Explorer:
![image](https://github.com/ChrisTitusTech/winutil/assets/135472912/d3269fa6-d112-41f1-b230-f2dd15ce7746)

## App List:
![image](https://github.com/ChrisTitusTech/winutil/assets/135472912/bdf7b064-4b79-42ef-81e2-1a74bff73571)

## WinGet upgrade:
![image](https://github.com/ChrisTitusTech/winutil/assets/135472912/8121be2f-6d6b-40ff-bce6-54ac2cca5d5d)

og-mrk:
Hello @charlescgs 
First of all, thanks for reporting this issue, so that contributors get informed of such edge cases we didn't have the opportunity to replicate.

Now, looking at the screenshots you'vs shared, two things come to my attention..
1. Robocopy (a CLI tool this tweak uses) has failed for some reason, giving an Access Denied error when trying to copy the OneDrive folder found in the User's Home Directory.. maybe because the folder owner has been set to something the script hasn't anticipated/excepted.
2. As far as I can remember (_not sure, maybe someone can correct if I'm wrong_), is that this tweak uses WinGet (a package manager used in Windows, and is used in multiple places in WinUtil).. for whatever reason, it couldn't find a way to uninstall OneDrive. Hence the second error shown in red color indicates it has failed (**Application not found**).. There isn't a fallback for this case, so OneDrive hasn't been uninstalled from your system after running the OneDrive Tweak.

Unfortunately, I don't have the time to figure out solution(s) for the above problems. Maybe someone more experienced can take this issue to come up with a good solution.

[...]

Issue #2089, opened 2024-06-13
A proper cleanup is required when winutil removes onedrive , microsoft copilot and edge


MrDisguised:
**Describe the bug**

After removing one drive from my pc I realised that this script does indeed remove the onedrive application but there are many registry entries related to onedrive left that needs to be removed properly (a proper ""clean-up"" needs to be done). For example :- 

`HKEY_CURRENT_USER\Software\Microsoft\OneDrive`
`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\OneDrive`
`HKEY_CLASSES_ROOT\CLSID{018D5C66-4533-4307-9B53-224DE2ED1FE6}` : This registry key contains information related to the OneDrive shell extension.
`HKEY_CURRENT_USER\Software\Policies\Microsoft\OneDrive`
`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\OneSyncSvc`
`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\OneSyncSvc_365c8a`

.... And many more. You can open the `logon` tab of the `autoruns` application and you can see for yourself all the registry entries that needs to be removed. The Onedrive 1, Onedrive 2, Onedrive 3, Onedrive 4 , Onedrive 5 , Onedrive 6 and the Onedrive 7 registry folders.

**Expected behavior**
I expected that this script will remove all the registry entries after I remove the one drive application.

MrDisguised:
I also think that the Microsoft edge and the Microsoft copilot registry entries should also be removed properly if someone removes them using this script. A proper clean-up is really necessary in my opinion.

Marterich:
Definitely a fair argument. When implementing this, the undo action should be considered as well to restore a fully working system. (Maybe it's necessary to backup the keys somewhere to facilitate the restore option)
This will have to be investigated further

MrDisguised:
> Definitely a fair argument. When implementing this, the undo action should be considered as well to restore a fully working system. (Maybe it's necessary to backup the keys somewhere to facilitate the restore option) This will have to be investigated further

I really don't have much idea about windows tbh lol. I am more of a linux guy but afaik winutil creates a system restore point by default before making changes to our pc. So if something goes wrong we can easily revert back to the old snapshot of our windows that was created by the winutil script.

[...]

PR #2155, opened 2024-06-23
Rework OneDrive removal Tweak

Marterich:
Resolves: #2125
Closes: #2089

This is my take on getting the OneDrive Removal tweak to run again and make them a bit more stable.

Changes:
- Queries the Registry to run the uninstallation locally. (The uninstallation using newer binaries from winget sometimes seemed to be the source of errors)
- Doesn't kill the Explorer and Onedrive services beforehand
- Change the script to only move downloaded files to the new document folder
- Changed the order to uninstall Onedrive **before** moving the files. This should fix a problem where Onedrive would error when trying to load files while the process was stopped.

Additions: 
- Only Execute all the manual Cleanup task if the uninstallation was successful. (At least successful enough to remove the Control Panel Reference which is most of the times one of the last things to be removed)
- Added/modified a few warning messages

Note: I have only tested this on Windows 11 Pro, but I see no reason why this wouldn't work on different Windows Versions

IMHO: 
The Git Diff view SUCKS for Tweaks that are fully scripted in the JSON file.
Should we start moving over to writing the Tweaks in separate functions/files that take arguments for the action to be done? (like I did previously with Invoke-OOSU and Invoke-TweakPS7) 
Moving these to separate functions would greatly improve the readability of the project.


ChrisTitusTech:
The main issue with a move is if OneDrive is running while the move happens for any reason it will REMOVE the files from onedrive and I need to double-check to make sure this never happens.

Marterich:
@ChrisTitusTech this isn't a problem anymore because the uninstallation is run **before** the files are moved. This way nothing is removed from the cloud, the virtual online files are removed and afterwards when OneDrive is uninstalled and therefore stopped, the files are moved

Marterich:
If you want to be really sure, just add the taskkill back in right before the move operation. The only thing to keep in mind is, that this will most likely always fail because OneDrive will be uninstalled already, so make sure that it doesn't trigger the try catch block

ChrisTitusTech (approved):

PR #2160, opened 2024-06-25
Modify Release Workflow

Marterich:
This PR modifies the Release Actions. 

### Modified Full Release
Full releases won't be created automatically on merge of the main branch anymore, but have to be started manually
 - A stable release will update/override the last automatically generated pre-release

### Added Pre-Release 
Adds an Action that automatically creates a release tagged as pre-release on every main branch merge. These will appear in the release overview but can be clearly distinguished from a stable release by the orange pre-release label.

image

**Optimization Potential for the Future if this proves to be working for winutil**

- Autogenerate a meaningful description for the pre-releases
- Remove all previous pre-releases on a new stable release and clean up the tags
- Autogenerate a description for the stable release, containing the info of the changes since the last stable release


MyDrift-user:
I am currently trying out auto generated descriptions for the releases/pre-releases using `release-drafter/release-drafter@v5`

With that it also seems to create tags that can be applied to PRs so that the PR list is more readable.

Releases should look something like following (Release / Pre-Release instead of Draft):

image

Thoughts on that?

Marterich:
@MyDrift-user 
I think the automation you mentioned could be a good starting point to build a system in the future that fills the description. 
Getting this to run well so it saves time and effort and needs no manual intervention will be quite the challenge but I'm all for experimenting and kicking around new ideas and eventually sending in a PR when it seems to be stable. 
I myself would probably not invest too much effort in polishing the solution right now, even before we know if the pre-release workflow will work at all for Chris but in the future, I definitely see something like this as an option

ChrisTitusTech (approved):
Wonderful! This is exactly what I needed to change. Thank you @Marterich

PR #2166, opened 2024-06-25
AveYo one no longer works with me, so i made one that works

SteavenGamerYT:


blusewill:
Can you add more to the description so we can know what tweaks have you done?

ChrisTitusTech:
Going to make a seperate ps1 script to do this. I don't want to replace it with a cmd that isn't quite there yet.

PR #2162, opened 2024-06-25
Add Search Button Toggle in the Tweaks Section

blusewill:
Adding the Toggle of the Search Button on the Taskbar.

https://github.com/ChrisTitusTech/winutil/assets/62820836/2e052cfc-c3cc-464e-a8b2-d7b6be8039bd

PR #2168, opened 2024-06-25
Added Search Box and started scripting change.

Real-MullaC:
Could someone please add whatever is missing to make this work.

Making tweak for Search Box so it is a dropdown for multi options.

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -3113,6 +3113,14 @@
     "Order": "a081_",
     "Type": "300"
   },
+  "WPFSearchBox": {
+    "Content": "Search Box",
+    "category": "z__Advanced Tweaks - CAUTION",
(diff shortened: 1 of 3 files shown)


blusewill:
I already done it in #2162 and on Some version of Windows don't have Big box option. So I just keep it a toggle that way

Real-MullaC:
> I already done it in #2162 and on Some version of Windows don't have Big box option. So I just keep it a toggle that way

I added it as a drop box so people have multiple options. But someone could add in something to alert the user if it won't work on there system. So people have access to all options. As some people will want a search box not search icon but some people might choose to have it hidden. That was my logic with it.

Real-MullaC:
@og-mrk Would you be able to see if this works if not fix it. As not sure how to.

PR #2169, opened 2024-06-25
Purging applications

Owen-3456:
Removed the following applications that I think are bloating the app list:

[WinDirStat](https://windirstat.net/) - WinDirStat is outdated. WizTree should be used instead

[BitComet](https://www.bitcomet.com) - BitComet is adware and has a history of unethical practices. [VirusTotal](https://www.virustotal.com/gui/file/7ccef9af5267c22a56bdbaf2f9109a02611bba461e0b0321bed42b5911163272), I don't know if these are false positives or not but there are so many better options

[BlueStacks](https://www.bluestacks.com/) - BlueStack 5's installation process forcibly install BlueStacks X and a crypto wallet alongside it

Additionally I have removed a duplicate entry of Ditto and changed the link on the entry remaining

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -151,14 +151,6 @@
 		"link": "https://github.com/sharkdp/bat",
 		"winget": "sharkdp.bat"
 	},
-	"WPFInstallbitcomet": {
-		"category": "Utilities",
-		"choco": "bitcomet",
-		"content": "BitComet",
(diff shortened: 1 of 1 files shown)


ChrisTitusTech:
Thanks for helping prune the old packages @Owen-3456

ChrisTitusTech (approved):

PR #2153, opened 2024-06-23
Add New Escape Character 'Em Dash' to 'Compile.ps1' File

og-mrk:
### Commit Description Message

This will Insure that an App Entry can use the '—' Character (aka. the 'Em Dash' Character) in their Content (Title) and Description Fields without any worries.

### Screenshots

An example of an effected app entry is Element's Description.

Before the changes.
image

After the changes.
image

### Issues this PR tries to resolve

Resolves: https://github.com/ChrisTitusTech/winutil/issues/2151

### References

- A list of HTML Entity Codes by [FreeFormatter.com](https://www.freeformatter.com/) website: https://www.freeformatter.com/html-entities.html
- Searching through Unicode Characters using [unicodelookup.com](https://unicodelookup.com/) website: https://unicodelookup.com/#em%20dash/1

PR #2171, opened 2024-06-26
Typo Fixed in Description of "Element"

SHR1SHAK:


og-mrk:
@SHR1SHAK This will be fixed in PR https://github.com/ChrisTitusTech/winutil/pull/2153

The dash instead of an end quote makes sense, as the description after the dash is for the Matrix (the technology which Element is using)

SHR1SHAK:
But out the GUI it changes to "?"
Screenshot (1)

SHR1SHAK:
Sorry, I didn't see it was already resolved. Just new to it and learning stuff.

og-mrk:
It's ok, @SHR1SHAK 
Happens from time to time, even for the best of us 😄

Issue #2079, opened 2024-06-12
VSCodium isn't installing


bckelley:
**Describe the bug**
VSCodium isn't installing despite proper params given, don't know why git is listed git is already installed.
![image](https://github.com/ChrisTitusTech/winutil/assets/15696431/44a9c85c-4336-40fd-ba25-39afb30c7e1a)
![image](https://github.com/ChrisTitusTech/winutil/assets/15696431/0715fb11-38a2-4c6c-965b-5c07a51d3d35)

**To Reproduce**
1. double click WinUtil shortcut
2. check VS Codium
3. click Install/Upgrade Selected

**Expected behavior**

**Screenshots**

**Additional context**

blusewill:
![Screenshot_20240615_110603_GitHub.jpg](https://github.com/ChrisTitusTech/winutil/assets/62820836/e9b9fe4f-e649-4bc4-963d-d98a7fa00b5f)

Who the hell named the packages like this....

Owen-3456:
This will be fixed in the next release
PR: #2172

bckelley:
I worked thank you don't understand why the user credentials box when my pc's user wasn't even listed??

PR #2172, opened 2024-06-26
Applications.json tweaks

Owen-3456:
Tweaks:
- Change Go title and description to make more sense
- Removed trying to install git along side some packages which causes issues (#2079). Dependencies should be installed automatically
- Update DevToys winget ID

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -802,8 +802,8 @@
 	"WPFInstallgolang": {
 		"category": "Development",
 		"choco": "golang",
-		"content": "GoLang",
-		"description": "GoLang (or Golang) is a statically typed, compiled programming language designed for simplicity, reliability, and efficiency.",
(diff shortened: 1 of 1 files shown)


og-mrk (approved):
### Some feedback on the PR

- The update to DevToys is to a another WinGet ID is valid, **see screenshot below**.
- removing the `Git.Git` for several other Apps is something I'm kinda wondering how it happened in the first place 😂 , and your argument on Packages should resolve needed dependencies by their own is the logical thing, as WinUtil isn't a WinGet Front-End GUI (It isn't our responsibility to resolve dependencies for programs.. it's the Package Manager's job to do this)
- the URL Update to GoLang's website isn't really necessary.. as the old URL redirects to the new one automatically.. but it's a nice change nonetheless 👍

### Screenshots

image

PR #2177, opened 2024-06-26
Update README.md so it gives people the latest release. Rather them having to add the release they want.

Real-MullaC:


og-mrk (approved):
Tested it on my side, and indeed it does work without a problem.

Good Commit(s) @Real-MullaC 😄

PR #2196, opened 2024-06-28
Add Overlay to Taskbaritem

MyDrift-user:
Since I and other did not find a way to correctly change the Icon of the Taskbaritem, i added the icon as an overlay.

Preview:
image
image

Edit:
- Removed old commented-out code to set Taskbar Icon since it does not work.


ChrisTitusTech:
Now that is pretty cool! Thanks @MyDrift-user

ChrisTitusTech (approved):

PR #2204, opened 2024-06-28
Change WinGet ID for 'Fastfetch' to be 'Fastfetch-cli.Fastfetch' instead of 'fastfetch'

og-mrk:
### Commit Description Message

This will ensure no ambiguity would happen in the near future for this app entry.

### References

This PR is a follow-up to https://github.com/ChrisTitusTech/winutil/pull/2159 PR


ChrisTitusTech (approved):
Good catch

PR #2219, opened 2024-06-29
Pump up the value of 'operations-per-run' Field for 'Close Old Issues' Workflow

og-mrk:
### Commit description message

Provides more 'head-room' for the 'stale' Actions to process all available Issues, as well as PRs.. because they're processed no matter what, and I can't force this action not to.. as far as I know.

### References

- This will also remove the warning shown when this workflow runs shown in the Annotations Section.
  [link to latest one at the time of writing](https://github.com/ChrisTitusTech/winutil/actions/runs/9719913811)
- The docs for `operations-pre-run` Field:
  https://github.com/actions/stale#operations-per-run

### DIFF
--- a/.github/workflows/close-old-issues.yaml
+++ b/.github/workflows/close-old-issues.yaml
@@ -30,6 +30,10 @@ jobs:
         # Sends a message for both the Stale and Close events of an issue.
         stale-issue-message: "This issue was marked as stale because it has been inactive for 7 days"
         close-issue-message: "This issue was closed because it has been inactive for 7 days since it was marked as stale"
+        # Increase this value if the project receives a lot of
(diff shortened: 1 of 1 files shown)


ChrisTitusTech (approved):

ChrisTitusTech:
No idea this was a thing! Thanks @og-mrk

og-mrk:
> No idea this was a thing! Thanks @og-mrk

No problem @ChrisTitusTech, also it was designed so you don't reach the API Rate Limit set by GitHub, This limit is for all Actions/Workflows that runs by your Account, so I set it to `100` items to process (each might be take 1-2 API Requests.. not sure tbh), which's a reasonable amount IMHO, more than enough for the time being 👍

PR #2221, opened 2024-06-29
Improvements/Fixes for the Release Actions

Marterich:
- Changes prerelease to include the prefix "pre" in the Tag to never overwrite/change main releases. (otherwise if you create a main release in the morning and merge a PR in the evening, the main Release will be marked as a pre-release again)
- Fix: Adds the Tag "Latest" to the Main Releases so that it works as expected
- change windev.ps1 to explicitly search for pre-releases (this is done because Github does not necessarily order the releases by last modified, but rather by the tag and name)


ChrisTitusTech (approved):
Great catch. Thanks @Marterich

PR #2235, opened 2024-06-30
fix syntax highlighting in .MDs

MyDrift-user:
welp, took some inspiration from #2214 and changed all ps code snippeds in .md files that I've found to have the right syntax highlighting.


meenbeese:
Great, thanks! I did miss these ones.

ChrisTitusTech (approved):

PR #2240, opened 2024-07-01
Added Application United Sets

SHR1SHAK:
An open-sourced program that groups windows into tabs, inspired by the unreleased Sets feature. https://github.com/Windows-Apps-Hub/UnitedSets


ChrisTitusTech:
Interesting program, but still a bit too new to be added. We might add this in the future after it matures a bit and isn't beta.

SHR1SHAK:
Thanks for the feedback. I will add it after it's out of beta.

Issue #2241, opened 2024-07-01
Icon for WinUtil shortcut


RuiNtD:
**Is your feature request related to a problem? Please describe.**
The shortcut created by the "Create WinUtil Shortcut" button under Tweaks doesn't have an icon, defaulting to the generic PowerShell icon.

**Describe the solution you'd like**
I've noticed that [there is code to add an icon](https://github.com/ChrisTitusTech/winutil/blob/b47036630c817c07291c1e69a1ea979785a807e6/functions/public/Invoke-WPFShortcut.ps1#L30-L33) if `%TEMP%\cttlogo.png` present, but that path is never written to anywhere in the code, leaving it to default. I'd assume this file path is meant to be written to at some point, but the developers may have forgotten to do so, since they may already have the logo file at that path.

**Describe alternatives you've considered**

**Additional context**

og-mrk:
Thanks for reporting this issue @RuiNtD
Created a PR which fixes this problem, it'll be available in the next stable release once the PR gets merged 👍

PR #2242, opened 2024-07-01
Fix Icon Support for Shortcut Creation

og-mrk:
### Issues this PR tries to resolve

Resolves: https://github.com/ChrisTitusTech/winutil/issues/2241

### Screenshot(s)

image

### DIFF
--- a/functions/private/ConvertTo-Icon.ps1
+++ b/functions/private/ConvertTo-Icon.ps1
@@ -11,7 +11,7 @@ function ConvertTo-Icon {
         [Parameter(Mandatory=$true)]
         $bitmapPath,
         [Parameter(Mandatory=$true)]
-	$iconPath
+        $iconPath
     )
 
     Add-Type -AssemblyName System.Drawing
--- a/functions/private/ConvertTo-Icon.ps1
+++ b/functions/private/ConvertTo-Icon.ps1
@@ -21,11 +21,8 @@ function ConvertTo-Icon {
             ConvertTo-Icon -bitmapPath "$env:TEMP\cttlogo.png" -iconPath "$env:TEMP\cttlogo.ico"
         } catch [System.IO.FileNotFoundException] {
             # Handle the thrown exception here...
-        } catch {
-            # Print out the exception to console, and see what went wrong
-            Write-Host "Could not finish 'ConvertTo-Icon' due to unhandled exception"
-            Write-Host "Error: $_"
         }
+
         This Example makes a '.ico' file at "$env:TEMP\cttlogo.ico" File Path using the bitmap file
         found in "$env:TEMP\cttlogo.png", the function overrides the '.ico' File if it's found.
         this function will throw a FileNotFound Exception at the event of not finding the provided bitmap File Path.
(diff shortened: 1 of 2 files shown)


functions/public/Invoke-WPFShortcut.ps1
```diff
@@ -27,8 +27,10 @@ function Invoke-WPFShortcut {
             $ArgumentsToSourceExe = "$powershell '$IRM'"
             $DestinationName = "WinUtil.lnk"
 
+            Invoke-WebRequest -Uri "https://christitus.com/images/logo-full.png" -OutFile "$env:TEMP\cttlogo.png"
+
             if (Test-Path -Path "$env:TEMP\cttlogo.png") {
-                $iconPath = "$env:SystempRoot\cttlogo.ico"
+                $iconPath = "$env:TEMP\cttlogo.ico"
```
Marterich, line 33: The use of temp could be problematic in the future. When Windows eventually clears the temp folder, the file will vanish and the generated link could lose the icon again
I would probably opt for creating a folder "winutil" or something under appdata\local

Marterich (commented):
If I'm correct, the icon is used multiple times during the execution of winutil. There could be potential for optimization to make sure that if the icon already exists on disk, nothing needs to be downloaded and converted again

functions/public/Invoke-WPFShortcut.ps1
og-mrk: I do agree on this, not to mention that the Window showing delay _is actually_ because of the `<Image />` Element in `inputXML.xaml` File, which will reject handling any User Input, until this file is downloaded, so after working on it (did some experiments here and there), it works as expected.

For this PR, it'll be only these changes.

og-mrk: Also, thanks for the suggestion @Marterich 😄

functions/private/ConvertTo-Icon.ps1
```diff
@@ -2,19 +2,75 @@ function ConvertTo-Icon {
...
+        [string]$iconPath,
+        [Parameter(position=2)]
+        [bool]$overrideIconFile = $true
     )
 
     Add-Type -AssemblyName System.Drawing
 
     if (Test-Path $bitmapPath) {
+        if ((Test-Path $iconPath) -AND ($overrideIconFile -eq $false)) {
+            throw "[ConvertTo-Icon] Icon File is found at '$iconPath', and the 'overrideIconFile' Parameter is set to '$overrideIconFile'. Can not continue, quitting the bitmap to icon convertion..."
```
Marterich, line 72: Shouldn't this rather be only a Warning/Info rather then a blocking errror? 
```suggestion
            Write-Host "[ConvertTo-Icon] Icon File is found at '$iconPath', and the 'overrideIconFile' Parameter is set to '$overrideIconFile'. Skipping the bitmap to icon convertion..." -ForegroundColor Yellow
            return
```

functions/public/Invoke-WPFShortcut.ps1
```diff
@@ -27,9 +27,19 @@ function Invoke-WPFShortcut {
             $ArgumentsToSourceExe = "$powershell '$IRM'"
             $DestinationName = "WinUtil.lnk"
 
+            Invoke-WebRequest -Uri "https://christitus.com/images/logo-full.png" -OutFile "$env:TEMP\cttlogo.png"
+
             if (Test-Path -Path "$env:TEMP\cttlogo.png") {
-                $iconPath = "$env:SystempRoot\cttlogo.ico"
-                ConvertTo-Icon -bitmapPath "$env:TEMP\cttlogo.png" -iconPath $iconPath
+                $iconPath = "$env:TEMP\cttlogo.ico"
```
Marterich, line 33: I would still rather put this into a persistent location like "$env:LOCALAPPDATA\winutil" to make sure the shortcut never loses its icon even if the user doesn't use it for a while

functions/private/ConvertTo-Icon.ps1
```diff
@@ -2,19 +2,75 @@ function ConvertTo-Icon {
...
+        } catch [System.IO.FileNotFoundException] {
+            # Handle the thrown exception here...
+        } catch {
+            # Print out the exception to console, and see what went wrong
+            Write-Host "Could not finish 'ConvertTo-Icon' due to unhandled exception"
+            Write-Host "Error: $_"
+        }
+        This Example makes a '.ico' file at "$env:TEMP\cttlogo.ico" File Path using the bitmap file
+        found in "$env:TEMP\cttlogo.png", the function overrides the '.ico' File if it's found.
+        this function will throw a FileNotFound Exception at the event of not finding the provided bitmap File Path.
```
Marterich, line 31: Could all of this be handled right inside this function to make using this as easy as invoking it ConvertTo-Icon without the need for having to build this try - catch construct every time the function is used?

Marterich (commented):
Add a few notes in the code comments about what I think could still be optimized:)

MyDrift-user:
If we are already trying to fix the multiple requests to the same image, I would suggest modifying:
- the Snipped I added in #2196.
- The Logo from inputXML.xaml

og-mrk:
> If we are already trying to fix the multiple requests to the same image, I would suggest modifying:
> - the Snipped I added in #2196.
> - The Logo from inputXML.xaml

Thanks for pointing these issues out @MyDrift-user 
Thankfully, and after some luck, I'd solved these issues some time during this day (when initially working on this PR), but I haven't pushed & made a PR for it.

Will try and compose a new PR that will provide a base system for managing resources and, in particular, images.

functions/private/ConvertTo-Icon.ps1
og-mrk: This does make sense, as in the events of the caller of this function, if there's a `.ico`, then doing an early return is better, and somewhat safer, then doing an Exception Throw.

Thanks for the suggestion @Marterich 😄

functions/private/ConvertTo-Icon.ps1
og-mrk: Done in https://github.com/ChrisTitusTech/winutil/pull/2242/commits/e8ca90e29f251d92be0458da98a4fd8e02dc7057 commit

There's still the case of bitmap File not being found, raising an exception is better then silently continuing IMHO, as it'll force the developer who uses this function to always check before call `ConvertTo-Icon` Function.

Also added a new Example which highlights that it's better to Check if Bitmap File exists or not, if it exists then call the `ConvertTo-Icon` Function, otherwise don't.

Marterich (approved):
Looks good to me (haven't run it but only glanced at the code and the changes)

MyDrift-user (approved):
Looks good & works well on my end

ChrisTitusTech (approved):
We might try to switch out the png to svg as I think it would be a bit faster, but for now lets roll with this.

PR #2257, opened 2024-07-03
Move Kicad to "Multimedia Tools"

Joanty24:
Move Kicad to "Multimedia Tools" as the other CAD software is already there.


Joanty24:
> The suitable Category should be Development as "Kicad" is a Schematic Capture and PCB Development Software.

I think that "Development" is for software development as all the creation software is in "Multimedia"

Real-MullaC (changes requested):
Don't include the WinUtil.ps1 file as it might deleted it from the source code what will make the compile script fail.

ChrisTitusTech:
Can't merge due to winutil.ps1 deletion. I recommend discarding changes or just grabing the winutil.ps1 from the repo if you overwrite and commit direct.

Issue #2233, opened 2024-06-30
Winutil (Issues when installing)


DarkSideAm:
Hi All. I observed a minor issue with the Win utility while installing the selected apps from the installed programs; however, when installing, it says 1 of 24 and remains that way while the programs continue to install. It remains stuck like that during the entire process. Does anyone else have the same issues? The programs still install without problems. Sometimes VLC doesn't always install.

also can someone please explain why it's asking this about credentials it's only just recently that this is happening.

**Screenshots**
![Screenshot_1](https://github.com/ChrisTitusTech/winutil/assets/173422923/84d89b30-97af-4bb6-9f26-21df4091ee29)
![Screenshot_2](https://github.com/ChrisTitusTech/winutil/assets/173422923/dfcc1c6c-f8f9-4712-9dbf-8324a37b6c76)

@og-mrk

og-mrk:
@DarkSideAm, thanks for reporting this issue

The progress bar not updating is probably some logic error in the code, I'll look into it.

The pop-up window is asking to use credentials to install an app for another user because some apps, like Spotify, don't like to be installed in the Admin context.. so we need a way to install these apps in a reqular user mode, and the current solution isn't perfect by any means (I believe it actually interferes with unattended config use.. could be wrong on this), but it worked at the time, but it needs some improvements.

og-mrk:
@DarkSideAm
I'd like to inform you that I've created a PR which solved this issue, [link to it](https://github.com/ChrisTitusTech/winutil/pull/2271).

PR #2271, opened 2024-07-04
Change Main loop from 'ForEach' to 'For' inside 'Install-WinUtilProgramWinget' Private Function

og-mrk:
### Why change it to a simple `For` Loop?

because I couldn't find a way to change the variable `$x` on each iteration of the `ForEach` loop (and no it isn't because the x is capital in `$X++`.. already tried changing this, but to no avail ...)

So I've changed it to be a simple "C Like" For Loop, which has a variable `$i` initialized to the value of `0`, and on each iteration it'll increase by one (the `$i++` part).

If anyone can find a better solution then this.. then I welcome anyone to do so, and open a PR for it 👍

### Issues this PR tries to resolve

Resolves: https://github.com/ChrisTitusTech/winutil/issues/2233

### DIFF
--- a/functions/private/Install-WinUtilProgramWinget.ps1
+++ b/functions/private/Install-WinUtilProgramWinget.ps1
@@ -22,17 +22,18 @@ Function Install-WinUtilProgramWinget {
         [Parameter(Position=1)]
         [String]$manage = "Installing"
     )
-    $x = 0
+
     $count = $ProgramsToInstall.Count
 
     Write-Progress -Activity "$manage Applications" -Status "Starting" -PercentComplete 0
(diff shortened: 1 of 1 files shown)


Marterich:
Info:
If you desperately want to force the "Powershell way" or doing things, you could probably use a ForEach loop with the .indexof($_) function to get the element number and therefore a counter. 
But, as this is most definitely way more resource intensive (running a function to query the index of a PSObject vs running a simple binary operation to increment the counter) and also not much more readable, I prefer the "c" way at the moment

ChrisTitusTech (approved):

PR #2297, opened 2024-07-07
Copliot removal tweak update

Nikoskampourakis:
This added copliot removal tweaks haven't tested because i don't have a pc but it removes bing search removes copliot on edge and your task bar and your will never see it again

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -1,4 +1,15 @@
 {
+  "WPFCopliotremovepermanent": {
+    "Content": "Disable Windows Copilot",
+    "Description": "This disables Windows Copilot permanently even if windows updates.",
+    "category": "Essential Tweaks",
+    "panel": "1",
+    "Order": "a005_",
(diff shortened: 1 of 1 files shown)


Real-MullaC (changes requested):
You need to edit lines between 2353 and 2395 as that is where copilot is not make a new bit of code in the tweaks.json file. Please can you get this changed and make a new PR.

MyDrift-user:
There is no parameter "url" and blob links are not raw links.
Your code does nothing with Group Policy as far as I've seen. You are only editing registry entries, which can be configured like following example where @Real-MullaC said:

"registry": [
      {
        "Path": "HKLM:\\SOFTWARE\\Policies\\Microsoft\\Windows\\System",
        "Name": "EnableActivityFeed",
        "Type": "DWord",
        "Value": "0",
        "OriginalValue": "1"
      },

ChrisTitusTech:
I don't allow links to outside repo scripts.

PR #2301, opened 2024-07-08
Add IPv6 DNS

Joanty24:


Joanty24:
I discarded the deletion of the winutil.ps1 file.
I hope it is not to late and that you can resolve the conflict and merge it. If not I will open a new pull request for the next release, sorry for the inconvinience, trying to do things bettter!

PR #2303, opened 2024-07-08
Added IPv6 DNS

Joanty24:

Hope I've done the pull request well this time!


ChrisTitusTech (approved):
Good job! Thanks for adding ipv6 @Joanty24

PR #2359, opened 2024-07-15
Make the 'Youtube Tutorial' Header be the link instead of the Youtube Thumbnail for it, Making the 'README.md' even shorter

og-mrk:
As described in the title, it'll make the `README.md` file shorter when rendered on GitHub.

Plus it'll follow the surrounding headers in its "linking-style".


ChrisTitusTech (approved):
Probably better than a Picture, but I do want to add an animated gif in the future. This will work for now though, thanks!

PR #2376, opened 2024-07-16
Fix: Remove "pre" vom versioning of pre-releases

MyDrift-user:
# Pull Request

## Type of Change
- [x] Bug fix

## Description
Since Pre-Releases will be renamed to the full release if it gets approved by chris, the versioning with the pre does not work with the versioning link in the about section. It does not really help anything and causes more harm the way releases work rn. This also fixes the versioning link in pre-releases in general. For those reasons I suggest just removing it. 

Screenshot 2024-07-16 145108

## Checklist


Marterich:
This will also have the inverse effect.
Say, Chris creates a new stable release in the morning and merges a command to the pre-release at noon , it will overwrite the stable release and list it as a pre release again which is not the intended behavior. That's why the prefix was added in the first place

MyDrift-user:
Thanks for pointing this out @Marterich

PR #2449, opened 2024-07-24
Use proper TranslucentTB name

sylveon:
It is named TranslucentTB, not Translucent Taskbar


sylveon:
Why?

riverar:
I'm confused as to why @ChrisTitusTech would prefer the wrong name here. Wonder if this was closed in error? I pinged him on X too to increase visibility. https://x.com/WithinRafael/status/1948098065296216088

PR #2513, opened 2024-08-03
Refactoring Edge Removal

### LINKED ISSUE #2485: Uninstalling Edge properly
Removing Edge on the script seems to cause problems like not being able to reinstall properly however I've seen a way to do it proper and actually getting the edge uninstaller to pop up. This works on Windows 10 and 11. First you have to go to a specific registry location to remove the NoRemove option from "1" to "0" from Edge. Its located in
"HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Microsoft Edge" This allows the uninstall button to not be grayed out.
Next you have to get system access in System32 to a file called "IntegratedServicesRegionPolicySet.json". This file seems to control the policies for users in countries that can remove Edge. Inside the file there is a bit of text including "Edge is uninstallable" with it set to disabled. If you change it to enabled it is possible to be removed through programs and will give you a pop up asking if you want to uninstall Edge. This keeps WebView2 and I believe this method may stop Microsoft from reinstalling Edge with an update as it tricks the system into thinking its allowed to be removed (Had this done on my machine for 2-3 months and it hasn't reinstalled itself). This does leave some Edge behind but is a more stable way of removing as full removal can cause problems with games and apps like Minecraft Bedrock if you do a full removal. If this doesn't work I have found that if you add below "Edge is uninstallable" the country you are in "US" for example (There should be a list of countries under region" in the ISRPS.json file, add "US", in front) this should 100% ensure it gets removed as it tricks it more in thinking the United States region is allowed to remove Edge. Im not a developer but I would love to see this somehow implemented into the winutil.
Here is the video I saw of it being removed. https://www.youtube.com/watch?v=_utrsNxQiR0

Linget0:
# Pull Request

## Title
Removing Microsoft Edge similar to how users in the EU would be able to

## Type of Change
- [ ] Refactoring

## Description
Fixing Edge removal to trick the computer to think that the user is in the EU and is allowed to remove it. When running the Remove Edge tweak there will be a pop up from Edge asking the user if they want to uninstall and clicking Uninstall will remove Edge. This only removes Edge and keeps WebView2 and is how Microsoft would have users in the EU remove Edge and this shouldn't break users from being able to reinstall it if they want it again

## Testing
Tested on fresh windows installs with Edge installed 

## Impact
[Discuss the impact of your changes on the project. This might include effects on performance, new dependencies, or changes in behaviour.]
If edge is the only browser installed you wont have a browser so installing another browser beforehand is highly reccommended

## Issue related to PR
[What issue/discussion is related to this PR (if any)]
- Resolves #2485 

## Additional Information
Tested this in the US only but I would imagine this should work in other countries.  Code might not be the prettiest as this is my first pull request but I think this way of removal is much better than the previous as you are tricking your system into thinking your in the EU and allowing your system to use the built in Edge uninstaller instead of force removal that can break things especially trying to reinstall. 

## Checklist
- [ *] My code adheres to the coding and style guidelines of the project.
- [* ] I have performed a self-review of my own code.
- [* ] I have commented my code, particularly in hard-to-understand areas.
- [ *] I have made corresponding changes to the documentation.
- [* ] My changes generate no errors/warnings/merge conflicts.
- If you run Remove Edge tweak again after Edge is removed there will be a error saying it can't find setup.exe because Edge is already removed


Real-MullaC:
This is not a way to remove MS Edge from a computer. So this will not be accepted.

MyDrift-user:
Hey @Linget0 
In addition to what @Real-MullaC said;
- you should not make your changes in the winutil.ps1 file, nor create a winutil1.ps1

PR #2556, opened 2024-08-12
Ultimate Performance via GUID rather than name

AtisFPS:
# Powerplan improvement 

The issue appears to be related to the Windows language. I've noticed that the Ultimate Power Plan doesn't activate automatically because its name varies across different Windows languages.

I've created a small script that may help. It retrieves the GUID we just created, applies it, and changes the name and description simultaneously.

## Script info

- [x] Import the power plan using the GUID (same as the one for the Ultimate Power Plan).
- [x] Customize the name and description of the power plan.
- [x] Apply the power plan using its GUID.

## Final render 
image

### DIFF
--- a/functions/public/Invoke-WPFUltimatePerformance.ps1
+++ b/functions/public/Invoke-WPFUltimatePerformance.ps1
@@ -2,43 +2,65 @@ Function Invoke-WPFUltimatePerformance {
     <#
 
     .SYNOPSIS
-        Creates or removes the Ultimate Performance power scheme
+        Enables or disables the Ultimate Performance power scheme based on its GUID.
 
     .PARAMETER State
-        Indicates whether to enable or disable the Ultimate Performance power scheme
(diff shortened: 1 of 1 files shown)


YusufKhalifadev:
this won't be accepted cause of the french text in here 
```powershell
foreach ($line in $duplicateOutput) {
                if ($line -match "GUID du mode de gestion de l'alimentation\s*:\s*([a-fA-F0-9\-]+)") {
                    $guid = $matches[1]
                    Write-Output "GUID: $guid has been extracted and stored in the variable."
                    break
                }
```

AtisFPS:
Ups my bad i will changed that

AtisFPS:
I have patch that part thanks for that :D

ChrisTitusTech (approved):
Very cool idea on how to search and find the power plan across languages. I'll try this out.

PR #2576, opened 2024-08-18
Roll back WicroWin-Helper

mews-se:
As you mentioned on the latest stream. Let's keep the MicroWin version as "normal" but without bloat. Xbox and Gaming is bloat - if gamers want to install them. they can...


ChrisTitusTech:
Xbox Identity and "Gaming" is integrated into various things like game pass and Microsoft sign-ins. Taking these out isn't advised for a global stock image.

PR #2575, opened 2024-08-18
[MicroWin] Add custom user settings and other changes (docs/bugfixes)

CodingWonders:
# Pull Request

## Title
(Same as title)

## Type of Change
- [X] New feature
- [X] Bug fix
- [X] Documentation update
- [X] UI/UX improvement

## Description
Some people on the Discord community had suggested the ability to configure a custom user, and that's what this PR is (mainly) about.

Result with empty values:

powershell_XLDCCdeMZj
notepad_GAvmxRMUoG

Result with custom values:

1SqchB9Sld
notepad_TMrmibkxmy

It also adds checks for expedited app removal to prevent it from throwing REG errors

## Testing
Testing concluded with custom users being reflected in the target unattended answer file

## Impact
Custom user accounts

## Issue related to PR
None (suggestion on Discord)

## Additional Information
None, apart from slight changes in documentation

## Checklist
- [X] My code adheres to the coding and style guidelines of the project.
- [X] I have performed a self-review of my own code.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] I have made corresponding changes to the documentation.
- [X] My changes generate no errors/warnings/merge conflicts.

PR #2581, opened 2024-08-19
Microwin ISO downloader

MyDrift-user:
## Type of Change
- [x] New feature

## Description
- trying to use Fido to get the Windows ISO.
- Available are Windows 11 ISOs from 21H2 to 13H2
- English and English (International) will always be available. In addition it will take the language of the system and add this to the selection if available for ease of use. If the user want's a very specific language he may download it via the link provided in the utility.
- Currently Fido saves the iso into the location where u run the script. I asked the dev if he would like to make an additional parameter to set an output dir and he rejected the Idea due to the project mainly being for Rufus.

## Testing
- Autoselects newest Windows Release, takes language & arch from current system but can be modified.
- Still gives users the chance to use their own ISOs
{F124AB61-4305-4B6F-975B-2804B9A6E9C2}
{DC5C917F-9CE8-4C12-87B7-E553F75881F0}

- Downloads fine and instantly proceeds with mounting it after download has finished:
- Fido also gives a progress bar for the download:
image
image

- rewritten instruction:
{94CDED47-7584-4892-98DF-E77FC26E08AE}

## Checklist


CodingWonders:
@MyDrift-user, please update your branch since #2575 got merged

CodingWonders:
Well then, it downloaded the Spanish version (and the wrong one at that) even though I chose the English version. I'll investigate it:

powershell_0s8TxXOwlp

And, after downloading, it can't find it for some reason. I'll investigate that as well.

By the way, you should remove the architecture drop-down as Windows 11 is only 64-bit, unless you want to add Windows 10 support

MyDrift-user:
I removed the mapping of the UI language definition to the iso downloader bc for mine (en-CH) it does not work, so it will currently get the one from the current system.
As for the ISO, it will be saved inside the the directory that u are running it from, so normally in the winutil folder but with this I tried to set the location to the current users temp folder and set it back afterwards.

CodingWonders:
> I removed the mapping of the UI language definition to the iso downloader bc for mine (en-CH) it does not work (meaning we can't use `(Get-Culture).Name)`, so it will currently get the one from the current system.

We could use a function that returns an appropriate language string based on the CultureInfo for the selected item. Something like this for culture detection:

powershell_CYQyY9VYLc

.NET power combined with PowerShell!

> [!NOTE]
> `en-US` will be the selected item in the language ComboBox

> As for the ISO, it will be saved inside the the directory that u are running it from, so normally in the winutil folder but with this I tried to set the location to the current users temp folder and set it back afterwards.

Absolute paths will work better in my opinion, in case the user doesn't want to save the downloaded ISO to the current working directory.

> Windows 11 is also x86 (I think)

Not true. While Windows 11 still supports 32-bit applications (and let's be honest, Microsoft will support them for years to come), it's not available on 32-bit systems. The final version for such architecture is Windows 10

MyDrift-user:
> > As for the ISO, it will be saved inside the the directory that u are running it from, so normally in the winutil folder but with this I tried to set the location to the current users temp folder and set it back afterwards.
> 
> Absolute paths will work better in my opinion, in case the user doesn't want to save the downloaded ISO to the current working directory.

The Issue is tha Fido works that way, it does it like that. I've made an Issue on it's repo asking if he is okay with adding an output path parameter to the script but he rejected it due to the script primarily being for Rufus. I am personally not able to understand how he fetches the final link entirely, so I am not able to work on an inhouse solution using this method.
I have already mentioned what else we could do for an inhouse solution in the PR description, but that would kind of be a strange thing for ux.

CodingWonders:
> The Issue is tha Fido works that way, it does it like that. I've made an Issue on it's repo asking if he is okay with adding an output path parameter to the script but he rejected it due to the script primarily being for Rufus. I am personally not able to understand how he fetches the final link entirely, so I won't work on an inhouse solution using this method as of now. I have already mentioned what else we could do for an inhouse solution in the PR description, but that would kind of be a strange thing for ux.

We may not have to modify anything about Fido. Let it download to `$env:TEMP` and then we move it to the file path.

Right now I can't work on this because I'm adding a WIP parser for culture codes so that we can make dealing with languages in Fido easier. I'll comment back with more details regarding that

CodingWonders:
Well, the language parser works:

powershell_6Pz0Q8KD7Z

Code:

```powershell
            & $fidopath -Win 'Windows 11' -Rel $sync["ISORelease"].SelectedItem -Arch $rel -Lang "$(Get-FidoLangFromCulture -langName `"$($sync.ISOLanguage.SelectedItem)`")"
    ...

function Get-FidoLangFromCulture {

    param (
        [Parameter(Mandatory, Position = 0)] [string] $langName
    )

    switch -Wildcard ($langName)
    {
        "ar*" { return "Arabic" }
        "pt-BR" { return "Brazilian Portuguese" }
        "bg*" { return "Bulgarian" }
        {($_ -eq "zh-CH") -or ($_ -like "zh-Hans*") -or ($_ -eq "zh-SG") -or ($_ -eq "zh-CHS")} { return "Chinese (Simplified)" }
        {($_ -eq "zh") -or ($_ -eq "zh-Hant") -or ($_ -eq "zh-HK") -or ($_ -eq "zh-MO") -or ($_ -eq "zh-TW") -or ($_ -eq "zh-CHT")} { return "Chinese (Traditional)" }
        "hr*" { return "Croatian" }
        "cs*" { return "Czech" }
        "da*" { return "Danish" }
        "nl*" { return "Dutch" }
        "en-US" { return "English" }
        {($_ -like "en*") -and ($_ -ne "en-US")} { return "English International" }
        "et*" { return "Estonian" }
        "fi*" { return "Finnish" }
        {($_ -like "fr*") -and ($_ -ne "fr-CA")} { return "French" }
        "fr-CA" { return "French Canadian" }
        "de*" { return "German" }
        "el*" { return "Greek" }
        "he*" { return "Hebrew" }
        "hu*" { return "Hungarian" }
        "it*" { return "Italian" }
        "ja*" { return "Japanese" }
        "ko*" { return "Korean" }
        "lv*" { return "Latvian" }
        "lt*" { return "Lituanian" }
        "nb*" { return "Norwegian" }
        "pl*" { return "Polish" }
        {($_ -like "pt*") -and ($_ -ne "pt-BR")} { return "Portuguese" }
        "ro*" { return "Romanian" }
        "ru*" { return "Russian" }
        "sr-Latn*" { return "Serbian Latin" }
        "sk*" { return "Slovak" }
        "sl*" { return "Slovenian" }
        {($_ -like "es*") -and ($_ -ne "es-MX")} { return "Spanish" }
        "es-MX" { return "Spanish (Mexico)" }
        "sv*" { return "Swedish" }
        "th*" { return "Thai" }
        "tr*" { return "Turkish" }
        "uk*" { return "Ukrainian" }
        default { return "English" }
    }
}

```

Now, you need to be a bit prudent with Microsoft's service, as it can stop letting you (*IDK if temporarily or permanently. I hope it's the former*), so I had to use my phone as a wireless hotspot:

image

functions/public/Invoke-WPFGetIso.ps1
```diff
@@ -84,11 +84,26 @@ function Invoke-WPFGetIso {
 
         Invoke-WebRequest "https://github.com/pbatard/Fido/raw/master/Fido.ps1" -OutFile $fidopath
 
-        # TODO: Have to correct the language to actually be a working one
+        # TODO: on get iso, script gets downloaded and function is taken & called to get & display languages. after selection user can press proceed, which will start the fido process and select the output iso for further processing
         Set-Location -Path $env:temp
-        & $fidopath -Win 'Windows 11' -Rel $sync["ISORelease"].SelectedItem -Arch "x64" -Lang "$(Get-FidoLangFromCulture -langName `"$($sync.ISOLanguage.SelectedItem)`")"
+        & $fidopath -Win 'Windows 11' -Rel $sync["ISORelease"].SelectedItem -Arch "x64" -Lang $sync["ISOLanguage"].SelectedItem
```
CodingWonders, line 89: @MyDrift-user, I don't think Fido works like that. I don't think you can pass en-US to it to download the English version

MyDrift-user: @CodingWonders yeah this is a WIP push bc I work on multiple clients without shared storage...

I want to remove the Locale selection (currently) bc it is not needed. I am working on making the dropdown select languages, which then are passed to fido. This is a neccessary step and therefore more relevant.

CodingWonders (approved):
This is good for an initial implementation

CodingWonders:
**BIG NOTE TO ALL CONTRIBUTORS INTERESTED:** If you use Microsoft's service too many times a day with Fido, you will be banned from the software download service (both the website and any ISO download tool that uses its underpinnings) with the following error message for (at least) a day:

powershell_VgY55b1sJ5

You can get past this by connecting your system to a different device (it could be another router or a mobile hotspot), but it's best to wait for a day, or *just not use Fido too many times*.

Make sure that users know about this

MyDrift-user:
Another way of downloading the ISO without having this Issue would be auto downloading & managing the Media Creation tool. AveYo has a pretty good way of doing this that we could adapt as an in house solution.

CodingWonders:
@MyDrift-user, that seems like a good idea but, for now, we could use Fido for the introduction. We could later do a whole revamp that uses AveYo's solution.

I think that depends on the reliability of this method. Since end-users will be downloading ISOs once (maybe twice), this could work fine. However, it could present those issues if you make that script go further.

What do you think?

MyDrift-user:
Yeah for the testing I have gotten the "Cooldown" many times, but after 1 to 1.5 days it always worked again. For now this is sufficient. 
I tested German, Italian, English and English (International), all of them worked.

MyDrift-user:
Should I make the ISO download the second option? people would not always want to download the ISO bc they already have one or for other reasons?

CodingWonders:
@MyDrift-user, yes. You should make it a secondary option. Sorry for not answering sooner - I was testing MicroWin on Windows 10

MyDrift-user:
Done, updated images in description aswell

CodingWonders:
Already noticed

PR #2596, opened 2024-08-20
Choco Logic Refactor and Add Checkbox to prefer Chocolatey over Winget

Marterich:
## Type of Change
- [x] New feature
- [x] Refactor
- [x] Documentation update

## Description
In the past I've encountered the question "Why is Winget forced?" or "Why don't we have the option to use Chocolatey instead because winget is too big/slow/broken"

This PR tries to implement a Checkbox Option to Choose if Chocolatey should be preferred when installing/updating/uninstalling apps

image

Additional Info: 
- Make the Checkbox persistent  (a user who likes choco and sets this property, will likely always prefer it even if Winutil is restarted)
- Searches for .install and .portable apps automatically in addition to the main package and uninstalls all at once without manual intervention
- Opens Choco in a different window so the log doesn't get completely unnecessarily bloated but still keeps the ability to interact on unexpected errors
- Get-Installed apps now works with choco as well (only works for packages installed via Choco)
- Loading Bar works like for winget
- **Complete Refactor/Rewrite of the Choco logic**

## Testing
"It works on my machine" xD
Please do your own testing as I changed so much that it is easy to have missed something... 

## Checklist


ruxunderscore:
Maybe, instead, we do a "Prefer Chocolatey" toggle instead, so that it's clear that Chocolatey will be handled first, unless there isn't a package in the choco repo. That was you can have the best of both worlds when it comes to Auto and Chocolatey only.

Marterich:
> Maybe, instead, we do a "Prefer Chocolatey" toggle instead, so that it's clear that Chocolatey will be handled first, unless there isn't a package in the choco repo. That was you can have the best of both worlds when it comes to Auto and Chocolatey only.

Interesting idea, will look into that tomorrow 
:) 

As a nice side factor this could simplify the implementation quite a bit

ChrisTitusTech (approved):
Looks awesome. I'm going to save this one for the Livestream as I want to checkout the branch and push it through some tests.

MyDrift-user (approved):

docs/dev/tweaks/Essential-Tweaks/Powershell7.md
Real-MullaC, line 1: Dont change this part of the file.

Real-MullaC (changes requested):

PR #2600, opened 2024-08-21
Added Git commit hash to the about page

psyirius:
## Type of Change
- [x] New feature

## Impact
Indicates relatively exact where the build is coming from.
For example: Can used to differentiate builds across different branches, build from the same date and vice versa.
Just a little touch from the world of modern SW Development standards :P

## Additional Information
If built with unsaved changes, will show `(dirty)`

## Screenshots
image
image

## Checklist


Real-MullaC:
I am not coming out offensive but this just bloats WinUtil up. With un nessicary information.

psyirius:
I don't think its unnecessary. It helps differentiating builds deployed on the same day.

CodingWonders:
While I think that this is interesting, the releases will be coming from a GitHub action and, if people look up the commit hash, they're not going to find anything interesting (apart from the results of an automated system)

This is only useful for developers IMO

ChrisTitusTech:
Most of devs will probably be doing a Compile -Run locally and should be able to reference it. 

Doing a more unique build version would probably be better, but I don't want to change from the current date format. Its easy to understand from a user perspective and this commit would confuse people. I get emails every day about all kinds of things from the utility and this might add to that pile.

PR #2606, opened 2024-08-22
 Added AyuGram Desktop

its-ashu-otf:
# Pull Request

## Type of Change
-  App Addition

## Description
Compiled the winutil.ps1 script and Checked The App Installation works well.


Real-MullaC:
This PR will be declined as apps are not being added currently.

its-ashu-otf:
> This PR will be declined as apps are not being added currently.

Ohh So should i  close this?

PR #2616, opened 2024-08-24
Explorer Refresh instead of restart

MyDrift-user:
# Pull Request

## Type of Change
- [x] Refactoring

## Description
Instead of restarting explorer, we could "refresh" it. For example when changing settings in the settings app it does not restart the explorer process, but still works fine. I am trying to recreate that.

## Testing
- Changes currently work with dark mode toggle and other light stuff, but not with snap window toggles for example.
- I need to test it out more and take a deeper look at how Windows does it.

## Impact
restarting explorer is kind of a trash experience, it would help the ux.

## Checklist
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -19,10 +19,6 @@ template: |
 
   $CHANGES
 
-  ## Contributors
-
-  $CONTRIBUTORS
-
 change-title-escapes: '\<*_&"'''
 autolabeler:
   - label: 'documentation'


functions/public/Invoke-WPFToggle.ps1
```diff
@@ -34,4 +34,6 @@ function Invoke-WPFToggle {
         "WPFToggleTaskbarAlignment" {Invoke-WinUtilTaskbarAlignment $(Get-WinUtilToggleStatus WPFToggleTaskbarAlignment)}
         "WPFToggleDetailedBSoD" {Invoke-WinUtilDetailedBSoD $(Get-WinUtilToggleStatus WPFToggleDetailedBSoD)}
     }
+
+    Invoke-WinUtilExplorerRefresh
```
og-mrk, line 38: ... isn't calling `Invoke-WinUtilExplorerRefresh` on every toggle interaction a bit over kill maybe? 😅
I think it's better to call it when necessary, and it'll show a clear dependency on this ExplorerRefresh function, which'll help with the `devdocs-generator` script.

PR #2683, opened 2024-09-05
Fix/Refactor windev Admin Elevation

### LINKED ISSUE #2668: Broken Run as admin (windev)
## Describe the bug
Broken Run as admin (windev)

## To Reproduce
1. Open powershell un-elevated
2. `irm "https://christitus.com/windev" | iex`

## Expected behavior
Obviously should be successful elevating to admin

## Screenshots
image

Marterich:
## Type of Change
- [x] Bug fix
- [x] Refactor

## Description
Changed the auto elevation logic to use the same URLs whether windev is called as admin or not.
(previously a static URL was used when the shell wasn't run as admin) 

## Issue related to PR
- Resolves #2668


MyDrift-user:
@Marterich it was done because otherwise when launchin the winutil dev release it will elevate and run the latest full release instead of the originally launched dev release. There might be a better fix but that was mine at the time..

Marterich:
> @Marterich it was done because otherwise when launchin the winutil dev release it will elevate and run the latest full release instead of the originally launched dev release. There might be a better fix but that was mine at the time..

Allright, I'll take another look

MyDrift-user:
If you only want to fix the attached issue, you may remove/fix the $arglist bc the rest should work fine

Marterich:
> If you only want to fix the attached issue, you may remove/fix the $arglist bc the rest should work fine

I updated the logic and the PR Description. Now the same URLs are being used, whether /windev is called as a privileged user or not.

MyDrift-user (approved):
looks good to me, thanks :)

MyDrift-user:
wait, shouldn't you escape `Invoke-RestMethod` under $script = `Invoke-RestMethod $url`?

Marterich:
> wait, shouldn't you escape `Invoke-RestMethod` under $script = `Invoke-RestMethod $url`?

This is not really necessary because Powershell is not as strict as bash for example, meaning, if Invoke-RestMethod is a known module/function it gets directly invoked. Winutil gets downloaded and stored in $script and later passed as a string to iex

ChrisTitusTech (approved):

ChrisTitusTech:
Fantastic way of elevating it properly. Thanks @Marterich 

I haven't forgotten about the Choco tweak, I just wanted to show that on stream when I merge it.

ChrisTitusTech (approved):

PR #2720, opened 2024-09-11
Fix the way arguments are passed to the runspace

Marterich:
## Type of Change
- [x] Hotfix

## Description
This is my take on fixing the Runspace problem where the arguments would be interpreted as a string instead of a list and couldn't be distinguished between the parameters of the function.
I added a new -ParameterList option to provide the parameter Name as well as the content and reverted the -ArgumentList to the old logic for backwards compatibility

## Testing
I tested installs/uninstalls of programs and tweaks but not much else.

### DIFF
--- a/functions/public/Invoke-WPFInstall.ps1
+++ b/functions/public/Invoke-WPFInstall.ps1
@@ -20,7 +20,7 @@ function Invoke-WPFInstall {
         return
     }
     $ChocoPreference = $($sync.WPFpreferChocolatey.IsChecked)
-    Invoke-WPFRunspace -ArgumentList $PackagesToInstall,$ChocoPreference -DebugPreference $DebugPreference -ScriptBlock {
+    $installHandle = Invoke-WPFRunspace -ParameterList @(("PackagesToInstall", $PackagesToInstall),("ChocoPreference", $ChocoPreference)) -DebugPreference $DebugPreference -ScriptBlock {
         param($PackagesToInstall, $ChocoPreference, $DebugPreference)
         if ($PackagesToInstall.count -eq 1) {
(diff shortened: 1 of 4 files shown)


functions/private/Install-WinUtilChoco.ps1
```diff
@@ -13,10 +13,11 @@ function Install-WinUtilChoco {
...
             return
         }
-
+        # Install logic taken from https://chocolatey.org/install#individual
         Write-Host "Seems Chocolatey is not installed, installing now."
-        Start-Process -FilePath "powershell" -ArgumentList "Set-ExecutionPolicy Bypass -Scope Process -Force; Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1')) -ErrorAction Stop" -Wait -NoNewWindow
-        Start-Process -FilePath "powershell" -ArgumentList "choco feature enable -n allowGlobalConfirmation" -Wait -NoNewWindow
+        Set-ExecutionPolicy Bypass -Scope Process -Force; 
+        [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; 
+        Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://community.chocolatey.org/install.ps1'))
```
ChrisTitusTech, line 20: Why did you change the method of choco install?

ChrisTitusTech (approved):

functions/private/Install-WinUtilChoco.ps1
Marterich: The other method seemed to be broken (see discord messages)

PR #2721, opened 2024-09-11
Minimalistic Hoftix for Tweak Problem

Marterich:
## Type of Change
- [x] Hotfix

## Description
Another minimalistic way of fixing the Tweaks problem


Marterich:
As #2720 was merged already, I'm closing this as it was simply an alternative solution to fix the same problem and illustrate what went wrong in the first place

PR #2788, opened 2024-09-21
Fix Dark Color Scheme

Marterich:
## Type of Change
- [x] UI/UX improvement

## Description

Most likely some of the Changes to the Theme were broken during some of the last commits, so here's my two cents on how to fix it (without actually taking the time and checking what settings were used before)

| Before    | After |
| -------- | ------- |
| image  | image    ||


MyDrift-user:
I feel like tha background could be a bit darker than this.

Marterich:
> I feel like tha background could be a bit darker than this.

It's actually the same as it is in the current stable release

MyDrift-user:
Yeah ik I meant that, maybe not as dark as current main branch but a bit darker

Marterich:
Let's leave the final design to our best maintainer :D

SHR1SHAK:
Oh this broke again.

SHR1SHAK:
@Marterich for color reference you can look youtube dark mode.

SHR1SHAK:
DNS Button color also broke

MyDrift-user:
It is the "Comboboxforegroundcolor" is used in many places, which is ducked up

Marterich:
> It is the "Comboboxforegroundcolor" is used in many places, which is ducked up

Rly? Haven't seen it anywhere else while taking a quick peek

SHR1SHAK:
i think we need to document what goes where for future ease.

ChrisTitusTech (approved):

PR #2789, opened 2024-09-21
Import Bugfix: Change Preset Type and contains check

### LINKED ISSUE #2772: Tweaks export/import bug.
## Describe the bug
If you select the standard tweaks then export it & import it, everything is same except `Change Windows Terminal default` is automatically selected.

## Expected behavior
Exported config should be the same as per user choice.

## Screenshots
Export:
B
Import:
A

Marterich:
## Type of Change

- [x] Bug fix

## Description

When the Powershell Tweaks Import didn't work reliably because until now, all of the imported tweaks were stored as a long string and searched for a substring. This resulted in both powershell tweaks being selected because the Internal tweak names are quite similar. 
Now, all the tweaks are loaded as an array, and checked for an exact match to fix the issue

## Testing

Created an Export of Applications / Tweaks and Features, cleared all checkmarks, imported them and checked if they are the same

## Issue related to PR

- Resolves #2772

### DIFF
--- a/functions/public/Invoke-WPFImpex.ps1
+++ b/functions/public/Invoke-WPFImpex.ps1
@@ -56,8 +56,7 @@ function Invoke-WPFImpex {
                 }
             }
         }
-
-        $flattenedJson = [string]$flattenedJson
+        
         Invoke-WPFPresets -preset $flattenedJson -imported $true
     }
 }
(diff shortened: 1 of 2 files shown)


SHR1SHAK:
will the exported .json still work after this update?

Marterich:
> will the exported .json still work after this update?

Yes, the export function didn't change in any way

SHR1SHAK:
> > will the exported .json still work after this update?
> 
> Yes, the export function didn't change in any way

nice

ChrisTitusTech (approved):

PR #2816, opened 2024-09-24
Use here-strings for JSON in Compile.ps1

ruxunderscore:
**NOTE: I recommend going through my commits to see the changes I made over time, to better understand why I coded things the way I did.**

Change Compile.ps1 to use here-strings with JSON.

## Type of Change
- [x] Refactoring

## Description
This modifies the Compile.ps1 script's handling of JSON to use here-strings and `-Raw`, similar to what we do with Xaml. This allows us to forego the replace when we `ConvertTo-Json -Depth 3`.

There are a couple formatting to CamelCasing changes for consistency, as well. 

## Testing
Tested the Compile in pwsh and powershell (pulled in changes from @og-mrk for Pre-processing for testing in PS5.)

## Checklist

### DIFF
--- a/Compile.ps1
+++ b/Compile.ps1
@@ -85,6 +85,7 @@ Get-ChildItem "config" | Where-Object {$psitem.extension -eq ".json"} | ForEach-
         }
     }
 
+    # Line 90 requires no whitespace inside the here-strings, to keep formatting of the JSON in the final script.
     $json = @"
 $($jsonAsObject | ConvertTo-Json -Depth 3)
 "@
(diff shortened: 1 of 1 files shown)


ruxunderscore:
@ChrisTitusTech I had to refactor my code where I used here-strings for the Xaml stuff because of the changes to Compile.ps1 and Preprocessing around the same time, which hasn't been causing issues, and hasn't been rolled back. JSON was a little more work to do.
The rollback you are thinking of, was related to `-Force` being used in Preprocessing when dealing with ignoring files/directories, because it was causing compile issues on Powershell 5. I believe OG rewrote all that algorithm to fix the issue and restore functionality. Unrelated to JSON and Xaml using here-strings.

PR #2881, opened 2024-10-06
Fix typo in Remove-Packages

leonardo-dgs:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Fix typo in Remove-Packages function: `"*indows-Client-LanguagePack*"` instead of `"*Windows-Client-LanguagePack*"`

### DIFF
--- a/functions/private/Invoke-WinUtilMicroWin-Helper.ps1
+++ b/functions/private/Invoke-WinUtilMicroWin-Helper.ps1
@@ -121,7 +121,7 @@ function Remove-Packages {
 
         $pkglist = $pkglist | Where-Object {
                 $_ -NotLike "*ApplicationModel*" -AND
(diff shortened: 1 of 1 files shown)


CodingWonders:
@leonardo-dgs, it has wildcard characters. It will always spot `Windows-Client-LanguagePack` without this change

CodingWonders:
It's not a big deal though. It looks nice and, if it still works, it can be merged. Have you made a test image with your change and verified that it works correctly, @leonardo-dgs?

MyDrift-user:
Powershell string comparison operators are usually not case sensitive, maybe that was the thought behind starting with *i.

leonardo-dgs:
I've tested, and it works

PR #3032, opened 2024-11-10
Fix: Correct Windows Terminal default profile path to PowerShell 7

Anmol-sudo:
## Type of Change
- [x] Bug fix

## Description
This pull request addresses an issue with the Windows Terminal configuration in the Invoke-WPFTweakPS7 function. Previously, the function was unable to correctly locate or interact with the PowerShell 7 installation due to an incorrect path. 

The changes include:
1. Updated the path in Invoke-WPFTweakPS7.ps1 to correctly locate the PowerShell 7 installation.
2. Improved the logic for setting the default profile in Windows Terminal settings.
3. Enhanced error handling and user feedback for cases where Windows Terminal or its settings file are not found.

These modifications ensure proper functionality of PowerShell 7 related tweaks and improve the reliability of PowerShell 7 operations in the winutil script.

## Testing
I have tested these changes by:
1. Running the script on a clean Windows installation.
2. Verifying that PowerShell 7 is correctly set as the default profile in Windows Terminal after running the tweak.
3. Ensuring that the script gracefully handles scenarios where Windows Terminal is not installed or the settings file is not found.
4. Checking that the script can still set PowerShell 5 as the default when requested.

## Impact
This fix will improve the user experience for those using the PowerShell 7 tweak, ensuring that their Windows Terminal is correctly configured. It also adds robustness to the script by improving error handling.

## Additional Information
This change does not introduce any new dependencies and should be backwards compatible with existing setups.

## Checklist

image

error-img

### DIFF
--- a/functions/public/Invoke-WPFTweakPS7.ps1
+++ b/functions/public/Invoke-WPFTweakPS7.ps1
@@ -1,5 +1,5 @@
-function Invoke-WPFTweakPS7{
-        <#
+function Invoke-WPFTweakPS7 {
+    <#
     .SYNOPSIS
         This will edit the config file of the Windows Terminal Replacing the Powershell 5 to Powershell 7 and install Powershell 7 if necessary
     .PARAMETER action
@@ -12,7 +12,7 @@ function Invoke-WPFTweakPS7{
     )
 
     switch ($action) {
-        "PS7"{
+        "PS7" {
             if (Test-Path -Path "$env:ProgramFiles\PowerShell\7") {
                 Write-Host "Powershell 7 is already installed."
             } else {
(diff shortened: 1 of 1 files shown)


Cryostrixx:
Hello, @Anmol-sudo! Thanks for the report, but just to give you a heads-up both PowerShell 5 and PowerShell 7 are on the path by default, it is not necessary to hardcode the paths as their respective entries are added when installed (see attached screenshot):
image

Anmol-sudo:
image
In Windows Terminal winutil sets the path like this for powershell7
image
which is wrong, the correct should be without strings.
image
So i fixed it

Cryostrixx:
> image In Windows Terminal winutil sets the path like this for powershell7 image which is wrong, the correct should be without strings. image So i fixed it

I have a different issue with it, it sets the path correctly, but it doesn't show up in the list of profiles I can select, and not even Ctrl+Shift+3 works to open it, Regardless of that, if the fix you proposed works as intended otherwise then I see no reason it shouldn't be added. Just be patient, and eventually, someone will approve it.

Anmol-sudo:
Thanks, I think the ctrl+shift+3 is the Windows terminal problem not the winutil problem, u can reset the terminal tho

Cryostrixx:
> Thanks, I think the ctrl+shift+3 is the Windows terminal problem not the winutil problem, u can reset the terminal tho

It's more on me than anything else, I won't report it as it's caused by user error, I am fixing that now, this PR should remain focused on the initial topic.

ChrisTitusTech (changes requested):
The path is incorrect this should use $env:HOMEDRIVE instead of C: 

This ensures it will work on non-standard installs where C: isn't the home drive.

PR #3268, opened 2025-03-14
Remove Balena Etcher due to privacy concerns

Marterich:
## Type of Change
- [x] Security patch

## Description

In the recent past, the privacy focused Linux Distro TAILS removed the recommendation to use Balena Etcher to create there USB Keys due to privacy concerns. 
An [investigation](https://gitlab.tails.boum.org/tails/tails/-/issues/16381) on their part resulted in at least the following data being shared with the balena company
* The IP of the user
* The filename of the image, which lets Balena knows that the user is installing Tails
* The brand and model of the USB stick

https://tails.net/news/rufus/index.en.html
https://cybernews.com/privacy/tails-raises-privacy-concerns-over-balenaetcher-recommends-rufus/

## Issue related to PR

- Resolves #3260

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -527,14 +527,6 @@
         "link": "https://espanso.org/",
         "winget": "Espanso.Espanso"
     },
-    "etcher": {
-        "category": "Utilities",
-        "choco": "etcher",
-        "content": "Etcher USB Creator",
-        "description": "Etcher is a powerful tool for creating bootable USB drives with ease.",
-        "link": "https://www.balena.io/etcher/",
-        "winget": "Balena.Etcher"
-    },
     "falkon": {
         "category": "Browsers",
         "choco": "falkon",


Marterich:
@og-mrk maybe also add balena to your pinned application list comment. Detailing the reasoning

Kingproone:
I see a bit of a double standard here, because if this app gets removed for privacy violations, google chrome should be nuked as a first follow up.

Marterich:
In the end, I think you should be thinking about what you want winutil to be (and probably note down the guidelines somewhere)

A few ideas to differentiate could be 

- a curated list of best in class (mostly) Open source software you would recommend others to use
- a list of general programs that the community asks for
- a curated list of recommended apps that follow the _winutil spirit_ of removing telemetry whatever possible 

All in all, personally I would gravitate towards the first option. (Of course keeping in mind, that you would most likely have to remove quite a few apps from the current list)

ruxunderscore:
I wonder how difficult it would be for us to add an option to hide apps that don't follow the WinUtil spirit as you mention.

ChrisTitusTech (approved):

PR #3277, opened 2025-03-19
Optimize UI Performance

Marterich:
## Type of Change
- [x] Refactoring
- [x] Hotfix

## Description

The new UI slowed the start time down quite a bit because it was poorly optimized. 

This PR tries to mitigate at least some of this by refactoring the code and reorganizing the resource definitions so the dont have to be seperateley created for each.

## Testing

In my Testing on my kinda potato VM the Time Taken from Script Start to Responsive UI was reduced by about ~30-40% (~4s)


ruxunderscore (approved):

ruxunderscore:
I love it when you can reduce overhead by consolidating frequently used code into it's own thing. I just went through my profile.ps1 today to improve consistency.

ChrisTitusTech (approved):

PR #3310, opened 2025-04-07
 Optimize Preprocessor by Implementing File Hashing for Modified Files Detection

Marterich:
## Type of Change
- [x] New feature

## Background
I noticed that I frequently disabled the preprocessor using the -SkipPreprocessing flag because processing all 200+ files on my VM consistently took several seconds. This effectively negated the purpose of having a preprocessor in the first place.

To address this issue, this PR removes the option to disable the preprocessor entirely. Instead, it introduces a feature that runs the preprocessor once for all files, stores their hashes in a temporary file, and only processes files that have been modified in subsequent runs. This enhancement aims to improve efficiency and streamline the preprocessing workflow.

## Changes Made

- Removed the -SkipPreprocessing flag option.
- Removed ThrowExceptionOnEmptyFileList flag option and replaced it with a Write-Debug message
- Implemented file hashing to track modifications.
- Added logic to store file hashes in a temporary file.
- Updated the preprocessing routine to only process modified files on consecutive runs.
- This change should significantly reduce processing time and enhance the overall functionality of the preprocessor.

### DIFF
--- a/tools/Invoke-Preprocessing.ps1
+++ b/tools/Invoke-Preprocessing.ps1
@@ -3,9 +3,6 @@ function Invoke-Preprocessing {
         .SYNOPSIS
         A function that does Code Formatting using RegEx, useful when trying to force specific coding standard(s) to a project.
 
-        .PARAMETER ThrowExceptionOnEmptyFilesList
-        A switch which'll throw an exception upon not finding any files inside the provided 'WorkingDir'.
-
         .PARAMETER SkipExcludedFilesValidation
         A switch to stop file path validation on 'ExcludedFiles' list.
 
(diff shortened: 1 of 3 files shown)


ChrisTitusTech (approved):

tools/Invoke-Preprocessing.ps1
```diff
@@ -36,35 +33,24 @@ function Invoke-Preprocessing {
...
-        [Parameter(position=0)]
-        [switch]$SkipExcludedFilesValidation,
-
-        [Parameter(position=1)]
-        [switch]$ThrowExceptionOnEmptyFilesList,
-
-        [Parameter(Mandatory, position=2)]
+        [Parameter(Mandatory, position=1)]
         [ValidateScript({[System.IO.Path]::IsPathRooted($_)})]
         [string]$WorkingDir,
```
og-mrk, line 45:
```suggestion
        [Parameter(position=0)]
        [switch]$SkipExcludedFilesValidation,

        [Parameter(Mandatory, position=1)]
        [ValidateScript({[System.IO.Path]::IsPathRooted($_)})]
        [string]$WorkingDir,
```
Looking at the diff, it seems like you've removed the `-SkipExcludedFilesValidation` switch, which's still being used in the script (see [this line](https://github.com/ChrisTitusTech/winutil/pull/3310/files#diff-d103e6d7182657166ff8b68dde3e3a06df621d5bd694f5833675232560f6c926R79)). Recommend re-adding it.

Marterich: Good catch @og-mrk
The question now is, do we really need the switch (and the functionality)?
At the moment, the Parameter is never used as far as i can see.

Should we maybe just remove the logic alltogether?

tools/Invoke-Preprocessing.ps1
```diff
@@ -76,9 +58,8 @@ function Invoke-Preprocessing {
     ForEach ($excludedFile in $ExcludedFiles) {
         $InternalExcludedFiles.Add($excludedFile) | Out-Null
     }
-
-    # Validate the ExcludedItems List before continuing on,
-    # that's if there's a list in the first place, and '-SkipInternalExcludedFilesValidation' was not provided.
+
```
og-mrk, line 61:
```suggestion

```
Unnecessary Whitespace characters.

og-mrk (approved):
Works great! thanks for the improvements @Marterich 😄

ChrisTitusTech (approved):

PR #3328, opened 2025-04-15
Fix ConvertFrom-Json compiler error

Marterich:
## Type of Change
- [x] Hotfix

Fixes the JSON Format so it can be parsed by the compiler


Marterich:
Already done in #3326

PR #3366, opened 2025-05-17
Upgrade Code Signing Algorithm from SHA1 to SHA256

### LINKED ISSUE #3359: Version 25.05.05 is signed with Sha1 instead of Sha256
### ⚙️ Issue Checklist

- [x] I checked for an existing pull request that addresses this issue.

### 📜 What part of Winutil are you having issues with?

EV Signature

### 📝 Provide a clear and concise description of the issue.

I noticed that the latest release's code signature uses Sha1 instead of Sha256

I have a hunch you might be using [Digicert's tool](https://www.digicert.com/kb/code-signing/digicert-certificate-utility-to-sign-code.htm), because I noticed myself a while back that for some bizarre reason, it signs powershell scripts with the old Sha1 unlike everything else with Sha256. 

Here's the signtool command I've always used which will work too but explicitly sets the hash type:
```cmd
signtool.exe sign /td sha256 /tr http://timestamp.digicert.com /fd sha256 /n "Whatever Exact Entity Name, LLC" <filepath>
```

### 🔄 Steps to reproduce the issue.

N/A

### ❌ Paste the full error output (if available).

N/A

### 🖼️ Additional context.

_No response_

Marterich:
## Type of Change
- [x] Security patch

## Description

As pointed out by @ThioJoe in the linked issue, winutil currently still uses SHA1 for code signing during compilation.
This is due to Microsoft using SHA1 as the default algorithm until PowerShell 7.3, as noted in [Set-AuthenticodeSignature](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.security/set-authenticodesignature?view=powershell-7.5#-hashalgorithm). 

While finding a SHA1 hash collision is still challanging, it is becoming increasingly feasible, and this trend is likely to continue. Given that there are no performance reasons to retain SHA1, transitioning to a more secure algorithm, such as SHA256, will significantly reduce the risk of hash collisions and enhance overall security. 

## Issue related to PR

- Resolves #3359

## Additional Information

# !  PLEASE TEST CAREFULLY AS YOU ARE THE ONLY ONE WHO HAS THE CERTIFICATE TO TEST IF THIS WORKS AS EXPECTED

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -60,7 +60,7 @@ jobs:
         run: |
           $cert = Get-ChildItem -Path Cert:\CurrentUser\My -CodeSigningCert | Select-Object -First 1
           if ($null -eq $cert) { throw "Code signing certificate not found" }
-          Set-AuthenticodeSignature -FilePath ./winutil.ps1 -Certificate $cert -TimeStampServer "http://timestamp.digicert.com"
+          Set-AuthenticodeSignature -FilePath ./winutil.ps1 -HashAlgorithm SHA256 -Certificate $cert -TimeStampServer "http://timestamp.digicert.com"
 
       - name: Verify code signature
         shell: pwsh


ChrisTitusTech:
Can't use this type of Code Signing any longer. Moving to an EV cert means I have to use a physical 2FA for each release.

PR #3367, opened 2025-05-17
Add Right-Click Action on App Install Page

Marterich:
… and info options

## Type of Change
- [x] New feature
- [x] UI/UX improvement

## Description

This pull request adds a right-click context menu on the app install page, allowing users to quickly access essential actions for each app.

Right-Click Menu: Users can right-click on an app to reveal a context menu.
Action Buttons:
- Install: Initiate app installation.
- Uninstall: Remove the app directly.
- Info: Open the app's web page for more details.

image

## Additional Information
The context menu, which includes Install, Uninstall, and Info buttons, is created once and dynamically updated with the relevant app information and button positioning. This approach reduces the number of WPF elements from over 1000 to just 5, significantly improving loading times compared to the initial redesign


FatBastard0:
That looks cool I like that

ChrisTitusTech (approved):

PR #3376, opened 2025-05-21
Cleanup "Set Services to Manual" Tweak

Marterich:
## Type of Change
- [x] Refactoring

## Description

Removes all service definitions contained in the Tweak for which the Startup Type was never changed away from the original (default) value.
This improves readability and maintainability of the tweaks.json as well as the log and powershell output, as only the values that are actually modified are displayed.


MyDrift-user (approved):

ChrisTitusTech:
The main issue with this is there were quite a few additional services in there from various installs. The Original value wasn't always 100% correct and did change a bit more. I'll need to do a log on a new install to see exactly what services we are changing to verify. 
Also this changes from each version of windows 23H2 -> 24H2 -> 25H2 and I need to create a better workflow to maintain the services tweak, as I agree its a bit of a mess right now.

ChrisTitusTech (changes requested):
The main issue with this is there were quite a few additional services in there from various installs. The Original value wasn't always 100% correct and did change a bit more. I'll need to do a log on a new install to see exactly what services we are changing to verify.
Also this changes from each version of windows 23H2 -> 24H2 -> 25H2 and I need to create a better workflow to maintain the services tweak, as I agree its a bit of a mess right now.

MyDrift-user:
> TL;DR: Maintain what we have when it comes to storing the services we want to change, but analyze and compare the current services on the current system and their startup configuration. Then, only change what is on the system and isn't already set to Manual.
> I believe this could also remove the need to check if the list of services needs any removals as we go forward, aside from the unlikely scenario that setting a service to manual ends up causing problems.

Me and @og-mrk played around with this Idea some time ago. If I recall correctly there are some issues with this tweak that don't come up as errors in the console even with debug mode (prob. due to the runspace config?). Around 2 services could not be configured. the state would stay the same (probs because of missing priviledges?). Running this without a runspace just spammed errors and gave you a long error shell history.

I think I had similar issues when looking into the way the reg changes were made.

Also when trying to implement runspaces for MicroWin. Some cmdlets worked and some not, debugging was hard due to the errors (if there even were any) not being helpfull.

I think to tackle this issue we first have to either fix runspace console logs or improve the debug mode. That way we can fix the problems more directly.

Marterich:
I agree with many of the points raised in the previous comments, but I feel the conversation has drifted from the main purpose of this PR. The original problem I intended to solve was the unnecessary inclusion of services in a list where both the recommended and default settings are 'manual,' meaning they would never be altered in the first place.

MyDrift-user:
Agree aswell, hence my approval of this PR inicially.
This PR should be kept for what it was meant for and not explode in size 👍

zoicware:
ive somewhat tackled setting services to manual without just hardcoding the whole thing...
something to keep in mind is that any "user" services (typically going to contain user in the name) will only be modifiable using registry 
also, you should be able to only get services where the provider is Microsoft to avoid clobbering third party ones and avoid a most likely never-ending list of excluded services 
```PowerShell
$services = Get-Service
      $servicesKeep = @(
        'AudioEndpointBuilder',
        'Audiosrv',
        'EventLog',
        'SysMain',
        'Themes',
        'WSearch',
        'NVDisplay.ContainerLocalSystem',
        'WlanSvc',
        'BFE', 
        'BrokerInfrastructure', 
        'CoreMessagingRegistrar', 
        'Dnscache', 
        'LSM', 
        'mpssvc', 
        'RpcEptMapper', 
        'Schedule', 
        'SystemEventsBroker', 
        'StateRepository', 
        'TextInputManagementService', 
        'sppsvc'
      )
      foreach ($service in $services) { 
        if ($service.StartType -like '*Auto*') {
          if ($servicesKeep -notcontains $service.Name) {
            try {
              Set-Service -Name $service.Name -StartupType Manual -ErrorAction Stop
            }
            catch {
              $regPath = "HKLM:\SYSTEM\CurrentControlSet\Services\$($service.Name)"
              Set-ItemProperty -Path $regPath -Name 'Start' -Value 3 -ErrorAction SilentlyContinue
            }
              
          }         
        }
      }
```

GabiNun:
In the tweaks.json file, the services section takes up about 1300 lines, which is over 30% of the file,

Even though these are the only services it actually changes in Windows 11 Pro with administrator permissions.

{
  "service": [
    {
      "Name": "AssignedAccessManagerSvc",
      "StartupType": "Disabled",
      "OriginalType": "Manual"
    },
    {
      "Name": "BITS",
      "StartupType": "AutomaticDelayedStart",
      "OriginalType": "Automatic"
    },
    {
      "Name": "DiagTrack",
      "StartupType": "Disabled",
      "OriginalType": "Automatic"
    },
    {
      "Name": "MapsBroker",
      "StartupType": "AutomaticDelayedStart",
      "OriginalType": "Automatic"
    },
    {
      "Name": "MsKeyboardFilter",
      "StartupType": "Manual",
      "OriginalType": "Disabled"
    },
    {
      "Name": "PcaSvc",
      "StartupType": "Manual",
      "OriginalType": "Automatic"
    },
    {
      "Name": "WSearch",
      "StartupType": "AutomaticDelayedStart",
      "OriginalType": "Automatic"
    },
    {
      "Name": "edgeupdate",
      "StartupType": "Manual",
      "OriginalType": "Automatic"
    }
  ]
}

ChrisTitusTech:
Ended up using #3732 as it wasn't quite as aggressive and looks good in tests.

PR #3385, opened 2025-05-25
Add overlay and block app list during install/uninstall

Marterich:
## Type of Change
- [x] New feature
- [x] UI/UX improvement

## Description

Adds a overlay and a blur effect to the app list during installation or uninstallation of a program do indicate to the user that it's not possible to start another install/uninstall right now. 

Install_Overlay

### DIFF
--- a/functions/private/Initialize-InstallAppArea.ps1
+++ b/functions/private/Initialize-InstallAppArea.ps1
@@ -53,9 +53,6 @@
         $targetGrid.Children.Add($Border) | Out-Null
 
         $overlay = New-Object Windows.Controls.Border
-        # TODO: Implement a dynamic way to set the size of the overlay based on the size of the parent element
-        $overlay.Width = "300"
-        $overlay.Height = "300"
         $overlay.CornerRadius = New-Object Windows.CornerRadius(10)
         $overlay.SetResourceReference([Windows.Controls.Control]::BackgroundProperty, "AppInstallOverlayBackgroundColor")
         $overlay.Visibility = [Windows.Visibility]::Collapsed
--- a/functions/private/Show-WPFInstallAppBusy.ps1
+++ b/functions/private/Show-WPFInstallAppBusy.ps1
@@ -3,6 +3,7 @@ function Show-WPFInstallAppBusy {
     .SYNOPSIS
         Displays a busy overlay in the install app area of the WPF form.
         This is used to indicate that an install or uninstall is in progress.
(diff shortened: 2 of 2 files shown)


functions/private/Initialize-InstallAppArea.ps1
```diff
@@ -17,18 +19,27 @@
...
         $scrollViewer.HorizontalAlignment = 'Stretch'
         $scrollViewer.VerticalAlignment = 'Stretch'
         $scrollViewer.CanContentScroll = $true
+        $sync.InstallAppAreaScrollViewer = $scrollViewer
+        $Border.Child = $scrollViewer
+
+        # Initialize the Blur Effect for the ScrollViewer, which will be used to indicate that an install/uninstall is in progress
+        $blurEffect = New-Object Windows.Media.Effects.BlurEffect
+        $blurEffect.Radius = 0
+        $scrollViewer.Effect = $blurEffect
```
ruxunderscore, line 36: Instead of a blur, maybe a darken of the background, instead? Darkening it can be better contrast-wise and would be less of a striking change, in my opinion.

functions/private/Initialize-InstallAppArea.ps1
```diff
@@ -38,8 +49,67 @@
...
         $itemsControl.SetValue([Windows.Controls.VirtualizingStackPanel]::VirtualizationModeProperty, [Windows.Controls.VirtualizationMode]::Recycling)
 
-        $scrollViewer.Content = $itemsControl
-        $Border.Child = $scrollViewer
-        $null = $targetGrid.Children.Add($Border)
+        # Add the Border containing the App Area to the target Grid
+        $targetGrid.Children.Add($Border) | Out-Null
+
+        $overlay = New-Object Windows.Controls.Border
+        # TODO: Implement a dynamic way to set the size of the overlay based on the size of the parent element
```
ruxunderscore, line 56: I'm looking into this, and will let you know when I have something.

functions/private/Initialize-InstallAppArea.ps1
Marterich: Initially, I tried overlaying a grey, translucent layer, but the resulting color change of the majority of the app page felt a bit jarring. So, I opted for the frosted glass look, similar to what's seen in some Windows apps. I'm not sure if this is the best approach, but it felt like an unobtrusive way to indicate a blocked element.

PR #3581, opened 2025-09-06
Fix: retry password prompt on failure to prevent script halting

ZeusCraft10:
This PR enhances the password prompt logic during application installs (e.g., Spotify) triggered by the script. Previously, entering incorrect credentials caused the prompt to close and halted the install script, forcing users to restart the process.

Key improvements:

Users are now re-prompted for credentials if authentication fails, instead of the script halting.

The process continues until the correct password is entered or the user cancels.

Prevents unnecessary script exits and improves overall user experience when installing password-protected applications.

Tested:
Tested with both valid and invalid credentials in a CLI (text prompt) environment to confirm retry logic works as intended.


ZeusCraft10:
Thank you 👍👍

PR #3613, opened 2025-09-26
Add Disable Edge Tweak

GabiNun:
# Safe Edge Disable / Undo Tweak

You removed "remove edge tweak" since it’s unstable, so instead of reinventing the wheel this provides a **safe way to disable Edge from running and stop Edge updates** — and it’s easily undoable.  

Having Edge and Edge Update running in the background may take up a lot of resources, so disabling it is useful.  

I marked this as an **advanced tweak** since I don’t want people to disable Edge and not be able to get it back if they don’t know about the undo tweak.

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2129,6 +2129,36 @@
     ],
     "link": "https://winutil.christitus.com/dev/tweaks/essential-tweaks/wifi"
   },
+  "WPFTweaksDisableEdge": {
+    "Content": "Disable Edge",
+    "Description": "Disables Microsoft Edge via the registry and prevents scheduled tasks from running Edge updates",
+    "Category": "z__Advanced Tweaks - CAUTION",
+    "Panel": "1",
+    "Order": "a016_",
+    "Registry": [
+      {
+        "Path": "HKCU:\\Software\\Microsoft\\Windows\\CurrentVersion\\Policies\\Explorer\\DisallowRun",
+        "Name": "DisableEdge",
+        "Type": "String",
+        "Value": "msedge.exe",
+        "OriginalValue": "<RemoveEntry>"
+      },
+      {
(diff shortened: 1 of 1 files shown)


TommyCowley101:
If you do wanna remove Microsoft edge just use this https://gist.github.com/ave9858/c3451d9f452389ac7607c99d45edecc6 it will remove edge but keep webview stuff so nothing gets broken, and edge wont come back with updates either, also Microsoft Edge Update service might remain, this is normal as it is required for updating WebView2

ghost:
@TommyCowley101

the all point is not removing edge since they removed the "remove edge tweak" since it was unstable

also that gist wont work unless your in the eu

TommyCowley101:
@GabiNun  erm i think you have made a mistake, what i posted works everywhere as well one i don't live in the eu and it works, but yea fair enough i have never had that gist be unstable even on canary versions and have been using it since 23h2 and onwards :)

GabiNun:
@TommyCowley101 I set up a test VM to try the script, but it doesn’t do anything. Also, removing Edge itself doesn’t cause instability it’s when people go overboard with removing to much, You just need to delete the Edge files, and avoid deleting the WebView folder inside the Microsoft directory.

ghost:
@TommyCowley101 You’re right it works even outside the EU

francisuk1989:
Nice work @GabiNun

Before this tweak runs, How about forcing Edge to close? @ChrisTitusTech 

`Stop-Process -Name “msedge” -Force`

cant be done with `msedge.exe` as it give this error.

`PS C:\Users\teacakes> Stop-Process -Name “msedge.exe” -Force`
Stop-Process: Cannot find a process with the name "msedge.exe". Verify the process name and call the cmdlet again.

PR #3649, opened 2025-10-16
Update Disable Edge Tweak

ghost:
### Update on Edge Update Handling

After further testing, I found that even when Edge updates are disabled via Task Scheduler, the `edgeupdate`  service will still trigger Edge Updates.

Additionally, since `Edge Updates` is required for updating WebView, rather than disabling the service, I have decided to remove the `Invoke` script entirely.

also the Order is now correct


francisuk1989:
Before this tweak runs, Can we force Edge to close?
`Stop-Process -Name “msedge” -Force`

ghost:
> Before this tweak runs, Can we force Edge to close? `Stop-Process -Name “msedge” -Force`

i dont think this is necessary because you have to restart your computer to apply the registry changes

PR #3659, opened 2025-10-23
OneDrive Removal Updated

ghost:


Tiagoquix:
I think accepting this PR looks like a regression. Now we don't clear-up the scheduled task and some other folders/entries.

PR #3660, opened 2025-10-23
Duplicates and Typos

ghost:


config/tweaks.json
```diff
@@ -2331,7 +2331,6 @@
       "Microsoft.GetHelp",
       "Microsoft.Getstarted",
       "Microsoft.Messaging",
-      "Microsoft.Microsoft3DViewer",
```
GabiNun: Microsoft.Microsoft3DViewer is also in line 2319

zoicware:
powershell is not a case sensitive language therefore this is unnecessary, also `microsoft.windowscommunicationsapps` is how it appears on the system (lower case)

ghost:
> powershell is not a case sensitive language therefore this is unnecessary, also `microsoft.windowscommunicationsapps` is how it appears on the system (lower case)

i know but it makes it easier to read

PR #3686, opened 2025-11-03
Removed WPFTweaksHome from Minimal preset

ghost:
The services that ```WPFTweaksHome``` disables, dont even exist in windows 11 and since the market share favorites windows 11 removing it from minimal preset make sense

you can verify if they exist with

```powershell
Get-Service HomeGroupListener
Get-Service HomeGroupProvider
```


ghost:
found out it was entirely removed so ye

PR #3687, opened 2025-11-03
Remove WPFTweaksHome

ghost:
HomeGroup has been deprecated and is no longer included in modern versions of Windows, so attempting to disable it on modern systems is unnecessary

https://support.microsoft.com/en-us/windows/homegroup-removed-from-windows-10-version-1803-07ca5db1-7bca-4d11-68a3-a31ff4a09979


Tiagoquix:
Even though it may do nothing, the services are still present in `services.msc`, so I would prefer to keep this tweak for now.

ghost:
> Even though it may do nothing, the services are still present in `services.msc`, so I would prefer to keep this tweak for now.

They are not there, you can test it with
```powershell
Get-Service HomeGroupListener,HomeGroupProvider
```

ChrisTitusTech (approved):
This was leftovers from the early Win10 era. Now that Win10 is EOL... we can remove this. Thanks for the PR.

PR #3709, opened 2025-11-13
Update-WPFTweaksDeBloat

ghost:
Improved for clarity and performance.

Teams doesn’t register itself in the standard uninstall registry paths, so that part of the code has been removed.
I also consolidated the main logic to make it more compact and efficient.

easier view at the code

https://github.com/ChrisTitusTech/winutil/blob/f3186622741643f3411f215d790a87f25ca5cf84/config/tweaks.json#L2359-L2370

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2364,7 +2364,7 @@
         Write-Host \"Uninstalling Teams\"
         Start-Process $TeamsPath -ArgumentList -uninstall -wait
 
-        Write-Host \"Removing Teams directory\"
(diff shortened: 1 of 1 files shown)


ChrisTitusTech (changes requested):
This is actually true and false. There are 4 different installers for MS Teams Microsoft was looking to consolidate these to two one exe/msi version and another appx version. Also when adding things to Appx you need to do a wildcard like `*msteams*`

This will need to be changed

PR #3710, opened 2025-11-13
Update-WPFTweaksRestorePoint

ghost:
This version is faster and simpler.


ChrisTitusTech:
This was how it used to be, but people ripped out system restore then would break their system running too many tweaks. 

We added all these checks so people don't do stupid things without a backup.

PR #3769, opened 2025-12-02
Delete-WPFTweaksRecallOff

ghost:
By deafult recall is disabled with payload microsoft has fixed having it beeing enabled by deafult so this tweak is no longer necessary


Tiagoquix:
@ChrisTitusTech @GabiNun I believe this one could be kept just to make users feel "extra safe" that Recall is disabled (or it could be added to Disable Telemetry).

ghost:
@Tiagoquix thats very unneccary

PR #3779, opened 2025-12-03
Delete-WPFTweaksStorage

ghost:
Storage Sense is a good feature, and there is no reason to disable it especially since it is disabled by default already


ChrisTitusTech (approved):
Move storage sense to advanced tweaks

PR #3793, opened 2025-12-09
Change-Winutil-Directory

ghost:
Programs normally store their data in LocalAppData, but this is a script, so there’s no reason for it to create a folder there. It will now create the folder in `$Env:Temp\winutil` instead of `$Env:LocalAppData\winutil`


ghost:
WPFTweaksDeleteTempFiles well break this

PR #3807, opened 2025-12-12
Update-WPFTweaksDisableExplorerAutoDiscovery

ghost:
Much better

btw i always test my pr's i never make a pr without testing it and making sure it works as expected

i could have made removed some of the invoke script and put it as the registery thing you made but i think its more readable this way


ghost:
for some reason this dosen't work and i dont know why

PR #3809, opened 2025-12-12
Bring-Back-#3710

ghost:
you close this pr because

```
This was how it used to be, but people ripped out system restore then would break their system running too many tweaks.

We added all these checks so people don't do stupid things without a backup.
```

but your code wouldn't even fix restore point if it was broken

you tried to `Import-Module Microsoft.PowerShell.Management` but if that module isn't loaded than the entire script wouldn't work

also the only way you could "break" restore point is if you disabled it or just stright up deleted system restore files which you couldn't fix

also people who break system restore point wouldn't run system restore

also fixes this warning
```
WARNING: The 'Microsoft.PowerShell.Management' module was not imported because the 'Microsoft.PowerShell.Management' snap-in was already imported.
```


ChrisTitusTech:
Did you test this on PS5? I think I remember in the past Computer-Checkpoint didn't work on Windows Powershell and only did on the newer version. That is the only other reason why we added the repair on System Restore.

ghost:
@ChrisTitusTech yes it dose work in powershell 5

PR #3817, opened 2025-12-15
Update-Compile.ps1

ghost:
winutil.ps1 well get overridden anyways


ChrisTitusTech:
I'd rather it just remove it before creating a new file.

PR #3822, opened 2025-12-18
Update-Remove-WinUtilAPPX.ps1

ghost:
Updated to use -AllUsers. Removed unnecessary error suppression. Removed quotes since they are already defined in the config file. Removed ** for the same reason, as it is handled in the config file.

i also removed the try and catch since it wouldn't be useful and there is no reason for this to fail at all


CodingWonders:
> and there is no reason for this to fail at all

You **REALLY** sure about that?

ghost:
@CodingWonders i am

ChrisTitusTech (changes requested):
I'd recommend leaving at least -ErrorAction SilentlyContinue incase it does fail. It shouldn't but this is Microsoft we are talking about.

ghost:
@ChrisTitusTech if it dose fail it wont give an error

PR #3827, opened 2025-12-19
Update-WPFTweaksRightClickMenu

ghost:
UndoScript dose not need to restart explorer

# ACCEPT https://github.com/ChrisTitusTech/winutil/pull/3835 BEFORE ACCEPTING THIS

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2246,7 +2246,7 @@
     "InvokeScript": [
       "
       New-Item \"HKCU:\\Software\\Classes\\CLSID\\{86ca1aa0-34aa-4e8b-a509-50c905bae2a2}\" -Name InprocServer32 -Value '' -Force
-      Invoke-WinUtilExplorerUpdate -action restart
+      Invoke-WinUtilExplorerUpdate
       "
     ],
     "UndoScript": [


config/tweaks.json
```diff
@@ -2245,17 +2245,13 @@
     "Order": "a027_",
     "InvokeScript": [
       "
-      New-Item -Path \"HKCU:\\Software\\Classes\\CLSID\\{86ca1aa0-34aa-4e8b-a509-50c905bae2a2}\" -Name \"InprocServer32\" -force -value \"\"
-      Write-Host Restarting explorer.exe ...
-      Stop-Process -Name \"explorer\" -Force
+      New-Item \"HKCU:\\Software\\Classes\\CLSID\\{86ca1aa0-34aa-4e8b-a509-50c905bae2a2}\" -Name InprocServer32 -Value '' -Force
+      Invoke-WinUtilExplorerUpdate
```
CodingWonders: I provided a review on #3835 that means that you should update this call.

CodingWonders (changes requested):

config/tweaks.json
GabiNun: that is kinda weird if i check the current dark mode tweak it just says `Invoke-WinUtilExplorerUpdate` and not `Invoke-WinUtilExplorerUpdate -action \"restart\"` nor `Invoke-WinUtilExplorerUpdate -action \"refresh\"` but it well still work just fine

CodingWonders: You can either explicitly add the -action parameter, or leave it blank and just use the default value which is "refresh". It's not a mandatory parameter.

GabiNun: @CodingWonders cool but it still works fine

ChrisTitusTech:
Not needed

PR #3828, opened 2025-12-19
Update-preset.json

ghost:
Most people that run standard tweaks aren't using powershell 7


CodingWonders:
What if they do? What if Microsoft decides at some point to bundle PWSH7 with Windows?

ghost:
@CodingWonders than we well all be happy and change 1 line

PR #3830, opened 2025-12-19
Update-WPFFeatureRegBackup

ghost:
no reason to be doing all of that


ghost:
@CodingWonders i tested it and it dose work i never make a pr without testing it

ChrisTitusTech (commented):
Microsoft does not enable backups by default and has disabled registry backups multiple times. https://learn.microsoft.com/en-us/troubleshoot/windows-client/installing-updates-features-roles/system-registry-no-backed-up-regback-folder

The extra properties specify how many copies and schedule the task to take place on a schedule

PR #3831, opened 2025-12-19
Update-WPFFeaturenfs

ghost:
no reason to be doing all of that


ChrisTitusTech:
NFS is very particular and here is a guide: https://4sysops.com/archives/mount-an-nfs-share-on-windows-11-with-powershell/

The big thing is how it handles the authentication with UID/GID and why the extra values are added.

I personally use persistent shares on all my machines to connect to my NAS drives.

PR #3833, opened 2025-12-19
Update-Updates-Tab

### LINKED ISSUE #3881: User Profiles disappears after reverting to Default Windows Update Configuration
### ⚙️ Issue Checklist

- [x] I checked for an existing pull request that addresses this issue.

### 📜 What part of Winutil are you having issues with?

Updates - Default Settings (Default Windows Update Configuration)

### 📝 Provide a clear and concise description of the issue.

Hi There, 
After Debloating my Windows 11 25H2 (OS Build 26200.7462) I noticed that Delivery Optimisations were greyed out. As I wanted to enable Delivery optimisation, So that my other devices can fetch updates from local computer on LAN, I tried reverting the settings by going to Updates - Default Settings (Default Windows Update Configuration) & After restart of PC, I noticed that all my other Windows User profiles have disappeared & Only the Admin Profile remained.

Winutil Version used 25.12.18

### 🔄 Steps to reproduce the issue.

1. Run the tool in powershell with administrator privilages
2. Go to Updates
3. Default settings
4. Restart PC

### ❌ Paste the full error output (if available).

No Error Messages, But all other user profiles did not show up in lock screen .

I repeated the same process two times after doing system restore, The same issue reccurs.

### 🖼️ Additional context.

Tried repeating the same process after system restore & same issue continues.

ghost:
### This keeps the same functionality while improving the code quality.

- removed `Out-Null` since it was unnecessary
- removed many unnecessary `-ErrorAction SilentlyContinue` usages and added global error suppression via `$ErrorActionPreference = 'SilentlyContinue'`
- changes are difficult to spot inline, so it is probably better to open the file in a new tab and compare

I also removed comments that were placed next to `Write-Host`. For example:

removed: `# Reset WaaSMedicSvc registry settings to defaults`
kept:    `Write-Host "Restoring WaaSMedicSvc settings..."`

Additional changes:
- removed the message box from `Invoke-WPFUpdatessecurity.ps1` since other files do not use one and it is unnecessary
- removed disabling the `uhssvc` service because it does not exist
- removed DLL renaming since it breaks Microsoft Store app installations (#3859)
- removed the “Additional registry settings” section because it attempts to modify a service that is being disabled
- removed `gpupdate /force` since the user would have to restart there computer anyway
- removed deleting `GroupPolicy` and `GroupPolicyUser` folder since it would undo other group policy changes
- adjusted the default settings logic to remove policies instead of reconfiguring them
- I also greatly improved the time it takes to run these tweaks by about 30 seconds

fixes #3881

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -848,14 +848,6 @@
         "link": "https://gerardog.github.io/gsudo/",
         "winget": "gerardog.gsudo"
     },
-    "guilded": {
-        "category": "Communications",
-        "choco": "na",
-        "content": "Guilded",
-        "description": "Guilded is a communication and productivity platform that includes chat, scheduling, and collaborative tools for gaming and communities.",
-        "link": "https://www.guilded.gg/",
-        "winget": "Guilded.Guilded"
-    },
     "handbrake": {
         "category": "Multimedia Tools",
         "choco": "handbrake",


ChrisTitusTech (commented):
This looks much cleaner, but I need to testing this on stream as Microsoft has a way of re-enabling Windows Update using a variety of methods and not sure this will be enough.

ghost:
@ChrisTitusTech well it didn't remove many of your methods mostly the renaming of dlls but i tested it and it works but you testing it also is a good idea and it also fixes alot of bugs

ChrisTitusTech (approved):

PR #3835, opened 2025-12-20
Update-Invoke-WinUtilExplorerUpdate.ps1

GabiNun2:
We never even use the refresh action

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -2634,14 +2634,10 @@
       }
     ],
     "InvokeScript": [
-      "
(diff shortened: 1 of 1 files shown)


CodingWonders:
The dark mode toggle **DOES** use the refresh action.

ChrisTitusTech:
The snap assist also needs the refresh action. I remember there being a few different tweaks. However, there is a bug in that code. There needs to be a Type-existence check in before adding a type for when people run this function more than once in the same session.

PR #3843, opened 2025-12-22
Update-windev.ps1

ghost:
This will fetch the latest release tag, whether it’s a pre-release or not, so there’s no need for a fallback when a pre-release isn’t available.

I simplified the script, making it faster and more efficient.

I also removed self-elevation, since winutil already handles that. windev.ps1 shouldn’t perform self-elevation itself, which would cause it to elevate twice unnecessarily.

also do you think you can make cloudflare automaticly get the pre release instead of having windev.ps1


Real-MullaC:
Has this been tested?

ghost:
@Real-MullaC yes i always test my PRs i never make a pr without testing it as best as i can

https://github.com/user-attachments/assets/09e4d926-95aa-412b-9773-0ec0e00c23e4

Real-MullaC:
> @Real-MullaC yes i always test my PRs i never make a pr without testing it as best as i can
> 2026-01-01.02-32-23.mp4

You are aware. That build you are looking at is a Latest Build. So currently there is no proper way of testing this till a Pre Release comes out.

ghost:
@Real-MullaC i tested it while it was a pre-release also my code works well it just grabs the latest release no matter if its a pre-release or not

ChrisTitusTech (approved):
Well that is a hell of a lot easier. Good job!

PR #3844, opened 2025-12-23
Update Invoke-WinUtilInstallPSProfile.ps1

ghost:
For some reason you tried to recreate what the setup.ps1 from the repo dose i dont know why you did that

anyways this well just create another powershell windows where the user could go over the prompts and it dosen't effect winutil

and i improved the backup feature


Real-MullaC (changes requested):
If you do this way you need to keep that part in about pwsh 7 as his profile does not work on powershell 5 and it should really run on powershell 7 as if it runs on powershell 5 and they don't have powershell 7 then there is no point installing the profile.

ghost:
> If you do this way you need to keep that part in about pwsh 7 as his profile does not work on powershell 5 and it should really run on powershell 7 as if it runs on powershell 5 and they don't have powershell 7 then there is no point installing the profile.

we should change the text inside winutil to say that it is powershell 7 only

PR #3845, opened 2025-12-23
Update Invoke-WinUtilUninstallPSProfile.ps1

ghost:
### Check out #3844 for more description since they are connected (this should've been 1 pr)

You original script uninstalled programs that are dependencies for your powershell profile but those programs could be dependencies for other things on the users system, so not removing them is the best course, also the user could just uninstall them manualy, and they dont take up much space at all


Cryostrixx:
@GabiNun The whole purpose of the changes I made to this feature was to ensure users wouldn’t lose their profile or have to restore the backup manually when uninstalling the profile. I carefully looked into this before adding the uninstall feature—there were cases where users didn’t know about the backup the install script made and thought their profile was lost forever. Removing these safety and convenience measures would effectively make the installer worse, and honestly, would make it harder to use.

Also, if you carefully examine the code in `Invoke-WinUtilUninstallPSProfile`, you’ll see that it scans the user’s original profile for optional tools and keeps them intact if they are found. Only if it doesn’t find OhMyPosh, Zoxide, or Terminal-Icons in the user's original profile does it uninstall them from the system, with Nerd Fonts always being removed since they are easy to reinstall. I want to emphasize that removing ~99% of the code from functions like `Invoke-WinUtilInstallPSProfile` and `Invoke-WinUtilUninstallPSProfile` would eliminate all the logic that preserves modules and tools installed manually by the user, which I don’t think is a safe or desirable change.

@ChrisTitusTech What’s your take on this? Is this acceptable to merge, or should it be rejected?

GabiNun:
@Cryostrixx i could add a backup feature so if the user runs the invoke script it would rename there profile and when they undo it could name it back, but i stand on not removing those programs

Cryostrixx:
> @Cryostrixx i could add a backup feature so if the user runs the invoke script it would rename there profile and when they undo it could name it back, but i stand on not removing those programs

Would it be worth the effort, though? The uninstall feature I implemented already does this, you can see how it restores the user's original profile here:
```powershell
# Attempt to move the user's original PowerShell profile backup back to its original location.
try {
    # Check if the backup PowerShell profile exists before attempting to restore the backup.
    if (Test-Path "$PSProfile.bak") {
        # Restore the backup PowerShell profile and move it to its original location.
        Move-Item "$PSProfile.bak" $PSProfile
        
        # Let the user know that their PowerShell profile backup has been successfully restored.
        Write-Host "===> Restored Profile Backup. <===" -ForegroundColor Yellow
    }
} catch {
    # Let the user know that an error was encountered when restoring the profile backup.
    Write-Host "Failed to restore profile backup. Error: $_" -ForegroundColor Red
}

# Silently cleanup the oldprofile.ps1 file that was created when the CTT PowerShell profile was installed.
Remove-Item "$env:USERPROFILE\oldprofile.ps1" | Out-Null
```

And takes the backup here:
```powershell
if (Test-Path "$env:USERPROFILE\oldprofile.ps1") {
    Write-Host "===> Backup File Exists... <===" -ForegroundColor Yellow
    Write-Host "===> Moving Backup File... <===" -ForegroundColor Yellow
    Copy-Item "$env:USERPROFILE\oldprofile.ps1" "$PSProfile.bak"
    Write-Host "===> Profile Backup: Done. <===" -ForegroundColor Yellow
} else {
    # If oldprofile.ps1 does not exist use $PSProfile as a profile backup source.
    # Check if the profile backup file has not already been created on the disk.
    if ((Test-Path $PSProfile) -and (-not (Test-Path "$PSProfile.bak"))) {
        # Let the user know their PowerShell profile is being backed up.
        Write-Host "===> Backing Up Profile... <===" -ForegroundColor Yellow

        # Copy the user's current PowerShell profile to the backup file path.
        Copy-Item -Path $PSProfile -Destination "$PSProfile.bak"

        # Let the user know the profile backup has been completed successfully.
        Write-Host "===> Profile Backup: Done. <===" -ForegroundColor Yellow
    }
}
```

Cryostrixx:
> @Cryostrixx i added a backup feature so if the user runs the invoke script it would rename there profile and when they undo it would rename it back, but i stand on not removing those programs
> 
> and whats wrong with removing ~99% of the code

@GabiNun Please see the latest response. I already added the auto-backup-restore logic into the uninstall functionality when I added the feature in... There's no reason to add more code.

GabiNun:
@Cryostrixx already added it and i didn't add more code i greatly reduced it

i also made it more simple there was really no reason for it to be that complicated

i also made it so if they do run the undo it immediately makes there profile work so they dont have to find it

GabiNun:
@Cryostrixx some of the points you make, make no sense like why would the user wont be able to find `OhMyPosh/Terminal-Icons/Zoxide` im not removing them nor would it fail its a simple rename

well you updated your comment so this is kinda akward

Cryostrixx:
> @Cryostrixx some of the points you make, make no sense like why would the user wont be able to find `OhMyPosh/Terminal-Icons/Zoxide` im not removing them nor would it fail its a simple rename
> 
> well you updated your comment so this is kinda akward

It's not that the user can't find it, as much as cleaning up dependencies that are no longer needed by the profile... It's assumed that if they aren't in the user's original profile, they don't want them, so I detect that and uninstall them if they are found. Otherwise, they are left alone. I don't see the harm with my original implementation; it does work, but it also increases the complexity of the code to provide real benefit for anyone who uses this feature. The way I chose to do backups also echoes this design of letting the user know what the script is doing; you can see this design throughout WinUtil, despite the complexity of doing it this way and notifying the user of what it's doing at every install or uninstall step. If there was no visible output, the user would think it had stalled at every step it's taking.

ghost:
> This should be left how it is as if people are uninstalling his profile they are not going to want any of the software that come with it.

than can just remove it and some people use it

ChrisTitusTech (approved):
Yeah I agree to just remove the profile to revert to stock no reason to try and remove nerd fonts or zoxide. 
This is crazy bloated.

PR #3846, opened 2025-12-23
Update-Install-WinUtilChoco.ps1

ghost:
Using -UseBasicParsing  makes it so it dosen't ask you to confirm execution, and you dont need the other things choco already dose them for you


Real-MullaC (commented):
This should kept in a try catch just so if there is any issues with it installing it will say.

ghost:
@Real-MullaC ye but the catch message dosen't do anything anyway

PR #3852, opened 2025-12-27
Update-tweaks-order

ghost:
`Old`
Old

`New `
{0C2B52DB-540C-4EF7-B7E5-FFEFF8BBEA1F}


Real-MullaC (changes requested):
These should all be separate numbers so there is no issues with anything.

ghost:
> These should all be separate numbers so there is no issues with anything.

i tested it there are no issues

ChrisTitusTech (approved):
I hate this ordering system, I forget why we did this, but its awful.

Remind me to rewrite this to remove Order and just do everything with AlphaNumeric ordering.

ChrisTitusTech (approved):

PR #3855, opened 2025-12-27
Delete-WPFTweaksDisableExplorerAutoDiscovery

ghost:
This is a useless tweak

It dosen't slow down broswing

And it also makes it so explorer dosen't remember the group by you set which is very annoying

example

with Explorer Auto Discovery disabled
{CAC9A36A-5185-40E1-9755-17D479EDD316}

with Explorer Auto Discovery enabled
{598DA375-BDC0-4CBF-9F84-092AF7F26AE7}

so basicly it well make it so you can't change the group by in explorer


CodingWonders (changes requested):
Open a folder with many music files and auto discovery enabled. Test the performance in that scenario.

PR #3856, opened 2025-12-27
Update Invoke-WPFFixesWinget.ps1

ghost:
This is the best way to of fixing winget.


CodingWonders:
@Real-MullaC, I agree and disagree with your statement.

- I agree with you in the sense that we should leave options. What if the nuget way fails? We can then use chocolatey, but
- I disagree with you in the sense that the user strips NuGet and the module, because it will still install the NuGet package manager and the module

Why not have both options?

ghost:
@CodingWonders it is better than choco since it basicly dose the same thing just downloads winget and dependencies from github and install it so it shouldn't fail

Issue #3861, opened 2025-12-30
Add Security hardening


ghost:
chris, in one of your streams you mentioned going into group policy to change multiple settings to harden security. it would be great to have a tweak like this, or even a button in the config tab that applies those policies automatically.

ghost:
this dosen't fit winutil

PR #3862, opened 2025-12-30
Add-WPFTweaksHiber-To-Preset

ghost:
# check out this cool feature #3861


ghost:
@Real-MullaC Hibernate dose exist on a desktop computer and this is a disable tweak so yes you should use it on pc

ChrisTitusTech:
We've gone back and forth on this over the years, but we settled on leaving it out because people just don't understand and end up removing it on their laptop which is where it is used most.

I hate hibernation and always disable it on every machine, and yes, it is on desktop and laptop.

ghost:
@ChrisTitusTech what about adding a - Not recommended for laptops

ChrisTitusTech:
> @ChrisTitusTech what about adding a - Not recommended for laptops

Check out https://github.com/hardentools/hardentools @GabiNun pretty neat security software at removing a lot of windows attack surfaces.

ghost:
@ChrisTitusTech no it wouldn't be for me im talking about for winutil

PR #3878, opened 2026-01-06
Update-WPFTweaksTele

ghost:
this is incorrect and can be verified by running the tweak. `WARNING: Unable to set PeriodInNanoSeconds due to unhandled exception`

to set the feedback frequency to 0, you must create the dword
`NumberOfSIUFInPeriod` and set it to `0`.

the qword `PeriodInNanoSeconds` must not be present, but it does not
exist by default, so removing it is unnecessary.

as a result, the section that removes `PeriodInNanoSeconds` should be replaced with `NumberOfSIUFInPeriod`

[Sources](https://www.ninjaone.com/blog/how-to-change-windows-10-feedback-frequency)

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -1441,13 +1441,6 @@
         "Type": "DWord"
       },
       {
-        "Path": "HKCU:\\SOFTWARE\\Microsoft\\Siuf\\Rules",
-        "OriginalValue": "<RemoveEntry>",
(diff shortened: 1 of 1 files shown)


ghost:
> This looks like a dupe and need to look at this from Microsoft's knowledge base. Look at line 1437

didn't notice that, than i well just change the pr not not have that part

PR #3880, opened 2026-01-06
Update-WPFTweaksServices

ghost:
Fix some OriginalType'a in WPFTweaksServices (based on
```powershell
$Services = (irm https://github.com/ChrisTitusTech/winutil/raw/main/config/tweaks.json).WPFTweaksServices.service
$SystemServices = Get-Service | Select-Object Name, StartType

$MismatchedServices = $Services | ForEach-Object {
    $Service = $_
    $SystemService = $SystemServices | Where-Object { $_.Name -eq $Service.Name }
    if ($SystemService -and $SystemService.StartType -ne $Service.OriginalType) {
        [PSCustomObject]@{
            Name          = $Service.Name
            OriginalType  = $Service.OriginalType
            CurrentType   = $SystemService.StartType
        }
    }
}

$MismatchedServices


# OriginalType is what service tweak reverts to and current type is what it should revert to based on a fresh install of windows 11 25H2
Name         OriginalType CurrentType
----         ------------ -----------
BthAvctpSvc  Automatic         Manual
InventorySvc Manual         Automatic
KeyIso       Automatic         Manual
Netlogon     Automatic         Manual
TermService  Automatic         Manual
VaultSvc     Automatic         Manual
WlanSvc      Automatic         Manual
camsvc       Manual         Automatic
```

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -454,11 +454,6 @@
         "OriginalType": "Disabled"
       },
       {
-        "Name": "Netlogon",
-        "StartupType": "Automatic",
-        "OriginalType": "Manual"
-      },
-      {
         "Name": "Netman",
         "StartupType": "Manual",
         "OriginalType": "Manual"
@@ -832,11 +827,6 @@
         "Name": "Winmgmt",
         "StartupType": "Automatic",
         "OriginalType": "Automatic"
-      },
-      {
-        "Name": "WlanSvc",
-        "StartupType": "Automatic",
-        "OriginalType": "Manual"
       },
       {
         "Name": "WpcMonSvc",


ChrisTitusTech (changes requested):
The problem with this is the default types do change based on certain systems.

For example: If the system is domain joined when it is setup the NetLogon service is set to Automatic. I think we really shouldn't be touching these services and probably should remove them, specifically the WLanSvc and Netlogon as those I know should not be messed with as the defaults change from system to system.

ghost:
> The problem with this is the default types do change based on certain systems.
> 
> For example: If the system is domain joined when it is setup the NetLogon service is set to Automatic. I think we really shouldn't be touching these services and probably should remove them, specifically the WLanSvc and Netlogon as those I know should not be messed with as the defaults change from system to system.

okay than i removed those services from the list

PR #3888, opened 2026-01-09
Update-Invoke-WPFUltimatePerformance.ps1

ghost:
I simplified the main logic.


ChrisTitusTech (changes requested):
Change to list for disable as this is only grabbing current plan.

ghost:
@ChrisTitusTech done

PR #3906, opened 2026-01-13
Remove-WPFToggleTaskbarWidgets

ghost:
We can't disable Widgets without running
```powershell
Get-AppxPackage MicrosoftWindows.Client.WebExperience -AllUsers | Remove-AppPackage -AllUsers
```
and that would be a remove not a disable, the only way to disable it would be trough settings since for some reason windows locked down that registery dword


ChrisTitusTech (approved):
Yup, back in the day this worked, but no longer. Sad face :(

PR #3938, opened 2026-01-25
Disable BitLocker on Home Editions with microwin due to #3937

### LINKED ISSUE #3937: MicroWin on Win11 Home have drive encryption off by default
### ⚙️ Issue Checklist

- [x] I checked for an existing pull request that addresses this request.

### ❓ Is your feature request related to a problem?

I've noticed in the various Windows 11 Home MicroWin installs I've done, the drive encryption (I'm calling it BitLocker Lite) is enabled by default.  My assumption is that if Secure Boot / TPM are disabled that this feature probably can't activate on its own, but I guess because the original image is based around requiring a MS account, its kinda null and void on most normal installs.  However since MicroWin enforces a local account, this creates an impasse where users are probably not aware that its active by default, and so in the event of hardware failure but the install disk is still fine the data likely can't be recovered.

The same isn't true on Windows 11 Pro MicroWin installs, regardless of Secure Boot / TPM it appears to be hard disabled always, this is expected behaviour in my opinion.

If at all possible, make all MicroWin images have BitLocker / Drive Encryption off by default given they force a local account, this makes logical sense to me.  Potentially a tweak that gives us the option to remove the feature so it can't reactivate itself, especially without the users knowledge, which would mean they wouldn't think to disable or save the recovery key.

### 🔄 Describe alternatives you've considered

_No response_

### 🖼️ Additional context

_No response_

ghost:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description

This Checks if BitLocker is enabled and if it is enabled than it well disable it

## Issue related to PR

- Resolves #3937

## Checklist
- [ ] I have made corresponding changes to the documentation.

### DIFF
--- a/functions/microwin/Microwin-NewFirstRun.ps1
+++ b/functions/microwin/Microwin-NewFirstRun.ps1
@@ -66,6 +66,11 @@ function Microwin-NewFirstRun {
 
     }
 
+    if ((Get-BitLockerVolume -MountPoint $Env:SystemDrive).ProtectionStatus -eq 'On') {
+        Write-Host "Disabling BitLocker..."
+        Disable-BitLocker -MountPoint $Env:SystemDrive
+    }
+
     # Get BCD entries and set bootmgr timeout accordingly
     try
     {


KonTy:
This is not the to do it, why make decisions for everyone, it should be an option a checkbox on the mirowin page, maybe ticked by default.

ghost:
> This is not the to do it, why make decisions for everyone, it should be an option a checkbox on the mirowin page, maybe ticked by default.

Chris wants Microwin to be as streamlined as possible, so even if I made this optional, he would probably not accept it. I also think most people who use Microwin won’t want BitLocker, and this doesn’t remove BitLocker you can always enable it later.

PR #4024, opened 2026-02-10
added frontmatter to pages

seanh1995:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Documentation update
- [ ] UI/UX improvement

## Description

- added frontmatter to widget.md
- added frontmatter to XboxRemoval.md‎
- added automatic year change and link to https://christitus.com/ in the copyright section

## Note
i had to use the old way off doing the javascript as for some reason hugo does not like the new way out off the box and i would have to make a new _partials/footer.html whitch just over complicates things futher

## Issue related to PR

- Resolves #

### DIFF
--- a/docs/i18n/en.yaml
+++ b/docs/i18n/en.yaml
@@ -1,4 +1,4 @@
-copyright: "© Chris Titus Tech. All rights reserved."
+copyright: "&copy; <script>document.write(new Date().getFullYear())</script> <a href='https://christitus.com'>Chris Titus Tech</a>. All rights reserved."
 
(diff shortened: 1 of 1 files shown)


ghost:
yep you should check all of the .md file because i didn't add frontmatter to all of the new once i made

seanh1995:
@GabiNun ill check but i think that was the only one because all the other file must off just been edited and not new ones. so the edited ones will be ok. i checked there was just 2 there. even if we forget to do the frontmatter my code should just fall back the file name instead. so it wont be missed from the pages ether way.

Issue #2993, opened 2024-10-29
Make the GUI translatable to more languages


sailinggalaxians:
A lot of the people helping out to make this great program are native English speakers, but lets say a user in Sweden can't read English as good as they read Swedish and the program's GUI is seen in English and there's no way to change to Swedish because it doesn't exist or any other language for that matter.

As a solution I'd like to see that this program uses Weblate to sync translations when people save the changes. 

There are no known solutions to my knowledge.

sailinggalaxians:
Ping! Is the setup for this extremely hard? I'm not angry, but rather eager to help start translating it. :)

swefd:
Ping! The issue is not closed and not resolved. We need a way to contribute. I think many people would help to translate it.

vagman:
Still a relevant and useful feature, I think. I can handle the Greek translation if this is implemented in the future.

swefd:
> Still a relevant and useful feature, I think.

There is already a PR with a feture

ghost:
> There is already a PR with a feture

he's trying to make github-actions remove the stale label

FluffyPunk:
I'm not sure if having this open makes much sense, since it was also discussed in #4032 that adding multi-language support will only make script bigger, so all the aspects of that script will become harder - compiling, downloading, running etc.
This script is PowerShell script and that's already technical downside, so the best solution is to ask @ChrisTitusTech for making initial upload of F# re-written WinUtil, so it will become technically more optimal and resourcefully optimal to add i18n.

ghost:
> I'm not sure if having this open makes much sense, since it was also discussed in [#4032](https://github.com/ChrisTitusTech/winutil/pull/4032) that adding multi-language support will only make script bigger, so all the aspects of that script will become harder - compiling, downloading, running etc. This script is PowerShell script and that's already technical downside, so the best solution is to ask [@ChrisTitusTech](https://github.com/ChrisTitusTech) for making initial upload of F# re-written WinUtil, so it will become technically more optimal and resourcefully optimal to add i18n.

so if you want winutil in a diffrent langague you need to pay 10$ ? because that's a good idea

ChrisTitusTech:
At this time I'm not going to be adding addition language support. At some point in the future I will though.

PR #4026, opened 2026-02-10
feat: Multi-Language Support and Vector Flag Icons

Lib-LOCALE:
This PR adds comprehensive multi-language support (English, French, Spanish, German, Italian, Portuguese, Polish) and implements a robust vector-based flag icon system for the language selection menu.

Key changes:
- Developed a JSON-based localization infrastructure.
- Added 5 new languages and updated existing ones.
- Implemented sharp, custom XAML vector flags to replace unreliable emoji icons.
- Resolved UTF-8 encoding issues in the compilation process.
- Optimized UI elements (navigation, tweaks, updates) for dynamic localization.


ghost:
Add

Resolves #2993

ChrisTitusTech:
The question is if we add all these locales... that ends up being one massive ps1 file we would be serving. Is this practical for a powershell script distributed to millions?

KamaleiZestri:
The other implementation method is to write a specific language into the .ps1 at compile time and have a bunch of versions like `winutil-eng.ps1`, `winutil-de.ps1`, etc.

Or following with this approach, you could do something like copyparty does and have an only english version to keep the current file size and also a full language version with all the other languages included.

Lib-LOCALE:
@ChrisTitusTech I understand the concern about file size, but we need to be realistic: English is only spoken by about 17% of the global population. For a tool as essential as WinUtil, assuming everyone should just 'deal with English' isn't just a technical choice, it's a barrier for millions of users in Europe, LatAm, and Asia who want to optimize their systems without a dictionary.

I’ve done the heavy lifting to make this project truly international, and it would be a shame to limit its reach just for the sake of a few kilobytes.

To address your technical concern without sacrificing accessibility:
We don't have to bundle everything into one monolithic .ps1. We can move the locales to external JSON files and have the script fetch only the one matching the user's system language at runtime.

This keeps the main script lightweight while finally making it a world-class tool, not just an English-speaking one. Let's make it easy for everyone, not just for those born in Dallas. 🤠

Ready to implement the dynamic loading if you're on board with the vision.

swefd:
+1 for the feature

KOUKOU221:
this would be great addition if it included time zones

Lib-LOCALE:
✨ **Clean Code & Localization Update**

I've refactored the implementation to align with project standards and 'Clean Code' principles:

1. 🎯 **Correct UI Integration**: Moved flag rendering to Initialize-InstallAppEntry.ps1 (the dedicated function for app creation) instead of the generic Invoke-WPFUIElements.
2. 🧹 **Dead Code Removal**: Reverted all experimental changes in Invoke-WPFUIElements to keep the codebase clean.
3. ⚡ **Optimized Loading**: Standardized Get-WinUtilCountryFlag.ps1 to use lightweight XAML vectors (matching Get-WinUtilFlagIcon pattern), removing all external dependencies.
4. 🌍 **Full Coverage**: Verified all **372 applications** now have valid ISO 3166-1 alpha-2 origin codes.
5. 🚀 **Locale Fix**: Optimized Invoke-WinUtilLocaleUpdate.ps1 by removing redundant API calls.

Ready for review! 🚀

ChrisTitusTech:
This would need to make seperate ps1 files and beyond the scope for this project.

Lib-LOCALE:
@ChrisTitusTech so your app it's English ONLY ? you are a bad guy really ...

PR #4072, opened 2026-02-18
Add undo to WPFTweaksXboxRemoval

ghost:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description

There isn't a better way since you need an invokescript for a undoscript

## Issue related to PR

- Resolves #


ChrisTitusTech:
We are relying on invoke scripts too much. I'm going to rework all this to not rely on it.

ghost:
> We are relying on invoke scripts too much. I'm going to rework all this to not rely on it.

how so?

PR #4092, opened 2026-02-19
NoUI mode and Concurrency edits

KamaleiZestri:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
Following on from #4051, this PR adds the `-noui` run switch for winutil. When combined with `-run` and `-config test.json`, winutil will run all the tweaks,toggles,features, and installs listed in test.json without loading the UI.

Changes:
- Create `Invoke-WPFUIThread.ps1` in order to add a guard clause preventing wpfform threads from running with `-noui`.
- Add `-noui` guard clause to `Set-WinUtilProgressbar.ps1`
- Add `-noui` guard clause to `Invoke-WPFImpex.ps1`
- Removed `$DebugPreference` from each call to `Invoke-Runspace` and instead set it once for the entire pool.
- Also pass `-noui` value into runspaces.
- Moved loop from `Invoke-WinUtilFeatureInstall.ps1` to the caller `Invoke-WPFFeatureInstall.ps1` to match with the tweaks equivalent.
 - Moved some $sync.variables inside Invoke-Runspace scriptblocks.
- Move package manager preference load and applications config to load earlier in `main.ps1`, before UI.


ChrisTitusTech:
Great PR, this also seemed to make the script launch a little faster as well.

PR #4099, opened 2026-02-21
Add FilePilot application entry

NabeelOG:
Adds FilePilot to the application installer list under the Utilities category.

FilePilot is a modern Windows file explorer with tabs, dual-pane support, 
and advanced file management features. It can be set as a Windows Explorer 
replacement after install.

## Type of Change
- Installs via winget (Voidstar.FilePilot)
- No Chocolatey package available (choco field left empty)

Capture


Kingproone:
the app list already has https://github.com/files-community/Files, which is open source and very popular, this on the other hand seems closed source with a very high license price for the full version

PR #4114, opened 2026-02-25
Composable Configs

KamaleiZestri:
## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description

This PR allows for the .json files in `config` to be seperated out into multiple files and joined under the `config/FOLDERNAME` name at compilation.

These .json files could be seperated out into categories(like [applications here](https://github.com/KamaleiZestri/winutil/tree/compose-examples/config/applications)), into individual files (as shown by [tweaks/services.json](https://github.com/KamaleiZestri/winutil/tree/compose-examples/config/tweaks) here), or just left whole like dns.json.

Also individual .json files can be excluded from the resulting `winutil.ps1` at compile. For example `./compile.ps1 -Exclude "browsers.json,advancedtweaks.json"`.


KamaleiZestri:
Ran a python script to split up applications. It didn't work for tweaks or features though because they include non-standard json.

```python
import json
import os

def split(fileName:str):
    seperations = {}
    folderName = fileName.split(".")[0]
    if not os.path.exists(folderName):
        os.makedirs(folderName)

    with open(fileName, "r") as file:
        loadedJson = json.load(file)
        for key, value in loadedJson.items():
            category = value["category"]
            if not category in seperations.keys():
                seperations[category] = {}
            seperations[category].update({key:value})

    for key,value in seperations.items():
        prettyKey = key.replace(" ","").lower()
        with open (f"{folderName}/{prettyKey}.json", "w", encoding="utf-8") as out:
            json.dump(value,out,indent=3)


split("applications.json")
# split("tweaks.json")
# split("feature.json")

```

ghost:
i think this is unnecessary, also your Complie & Check are failing

KamaleiZestri:
> i think this is unnecessary, also your Complie & Check are failing

Fair, I see how this could be more work than good. It would be useful for something like seperate compiles for each language. No hard feelings if this is just closed as an unneeded feature though.

As for the test failing, I will look into that. It is successfully compiling on my local machine.

ghost:
> > i think this is unnecessary, also your Complie & Check are failing
> 
> Fair, I see how this could be more work than good. It would be useful for something like seperate compiles for each language. No hard feelings if this is just closed as an unneeded feature though.
> 
> As for the test failing, I will look into that. It is successfully compiling on my local machine.

ye no hard fellings

ChrisTitusTech:
But why? 

This would make searching worse and the files aren't very big.

Issue #4155, opened 2026-03-04
Standardize JSON files


ChrisTitusTech:
#### Re-work the json files to get away from invoke scripts. Example of a snippet that would need to be changed.

```
"WPFToggleDarkMode": {
    "Content": "Dark Theme for Windows",
    "Description": "Enable/Disable Dark Mode.",
    "category": "Customize Preferences",
    "panel": "2",
    "Type": "Toggle",
##############THIS IS GOOD#############
    "registry": [
      {
        "Path": "HKCU:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize",
        "Name": "AppsUseLightTheme",
        "Value": "0",
        "Type": "DWord",
        "OriginalValue": "1",
        "DefaultState": "false"
      },
      {
        "Path": "HKCU:\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize",
        "Name": "SystemUsesLightTheme",
        "Value": "0",
        "Type": "DWord",
        "OriginalValue": "1",
        "DefaultState": "false"
      }
    ],
###############THIS NEEDS TO BE ACCOUNTED FOR OR CODED AROUND##########
    "InvokeScript": [
      "
      Invoke-WinUtilExplorerUpdate
      if ($sync.ThemeButton.Content -eq [char]0xF08C) {
        Invoke-WinutilThemeChange -theme \"Auto\"
      }
      "
    ],
    "UndoScript": [
      "
      Invoke-WinUtilExplorerUpdate
      if ($sync.ThemeButton.Content -eq [char]0xF08C) {
        Invoke-WinutilThemeChange -theme \"Auto\"
      }
      "
    ],
    "link": "https://winutil.christitus.com/dev/tweaks/customize-preferences/darkmode"
  },
```

This would allow for other apps to easily use these tweaks without the PowerShell dependency.

ghost:
This is a better example
```powershell
  "WPFTweaksRemoveHome": {
    "Content": "Remove Home from Explorer",
    "Description": "Removes the Home from Explorer and sets This PC as default.",
    "category": "z__Advanced Tweaks - CAUTION",
    "panel": "1",
    "InvokeScript": [
      "
[...]

PR #4161, opened 2026-03-06
Update Bing Search Toggle in tweaks.json

cpb34:
## Type of Change
- [ ] New feature
- [X] Bug fix
- [X] Documentation update
- [ ] UI/UX improvement

## Description

Added an UndoScript in WPFToggleBingSearch that removes the DWORD CortanaConsent if it exists in HKCU:\Software\Microsoft\Windows\CurrentVersion\Search.

When CortanaConsent is set to 1 in Windows 11, Bing Search may be enabled even when the DWORD BingSearchEnabled is set to 0. CortanaConsent only exists when a Windows 10 PC consented to Cortana and later migrated to Windows 11; a fresh Windows 11 installation does not include CortanaConsent and the script otherwise works in Windows 10 even with CortanaConsent set to 1.

The images below were taken after refreshing Windows Explorer and the Registry Editor.

Without CortanaConsent removal (live internet build)
winutil without CortanaConsent removal
With CortanaConsent removal (local compiled build)
winutil with CortanaConsent removal

### DIFF
--- a/docs/content/dev/tweaks/Customize-Preferences/BingSearch.md
+++ b/docs/content/dev/tweaks/Customize-Preferences/BingSearch.md
@@ -20,6 +20,15 @@ description: ""
         "DefaultState": "true"
       }
     ],
+    "UndoScript": [
+      "
+      $SearchPath = Get-Item -Path 'HKCU:\\Software\\Microsoft\\Windows\\CurrentVersion\\Search'
+
+      If ($SearchPath.GetValue('CortanaConsent') -ne $null) {
+        Remove-ItemProperty -Path 'HKCU:\\Software\\Microsoft\\Windows\\CurrentVersion\\Search' -Name 'CortanaConsent'
+      }
+      "
+    ],
 ```
 
 ## Registry Changes


cpb34:
Apologies, I just read issue [#4155](https://github.com/ChrisTitusTech/winutil/issues/4155) after I submitted the pull request. I used a script since the end result removes the edge case while retaining what the registry structure should be; the standardized fix adds a DWORD that nearly every PC is perfectly fine without. If I had noticed the request for standardization, I would have submitted the following code.

This adds CortanaConsent if it doesn't exist and sets it as 0 or 1 depending on the toggle state. It has been tested the same as the code submitted in the pull request.

```
"registry": [
      {
        "Path": "HKCU:\\Software\\Microsoft\\Windows\\CurrentVersion\\Search",
        "Name": "BingSearchEnabled",
        "Value": "1",
        "Type": "DWord",
        "OriginalValue": "0",
        "DefaultState": "true"
      },
      {
        "Path": "HKCU:\\Software\\Microsoft\\Windows\\CurrentVersion\\Search",
        "Name": "CortanaConsent",
        "Value": "1",
        "Type": "DWord",
        "OriginalValue": "0",
        "DefaultState": "true"
      }
],
```

ghost:
You dont need to update the `BingSearch.md` file it well automaticly update, 

### Also the undo script wont run if there is no InvokeScript,

And there is no reason to use a if statment there because winutil wont report an error anyway

cpb34:
> You dont need to update the `BingSearch.md` file it well automaticly update,
> 
> ### Also the undo script wont run if there is no InvokeScript,
> And there is no reason to use a if statment there because winutil wont report an error anyway

Not true; the UndoScript is firing when toggled off and it is not fired when toggled on. Those screenshots are the results of the exact changes submitted.

About the errors, there was an error thrown in the CTT Powershell window while testing a different method regarding CortanaConsent not existing. Either way, not executing after an if statement is false is better in my opinion than silently failing. If you like the silent fail route more, the following script can be adapted for the winutil.

Remove-ItemProperty -Path "HKCU:\Software\Microsoft\Windows\CurrentVersion\Search" -Name "CortanaConsent" -ErrorAction SilentlyContinue

Thanks for letting me know about BingSearch.md automatically updating. Later today if this PR isn't closed, I'll come back and submit a new standardized fix with test screenshots.

ghost:
> > You dont need to update the `BingSearch.md` file it well automaticly update,
> > ### Also the undo script wont run if there is no InvokeScript,
> > And there is no reason to use a if statment there because winutil wont report an error anyway
> 
> Not true; the UndoScript is firing when toggled off and it is not fired when toggled on. Those screenshots are the results of the exact changes submitted.
> 
> About the errors, there was an error thrown in the CTT Powershell window while testing a different method regarding CortanaConsent not existing. Either way, not executing after an if statement is false is better in my opinion than silently failing. If you like the silent fail route more, the following script can be adapted for the winutil.
> 
> Remove-ItemProperty -Path "HKCU:\Software\Microsoft\Windows\CurrentVersion\Search" -Name "CortanaConsent" -ErrorAction SilentlyContinue
> 
> Thanks for letting me know about BingSearch.md automatically updating. Later today if this PR isn't closed, I'll come back and submit a new standardized fix with test screenshots.

okay but why add remove CortanaConsent when doing a undo?

jnsh-rf:
If I understood correctly, currently the tweak has an edge case when upgraded from Win10 to 11, which is:

- clean installs of Win11 would have only the reg key that the tweak in main branch changes - in that case, it works as intended;
- those migrated/"upgraded" from win10 may have reg key that this PR accounts for;
- Win11 works fine without it, but when detected, it either overrides it, treats it with higher priority, or have presedence over reg key in tweak, or something like that;
- when `BingSearchEnabled` is set to 1 toggle shows "enabled" and from a user perspective, it works as intended, but when this key is present and `BingSearchEnabled` is set to 0, the toggle will show "disabled", while it's not. If not present, both states of toggle works as intended;
- as lack of `CortanaConsent` does not have a negative effect on OS itself, it existing when  `BingSearchEnabled` = 1 does not matter, but when = 0, it must be removed for the toggle to report proper state of Bing search.

So, setting the Invoke is not needed. Undo simply can remove it, and when user re-enables Bing search, it will work as it would on clean install.

Did I get it right?

ghost:
> If I understood correctly, currently the tweak has an edge case when upgraded from Win10 to 11, which is:
> 
> * clean installs of Win11 would have only the reg key that the tweak in main branch changes - in that case, it works as intended;
> * those migrated/"upgraded" from win10 may have reg key that this PR accounts for;
> * Win11 works fine without it, but when detected, it either overrides it, treats it with higher priority, or have presedence over reg key in tweak, or something like that;
> * when `BingSearchEnabled` is set to 1 toggle shows "enabled" and from a user perspective, it works as intended, but when this key is present and `BingSearchEnabled` is set to 0, the toggle will show "disabled", while it's not. If not present, both states of toggle works as intended;
> * as lack of `CortanaConsent` does not have a negative effect on OS itself, it existing when  `BingSearchEnabled` = 1 does not matter, but when = 0, it must be removed for the toggle to report proper state of Bing search.
> 
> So, setting the Invoke is not needed. Undo simply can remove it, and when user re-enables Bing search, it will work as it would on clean install.
> 
> Did I get it right?

oh i just remebered this is a toggle so the undo makes sense lol

and yes this pr make sense

cpb34:
> If I understood correctly, currently the tweak has an edge case when upgraded from Win10 to 11, which is:
> 
> * clean installs of Win11 would have only the reg key that the tweak in main branch changes - in that case, it works as intended;
> * those migrated/"upgraded" from win10 may have reg key that this PR accounts for;
> * Win11 works fine without it, but when detected, it either overrides it, treats it with higher priority, or have presedence over reg key in tweak, or something like that;
> * when `BingSearchEnabled` is set to 1 toggle shows "enabled" and from a user perspective, it works as intended, but when this key is present and `BingSearchEnabled` is set to 0, the toggle will show "disabled", while it's not. If not present, both states of toggle works as intended;
> * as lack of `CortanaConsent` does not have a negative effect on OS itself, it existing when  `BingSearchEnabled` = 1 does not matter, but when = 0, it must be removed for the toggle to report proper state of Bing search.
> 
> So, setting the Invoke is not needed. Undo simply can remove it, and when user re-enables Bing search, it will work as it would on clean install.
> 
> Did I get it right?

All of that is correct other than the fourth bullet. It's while `CortanaConsent` is set to 1 where the Bing Search toggle is off but Bing Search is still enabled. `CortanaConsent` existing as 0 has the same effect as if `CortanaConsent` is not present at all. Once removed, the computer functions as a clean install would, exactly.

If the target behavior is that of a clean install, the PR achieves that. If it's more important for the code to be standardized, adding `CortanaConsent` as a toggle the same as `BingSearchEnabled`, as shown above, is the better route.

Thanks for verifying! It's a pretty common issue to see in the repair shop I work at; I just had to manually modify the Registry for a PC that came in 5 minutes ago too 😅

Also, thanks Chris for updating the offline EXE wrapper! It's awesome to see and I'll make sure to toss a few dollars toward the project when I get around to it.

ghost:
> I'm going to be revamping the way we remove registry values and essentially getting rid of all these independent invoke scripts in winutil. Stay tuned, I'm going to leave the PR open while I add tooling to make this type of thing easier.

I would love to see it, are you going to work on it in the next stream?

cpb34:
> > I'm going to be revamping the way we remove registry values and essentially getting rid of all these independent invoke scripts in winutil. Stay tuned, I'm going to leave the PR open while I add tooling to make this type of thing easier.
> 
> I would love to see it, are you going to work on it in the next stream?

+1. Excited to see what comes of this.

ChrisTitusTech:
Windows 10 is EOL and the only registry keys are from Win10 systems and Win10 system that were upgraded.

cpb34:
> Windows 10 is EOL and the only registry keys are from Win10 systems and Win10 system that were upgraded.

The commit fixes broken behavior only on Windows 11, not 10. I know W10 is EOL, but as it stands, winutil will be broken for anyone who clicked OK to installing Windows 11 from the Windows 10 update screen. That's a lot of people and I've seen it in this state dozens of times.

PR #4268, opened 2026-03-27
Make Auto Elevate auto close the non elevated powershell window

GabiNun2:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description

Closing the window after it self-elevates is more logical.

I also removed the `Write-Host` statement, since the user would not see the message anyway because the PowerShell window closes.
## Issue related to PR

- Resolves #


ghost:
it dosen't work

PR #4270, opened 2026-03-27
Remove the annoying install/uninstall pop out

GabiNun2:
## Description  
Removes the install/uninstall pop out that is unneeded

WinUtil Blocking Overlay

## Reason for Removal  
- The progress indication is frequently inaccurate and does not represent real task status.  
- It duplicates existing feedback already visible in the terminal window.  
- It negatively impacts usability by locking the interface without providing meaningful value.


ChrisTitusTech (changes requested):
Need something for fullscreen users that notifies them of what is happening. 
Most users do not check terminal output.

PR #4276, opened 2026-03-27
Remove not recommended from WPFTweaksDeBloat tweak

GabiNun2:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Documentation update
- [ ] UI/UX improvement

## Description

Windows 11 creator removes the same apps but dosen't have a huge tag labeling it as "NOT RECOMMENDED" so it dosen't make sense that this tweak dose

Also this tweak isn't unsafe there is really not problem with removing those garbage microslop pre-installed apps
## Issue related to PR

- Resolves #


seanh1995:
@GabiNun2 thanks yeah this is better. make more sense lf it does the same as the win11 creator. also seen a typo u done well instead off will.

seanh1995:
@GabiNun2 @ChrisTitusTech i was thinking ovs am no expert in this. but is there a way to change this to from removing apps to keep apps. because i have just reinstalled a windows vm and done this tweak and seen dev home, outlook and linkedin was left. so was thinking is there a way to say keep the important apps but remove everything else even in future when ms adds more bloat. if any off that makes sense

ghost:
> @GabiNun2 @ChrisTitusTech i was thinking ovs am no expert in this. but is there a way to change this to from removing apps to keep apps. because i have just reinstalled a windows vm and done this tweak and seen dev home, outlook and linkedin was left. so was thinking is there a way to say keep the important apps but remove everything else even in future when ms adds more bloat. if any off that makes sense

It's impossible for us to remove linkedin, pretty sure outlook isn't an appx package and devhome is a system package so we probaly shouldn't remove it

PR #4374, opened 2026-04-16
[Tweaks] New tweak for unsigned RDP files

### LINKED ISSUE #4390: RDP: New Security Warning disable

Please could we have an option under "Tweaks" to disable the new RDP Warning. 

**See this article for more information:** https://www.bleepingcomputer.com/news/microsoft/microsoft-adds-windows-protections-for-malicious-remote-desktop-files/

**The registry entry is:**
_Microsoft says that Administrators can temporarily disable these protections by going to the HKLM\Software\Policies\Microsoft\Windows NT\Terminal Services\Client Registry key and modifying the RedirectionWarningDialogVersion value so it is set to 1._

CodingWonders:
## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
This PR adds a new tweak to get rid of new warnings introduced by Microsoft when launching unsigned RDP files. Specifically, the following warnings:

These warnings were introduced with the latest cumulative updates for both Windows 10 and Windows 11. I learned about signed RDP files the hard way when I tried to remote into my Linux host.

I put it in the advanced tweaks because it modifies a security preference. Microsoft even confirms that these will stop working in a future update.

More info: https://www.bleepingcomputer.com/news/microsoft/microsoft-adds-windows-protections-for-malicious-remote-desktop-files/

## Issue related to PR
- Fixes #4390


CodingWonders:
Quick clarification of the future of such policies:

Microsoft did claim that support for such preferences would be removed in a future update:

> A future Windows update might remove support for this setting, even on older versions of Windows. Plan to transition your environment to work with the new security dialog.
>
> — *https://learn.microsoft.com/en-us/windows-server/remote/remote-desktop-services/remotepc/understanding-security-warnings#im-an-it-administrator-how-do-i-temporarily-revert-the-new-security-dialog*

However, that future update does not have a release date, so we don't know when these preferences will stop working. Maybe next month, maybe next year... even never, knowing Microsoft.

The ultimate solution would be to sign such RDP files. There's a tool for that: [rdpsign](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/rdpsign). I haven't used this tool myself, but I guess that you can use self-signed certificates and get away with that.

ChrisTitusTech (approved):

PR #4428, opened 2026-04-29
Removed anythingllm from applications.json

GabiNun2:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
who was the dumbass that added this and put both winget and choco as na, it was [gcampton](https://github.com/ChrisTitusTech/winutil/commits?author=gcampton)


ChrisTitusTech (approved):
I need to review and add Winget packages for a few other packages from his PR that weren't added. This might not be the only package that was missed.

GabiNun2:
> I need to review and add Winget packages for a few other packages from his PR that weren't added. This might not be the only package that was missed.

i actuly checked every single winget id in application.json which is how i found out about this so this is the only one

PR #4465, opened 2026-05-09
Cleanup programs install

GabiNun2:
Someone was on drugs while making this, you would think a microsoft employe coded this or grok 1


ChrisTitusTech (approved):
WOAH... that is hilarious. Gotta show this for live stream!

ChrisTitusTech:
Need to test this looks like it might cause some confusion on installs if we lose the GUI HUD.

ghost:
> Need to test this looks like it might cause some confusion on installs if we lose the GUI HUD.

the gui for installing apps still shows up

ChrisTitusTech (approved):

PR #4484, opened 2026-05-13
simplify Compile.ps1 by alot

ghost:
## Changes:
- Removed removing `winutil.ps1` at the start because it's unneeded `Set-Content` overrides it
- Removed progress output since **execution is now almost instant**
- Removed `$Arguments` because `winutil.ps1` has arguments not `Complie.ps1`
- Removed preprocessing
- Removed syntax check for `winutil.ps1` since errors are already shown during execution
- Removed `Set-ExecutionPolicy` for `-Run` argument because if you can run complie.ps1 you dont need to change `Set-ExecutionPolicy`


ChrisTitusTech (approved):
The initial idea for preprocessing was that it will cleanup and standardize the code (like removing spaces at end of lines and capitalization for common functions). 

In practice, I do find it more annoying than helpful. I think removing it is the right call, but lets do this on stream as I'd like more feedback.

ghost:
> The initial idea for preprocessing was that it will cleanup and standardize the code (like removing spaces at end of lines and capitalization for common functions).
> 
> In practice, I do find it more annoying than helpful. I think removing it is the right call, but lets do this on stream as I'd like more feedback.

you shouldn't write bad code anyways :)

PR #4496, opened 2026-05-15
Cleanup Get-WinUtilSelectedPackages.ps1 and remove annoying messages

ghost:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description

This was to remove the "Queing" messages but might aswell clean this up while i'm at it
i also removed wingets if "na" statment because there isn't a single package in winutil that dosen't have winget id nor would any be added
## Issue related to PR

- Resolves #

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -46,7 +46,7 @@ jobs:
             config/feature.json
           labels: |
             automated
-            documentation
+            skip-changelog
 
       - name: Check outputs
         shell: bash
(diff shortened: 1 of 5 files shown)


ChrisTitusTech (changes requested):
This would cause certain packages to fail that don't have both winget and choco filled out. The current method would install a choco package with winget selected if it only has choco packages and vice versa.

ghost:
> This would cause certain packages to fail that don't have both winget and choco filled out. The current method would install a choco package with winget selected if it only has choco packages and vice versa.

fixed

ChrisTitusTech (approved):

PR #4505, opened 2026-05-16
feature/New config Sysinternals-Live map with local save option + PATH 

gepardjaro:
## Type of Change
- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
Added Systinternals-Live mapping to S:\ (User). Added SSL and popup for local copy option with added PATH to use tools directly from it. Added temp net use for elevated to robocopy and changed FileSizeLimitInBytes to 100mb to allow copy (default 50mb) and bring it back to the original state afterwards.

 - Maps \\live.sysinternals.com@SSL\tools as drive S: for the interactive user (not the
   elevated context), so the drive shows up in the Explorer
  - Optionally (via a Yes/No popup) copy all Sysinternals to C:\Program Files\Sysinternals and appends that path to the System PATH, making the tools callable for both User/elevated shell, not depending on the network share

### DIFF
--- a/config/feature.json
+++ b/config/feature.json
@@ -96,6 +96,62 @@
     ],
     "link": "https://winutil.christitus.com/dev/features/features/regbackup"
   },
+  "WPFFeatureSysinternalsLiveEnable": {
+    "Content": "Sysinternals Live - Map S: and optional local copy + to PATH",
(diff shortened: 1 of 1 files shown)


ghost:
to niche and the code is horrible

ChrisTitusTech:
We already have a few sysinternals tools in install tab, but most of this should be done outside winutil as they are very niche for some.

PR #4507, opened 2026-05-16
Minimize disable/enable updates

ghost:
Last time i did this you said "windows will re renable updates if you only disable UsoSvc service" but i disable updates by only disabling `UsoSvc` service and windows hasn't re enabled updates or that service which proves you are wrong

also the defaults update shouldn't try to fix updates it should only undo disable updates because we already have a fix updates button in config tab

This also means disable/enable updates dont require a restart

### DIFF
--- a/functions/public/Invoke-WPFUpdatesdefault.ps1
+++ b/functions/public/Invoke-WPFUpdatesdefault.ps1
@@ -1,59 +1,17 @@
 function Invoke-WPFUpdatesdefault {
-    <#
-
-    .SYNOPSIS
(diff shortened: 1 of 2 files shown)


ChrisTitusTech:
any windows update repair would fix this or installing a ms tool will re-enable updates. This won't work.

ghost:
> any windows update repair would fix this or installing a ms tool will re-enable updates. This won't work.

there is no windows update repair because we disabled updates and what are you even talking about? so your saying install studio will re enable updates?

PR #4530, opened 2026-05-20
simplify admin check

ghost:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
This does means if you have `pwsh` without `wt` than it will use `powershell` but who has `pwsh` without `wt`


Marterich:
Is this change really necessary? 
All this PR does is removing 2 features whilst saving _only_ 5 lines doing so. 

### Argument Handling
Previously, the $argList logic automatically enumerated all arguments and passed them on, so if a new argument was added to the script, you did not have to touch the logic to keep the script working as expected.

Now one would have to add the new arguments in multiple places. (at the start of the script, and also inside the new logic block)

### Powershell Start 
On Windows 10, wt.exe was not preinstalled. This only started with Win11, so when one installed Powershell 7, there was no guarantee that the Windows Terminal is also installed.
As long as the Script claims to be compatible with Win10 as well, i see no point in removing the check.

ghost:
> Is this change really necessary? All this PR does is removing 2 features whilst saving _only_ 5 lines doing so.
> 
> ### Argument Handling
> Previously, the $argList logic automatically enumerated all arguments and passed them on, so if a new argument was added to the script, you did not have to touch the logic to keep the script working as expected.
> 
> Now one would have to add the new arguments in multiple places. (at the start of the script, and also inside the new logic block)
> 
> ### Powershell Start
> On Windows 10, wt.exe was not preinstalled. This only started with Win11, so when one installed Powershell 7, there was no guarantee that the Windows Terminal is also installed. As long as the Script claims to be compatible with Win10 as well, i see no point in removing the check.

you can install wt in windows 10 and the script does not claim it is compatible with Win10 it fact it says says otherwise and i do see how you it dosen't really make sense to not get the arguments recursively so i will close this or

PR #4531, opened 2026-05-21
Removed -NoUi

ghost:
`-NoUi` was only used alongside `-Config`, so there’s need to specify both flags. Using `-Config` now automatically disables the UI.

I also removed the `-Config` + UI option, since the same behavior is already available through the Import button in the top-right corner.


gepardjaro:
nice one, I like it

ghost:
> nice one, I like it

thank you

PR #4536, opened 2026-05-22
Made Compile.ps1 not load the script into memory and use -Raw

ghost:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
pr #2062 made complie.ps1 add winutil into memory instead of constantly using `| Out-File` which made sense but it's simpler and more strightforward to use `$script +=`

Using `-Raw` makes this about 6 times faster from 120 mileseconds to 20 mileseconds so the "performance lost" from not using `Collections.Generic.List` dosen't matter
### ofc i tested this :)


Marterich:
Whilst looking a bit simpler, switching out the Generic List Class for the simple "+=" could have quite some performance implications, so i would not recommend merging this.

Native PowerShell Objects (Strings, Arrays etc) are often immutable by design, so calling += to add new data to them duplicates the object in memory, appends the new data, stores it as a new object and then drops the previous object. This is quite a slow and memory intensive process. 

In the newest PowerShell Versions the impact seems to be a bit reduced, but as Windows dosn't ship these versions by default, i still would keep the old logic for this. 

You can learn more about PowerShell performance quirks at:
[learn.microsoft.com/en-us/powershell/scripting/dev-cross-plat/performance/script-authoring-considerations?view=powershell-7.6](https://learn.microsoft.com/en-us/powershell/scripting/dev-cross-plat/performance/script-authoring-considerations?view=powershell-7.6)

ghost:
> Whilst looking a bit simpler, switching out the Generic List Class for the simple "+=" could have quite some performance implications, so i would not recommend merging this.
> 
> Native PowerShell Objects (Strings, Arrays etc) are often immutable by design, so calling += to add new data to them duplicates the object in memory, appends the new data, stores it as a new object and then drops the previous object. This is quite a slow and memory intensive process.
> 
> In the newest PowerShell Versions the impact seems to be a bit reduced, but as Windows dosn't ship these versions by default, i still would keep the old logic for this.
> 
> You can learn more about PowerShell performance quirks at: [learn.microsoft.com/en-us/powershell/scripting/dev-cross-plat/performance/script-authoring-considerations?view=powershell-7.6](https://learn.microsoft.com/en-us/powershell/scripting/dev-cross-plat/performance/script-authoring-considerations?view=powershell-7.6)

i just tested this in pwsh and powershell and it's about 30 mileseconds slower while having way better and more readable syntax

Marterich:
Fair
As the compile script is only relevant for devs and the Github action anyways, this is definitely no major issue, but the performance implications might vary quite a bit depending on the machine.

But we should consider if we want compromise on performance (even if the impact in this case is relatively little) when the only change needed in the code is a single line to setup the generic list and running $var.add(<something>) instead of $var += "something".

The reduction in lines in your PR mostly stems from removing comments and line breaks.

ghost:
> Fair As the compile script is only relevant for devs and the Github action anyways, this is definitely no major issue, but the performance implications might vary quite a bit depending on the machine.
> 
> But we should consider if we want compromise on performance (even if the impact in this case is relatively little) when the only change needed in the code is a single line to setup the generic list and running $var.add() instead of $var += "something".
> 
> The reduction in lines in your PR mostly stems from removing comments and line breaks.

it's not about removing lines it's about making the code more understandable and i wouldn't want to compromise the performance of winutil but the complie.ps1 is a diffrent thing and not to long ago i made the complie.ps1 script go from taking a few seconds to 100 milesconds so i think it's fine to compromise the performance by so little

ChrisTitusTech (approved):

Issue #4534, opened 2026-05-22
Error when attempting to run winutil


luckehal:
### I have read the known issues

- [x] Yes, I did

### What part of Winutil are you having issues with?

None

### Provide a clear and concise description of the issue.

See error message

"PS C:\WINDOWS\system32> irm christitus.com/win | iex
< : The term '<' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ <html>
+ ~
    + CategoryInfo          : ObjectNotFound: (<:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException"

<img width="941" height="495" alt="Image" src="https://github.com/user-attachments/assets/4301952c-1ab6-4c35-914c-6a46afc9c535" />

ghost:
https://winutil.christitus.com/knownissues/

/close

Issue #4444, opened 2026-05-03
Cant Execute


PugInARug121:
### What part of Winutil are you having issues with?

None

### Provide a clear and concise description of the issue.

Im trying to run Winutils and I get the an error code!

### Paste the full error output (if available) or Screenshot or Video.

< : The term '<' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ <html>
+ ~
    + CategoryInfo          : ObjectNotFound: (<:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

< : The term '<' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:2 char:1
+ <head>
+ ~
    + CategoryInfo          : ObjectNotFound: (<:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

< : The term '<' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:3 char:1
+ <meta HTTP-EQUIV="REFRESH" content="0; url=http://blocking.asus.hns.t ...
+ ~
    + CategoryInfo          : ObjectNotFound: (<:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

< : The term '<' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:4 char:1
+ </head>
+ ~
    + CategoryInfo          : ObjectNotFound: (<:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

< : The term '<' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:5 char:1
+ <body></body>
+ ~
    + CategoryInfo          : ObjectNotFound: (<:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

< : The term '<' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:6 char:1
+ </html>
+ ~
[...]

PR #4537, opened 2026-05-22
Update KnownIssues.md

ghost:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Documentation update
- [ ] UI/UX improvement

## Description
changing your dns wont do anything if your isp is blocking winutil
hopefully makes it so issues like #4534 and #4444 wont get created it the first place

also removed `### Ultimate Performance Plan Not Working`  because now it says ` - NOT FOR LAPTOPS`


ChrisTitusTech (changes requested):
We are still having the india issue and sometimes switching from ISP DNS to google / cloudflare does resolve the issue

ghost:
> We are still having the india issue and sometimes switching from ISP DNS to google / cloudflare does resolve the issue

i always ask people to switch there dns when they run into this problem and it never helps also we don't use a raw github link for winutil so it should work in india

ChrisTitusTech (approved):

PR #4545, opened 2026-05-24
Removed Invoke-WinUtilExplorerUpdate.ps1 function

ghost:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
Invoke-WinUtilExplorerUpdate.ps1 was just bloat this also greatly helps winoneshot so it can now actually restart explorer whe needed


Marterich:
Previously the old Invoke-WinUtilExplorerUpdate logic implemented both, a hard restart and a soft refresh. 

### Explorer Restart
The **hard restart** makes sure to end the explorer process forcefully. If memory serves correctly, this was done to overcome some problems where onedrive would keep the explorer from  being restarted or stopped normally.
Afterwards the logic explicitly restarts the explorer.

### Explorer Refresh
The **soft refresh** was used to tell the explorer "hey, please recheck the registry if some configurations changed"
This was mainly used to update the explorer to apply a theme change that was invoked from winutil without needing a full restart. This has the advantage of all explorer windows staying open and the theme change being applied nearly instantaneously. 


After this PR a Theme change (via tweaks) ALWAYS requires a Explorer Restart (via Stop-Process), and we _hope_ that windows correctly restarts it automatically.  Also this could lead to the ondrive weirdness again i alluded to earlier

ghost:
> Previously the old Invoke-WinUtilExplorerUpdate logic implemented both, a hard restart and a soft refresh.
> 
> ### Explorer Restart
> The **hard restart** makes sure to end the explorer process forcefully. If memory serves correctly, this was done to overcome some problems where onedrive would keep the explorer from being restarted or stopped normally. Afterwards the logic explicitly restarts the explorer.
> 
> ### Explorer Refresh
> The **soft refresh** was used to tell the explorer "hey, please recheck the registry if some configurations changed" This was mainly used to update the explorer to apply a theme change that was invoked from winutil without needing a full restart. This has the advantage of all explorer windows staying open and the theme change being applied nearly instantaneously.
> 
> After this PR a Theme change (via tweaks) ALWAYS requires a Explorer Restart (via Stop-Process), and we _hope_ that windows correctly restarts it automatically. Also this could lead to the ondrive weirdness again i alluded to earlier

first this won't cause problems with onedrive second it's really not a problem to restart explorer inorder to apply a theme change we restart explorer for litarly everything else exepect the theme change so having a function just for 1 toggle wouldn't really make sense and i think this makes things more standard and it's probaly safer to just restart explorer also if your someone like me that dosen't understand low level windows calls this will help us understand the script better this will basicly just run c# code which seems kinda counterintuitive in a powershell script

PR #4546, opened 2026-05-24
Fixed irfanview's and openvpn's foss selection

ghost:


ChrisTitusTech (approved):

ChrisTitusTech:
I regret adding foss flag... maintaining this just sucks. We should remove it

ghost:
> I regret adding foss flag... maintaining this just sucks. We should remove it

nah i mean that's the only fix needed after this we don't really need to maintane it and it makes the program list look cool

ChrisTitusTech (approved):

PR #4571, opened 2026-05-27
Removed -offline

ghost:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
Now it will simplify active upon not having an active internet connection instead of needing to use -offline

it should ping a domain (aka google.com) instead of an ip to ensure dns works fine


ChrisTitusTech:
Will break exes

PR #4601, opened 2026-05-29
Removed Set-WinUtilService

ghost:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
It's just bloat at this point
nothing else was using `service` this is everything


ChrisTitusTech:
I'd like to keep these commands because if I do have a C# program having them already in JSON is easier for translation than feeding powershell.

ghost:
> I'd like to keep these commands because if I do have a C# program having them already in JSON is easier for translation than feeding powershell.

sure but this is just bloat and it's not like it's better to disable a service in c# rather than set-service

PR #4602, opened 2026-05-29
Moved inputXML.xaml down a dir

ghost:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
No reason for it not to be in the root
you should probaly merge this after all the other prs because it might create a conflict


ChrisTitusTech:
Eww I don't like bloating up the root. I like it in its own folder, just incase we do add more xaml in the future.

PR #4606, opened 2026-05-31
Deleted Test-WinUtilPackageManager.ps1 + custom winget and choco installs + remove winget/choco enums and exeptions

ghost:
## Type of Change
- [x] Cleanup
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
Using our own script (that i made) makes it install choco or winget alot faster
Both winget and choco installs have been tested on `Windows 11 IoT Enterprise LTSC`

I don't know why we even did it this way. (winget/choco enums)

i also removed the "classes for different exceptions" because we never used them


ChrisTitusTech (commented):
Ugh, I gotta test winget / choco installs again. I remember going back and forth on these install methods.

ChrisTitusTech (approved):

ChrisTitusTech (changes requested):
Thanks for working through the package-manager cleanup, but I do not want to replace the current WinGet installation path here. Please keep the existing `Install-WinUtilWinget` flow that uses `Microsoft.WinGet.Client` / `Repair-WinGetPackageManager -AllUsers`.

A few blockers from review:

- Removing the `PackageManagers` enum is incomplete. `Get-WinUtilPackageLogSummary.ps1` still has `[PackageManagers]$Preference` and compares against `[PackageManagers]::Choco`, so install/uninstall now fail before the runspace work when the summary is generated. I reproduced the PR head behavior with: `Unable to find type [PackageManagers]`.
- The focused checks do not pass on the PR head. `Compile.ps1` completed, but `Invoke-Pester -Path @('pester/package.Tests.ps1','pester/install-workflow.Tests.ps1','pester/preferences-theme.Tests.ps1') -Output Detailed` reported `15 passed, 16 failed`; `package.Tests.ps1` still dot-sources the deleted `functions/private/Test-WinUtilPackageManager.ps1`, and the install workflow tests still expect the previous package-manager split contract.
- The new direct WinGet installer writes `winget.msixbundle`, `winget.zip`, and `winget\...` into whatever the current working directory is, then deletes `winget*` recursively. That is too risky for WinUtil because the launch directory is not guaranteed to be a scratch directory.
- The new Chocolatey path also bypasses Chocolatey's installer script and manually moves package contents / edits PATH, which risks leaving a partial install without the normal Chocolatey environment setup.

Given the above, I think this PR should leave the existing WinGet method in place and either keep the current package-manager helpers/tests or split the broader package-manager cleanup into a separate PR with matching runtime changes and tests.

ChrisTitusTech (approved):
Verified the current head after the manual conflict resolution and follow-up fix.

Checks performed:
- GitHub reports the PR as mergeable against `main`.
- GitHub `Compile-and-Check` is passing.
- Locally verified GitHub's synthetic merge ref `80ae1dc5d280d4f970d53fde3ee6d5caa3094016`.
- `Compile.ps1` succeeded on the merge ref.
- Full Pester suite passed on the merge ref: 403 passed, 0 failed.
- Confirmed no remaining `PackageManagers` references in source/tests/docs.
- `git diff --check` passed.

Looks ready to merge.

PR #4643, opened 2026-06-05
Fix null category error

ghost:
## Type of Change
 - [x] Bug fix
 
 ## Description
 Fixes a null reference error that occurs when loading data from `lastrun.json`. The issue happens because entries from `lastrun.json` don't have a `Category` field, causing `ContainsKey()` and hashtable indexing operations to fail with `MethodInvocationException` and `InvalidOperation` errors.
 
 The fix adds a default value of `"Uncategorized"` for the Category field when it's missing or null, consistent with how the Panel field already defaults to `"0"`. This ensures all entries have valid keys for hashtable operations.


ghost:
i have made my very own ai slop pr now :)

PR #4660, opened 2026-06-08
Fix devdocs generator mis-placing link / failing on inconsistent indentation + pester update to railguard this

gepardjaro:
## Type of Change
- [ ] New feature
- [X] Bug fix
- [X] Documentation update
- [ ] UI/UX improvement

## Description

Last time I made a mistake pasting link in the json file while adding Scrollbars feature in https://github.com/ChrisTitusTech/winutil/pull/4402. Then I raised PR to fix this https://github.com/ChrisTitusTech/winutil/pull/4528. It was fine until doc generator ran and regressed the fix.

As it was taking bunch of time to fix this, I asked my friend Claude to get a solution, I didn't have a time yet to go through all the code yet (spoke with ANGRYxScotsman to raise a PR already so we can both check it), but testing proved it's working now and it's guarded with additional script that checks on that + pester is updated to cover failures like this in the future, when somebody is making a mistake within the JSON structure

**tl;dr:**

The generator located structure by matching indentation, which breaks for any item whose key is indented differently from its siblings (today: WPFToggleScrollbars, indented 4 spaces instead of 2). Because the docs/links are machine-generated, the hand-fix in #4528 was wiped by the next generation (chore re-run in #4619), so the defect keeps returning. Two functions were affected, both hardened here.

Add-LinkAttributeToJson located each item link via propIndent = itemKeyIndent + 2, so a deeper-indented key made propIndent coincide with the nested registry object indent: the updater latched onto a stray link inside the registry block (and inserts targeted the registry brace, leaving a dangling comma). Rewritten to find each item by its unique name and take its own closing brace as the last brace before the next item start (root brace for the last item), then update/insert the top-level link there.

Get-RawJsonBlock found an item closing brace by matching a brace at the same indent as the key, so a mis-indented key produced a 'Could not find closing brace' warning and emitted no JSON doc block. Rewritten to match the closing brace by string-aware brace counting that ignores braces inside quoted values.

Also adds a run-guard so the script loads its functions without running the generation pipeline when dot-sourced, and pester/devdocs-generator.Tests.ps1 with regression tests.

**Detailed changes:**
`tools/devdocs-generator.ps1` located JSON structure by indentation, which breaks for any item whose key is indented differently from its siblings (today: `WPFToggleScrollbars`, indented 4 spaces). Because the docs/links are machine-generated, hand-fixing the JSON (as in #4528) doesn't stick — the next generation (e.g. the chore re-run in #4619) reintroduces the defect. This PR fixes the **generator** so placement is always correct, making the fix permanent. Generator-only change; no behavior change to WinUtil itself.

## Details

Two functions were fooled by the mis-indented key, both hardened here:

- **`Add-LinkAttributeToJson`** placed each item's `"link"` using `propIndent = itemKeyIndent + 2`, so a deeper-indented key made `propIndent` coincide with the nested `registry` object's indent — the updater latched onto a stray link inside the registry block (and inserts targeted the registry brace, leaving a dangling comma). Rewritten to find each item by its **unique name** and take its own closing brace as the last `}` before the **next item's start** (root brace for the last item), then update/insert the top-level `"link"` there.
- **`Get-RawJsonBlock`** found an item's closing brace by matching `}` at the **same indent as the key**, so a mis-indented key produced a `Could not find closing brace` warning and emitted no JSON doc block. Rewritten to match the closing brace by **string-aware brace counting** (ignoring braces inside quoted values).

Both approaches are immune to inconsistent indentation **and** to braces inside multi-line `InvokeScript`/`UndoScript` string values.

Also adds a run-guard so the script loads its functions without running the generation pipeline when dot-sourced, and `pester/devdocs-generator.Tests.ps1` covering: no link inside registry blocks, in-place update of a mis-indented entry, brace-in-string safety, insert-when-missing, idempotency, and `Get-RawJsonBlock` brace finding.

## Verification

- `Invoke-Pester -Path 'pester/*.Tests.ps1'` → 12 passed (4 existing + 6 new + 2 for Get-RawJsonBlock).
- Running the generator directly no longer warns and renders the previously-missing Scrollbars doc block.
- `git diff config/tweaks.json` shows registry blocks untouched (only top-level links may change).

> Note: the one-time cleanup of the already-corrupted `WPFToggleScrollbars` entry in `config/tweaks.json` (stray inner `link` + dangling comma) is handled separately.

## Issue related to PR

- Resolves #

### DIFF
--- a/pester/devdocs-generator.Tests.ps1
+++ b/pester/devdocs-generator.Tests.ps1
@@ -1,0 +1,131 @@
+#===========================================================================
+# Tests - devdocs-generator (Add-LinkAttributeToJson)
+#===========================================================================
+# Regression test for the "link" placement bug (re-fixes #4528 durably): the
+# generator must always update/insert each item's top-level "link" and never
+# place one inside a nested registry block, regardless of indentation or of
+# braces appearing inside string values.
+Describe "devdocs-generator Add-LinkAttributeToJson" {
+    BeforeAll {
+        # Dot-sourcing loads the functions only; the script's run-guard skips the
+        # generation pipeline so this stays a fast, side-effect-free unit test.
+        . "$PSScriptRoot/../tools/devdocs-generator.ps1"
+
+        $script:cut    = 'WPF(WinUtil|Toggle|Features?|Tweaks?|Panel|Fix(es)?)?'
+        $script:prefix = 'https://example/dev/tweaks'
+
+        # Fixture covering the three tricky cases:
+        #  - WPFToggleScrollbars : key indented deeper than its siblings (the real bug),
+        #                          plus an existing top-level link that must be UPDATED.
+        #  - WPFTweakBraces      : has { and } INSIDE a multi-line string value, which must
+        #                          not be mistaken for structural braces.
+        #  - WPFTweakNoLink      : has NO link yet and must get one INSERTED.
+        $script:fixture = @'
+{
+    "WPFToggleScrollbars": {
+    "Content": "Scrollbars",
+    "category": "Customize Preferences",
+    "registry": [
+      {
+        "Path": "HKCU:\\Control Panel",
+        "DefaultState": "false"
+      }
+    ],
+    "link": "https://OLD/x"
+  },
+  "WPFTweakBraces": {
+    "Content": "Braces",
+    "category": "Essential Tweaks",
+    "InvokeScript": [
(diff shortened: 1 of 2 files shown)


seanh1995:
so this broke some links lol. this is one example.

image

gepardjaro:
> so this broke some links lol. this is one example.
> 
> image

let's see if we can get something from it later on this week, Claude must have been drunk

PR #4682, opened 2026-06-09
Fix font scaling slider not working on touchscreens

Dhananjay-JSR:
## What

Makes the Font Scaling slider usable on touchscreen devices.

## Why

WPF `Slider` thumbs only respond to mouse drag. On touch-only devices (handhelds, tablets, 2-in-1s) on my Asus Rog ally x it doesn't  work , touch isn't promoted to the mouse input the thumb relies on, so the slider can't be moved by finger at all.

## How

- Added `Set-WinUtilSliderFromTouch`, which maps a horizontal touch position onto the slider's range and snaps to the nearest tick.
- Wired the slider's `PreviewTouchDown/Move/Up` events to capture the touch and drive the value as the finger moves.
- Set `IsMoveToPointEnabled` and disabled the press-and-hold stylus gesture so tapping the track jumps to that point without a long-press delay.

## Testing

Compiled with `.\Compile.ps1 -Run` and confirmed on a touchscreen handheld: the slider now drags and snaps by touch. Mouse and keyboard behavior is unchanged.

### DIFF
--- a/functions/private/Set-WinUtilSliderFromTouch.ps1
+++ b/functions/private/Set-WinUtilSliderFromTouch.ps1
@@ -1,0 +1,22 @@
+function Set-WinUtilSliderFromTouch {
+    <#
+    .SYNOPSIS
+        Moves a Slider's value to a horizontal touch position. WPF Sliders ignore touch drag on touch-only devices, so this is wired up to the touch events manually.
+    #>
+    param (
+        [Parameter(Mandatory)] $Slider,
+        [Parameter(Mandatory)] [double]$PositionX
+    )
+
+    if ($Slider.ActualWidth -le 0) { return }
+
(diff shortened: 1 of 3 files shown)


vyas-devgna:
which device do you have is it a convertible or a tablet

vyas-devgna:
Could you test a simpler XAML-only approach? WPF sliders usually ignore touch because the parent ScrollViewer intercepts them for panning. Instead of manually capturing touch events and calculating the slider value in PowerShell, you can drop the PowerShell code entirely and just add ScrollViewer.PanningMode="VerticalOnly" to the <Slider> element in inputXML.xaml (alongside your other additions). This stops the ScrollViewer from stealing touch events, allowing the slider's native touch behavior to work automatically

PR #4714, opened 2026-06-15
Deleted lastrun.json

### LINKED ISSUE #4705: lastrun.json false positive
when ever i open winutil (windev) and run the `disable location tracking tweak` than close winutil and open it up again it will say

Image

which is wrong because i just ran the tweak

this happens with most tweaks

ghost:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description
to understand the issue read https://github.com/ChrisTitusTech/winutil/issues/4705

1, we can never truly detact a that a tweak was reverted because of invokescript
2, this has alot of false positives which we can't really fix
3, minor changes for `WPFTweaksRightClickMenu`

## Issue related to PR
- Resolves #4705

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -1104,9 +1104,7 @@
       "
     ],
     "UndoScript": [
-      "
-      Remove-Item -Path \"HKCU:\\Software\\Classes\\CLSID\\{86ca1aa0-34aa-4e8b-a509-50c905bae2a2}\" -Recurse
-      "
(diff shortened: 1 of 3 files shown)


ChrisTitusTech (changes requested):
why did you push registry to invoke script?

ChrisTitusTech (approved):
Perfect

ghost:
> why did you push registry to invoke script?

that was only for lastrun.json also it makes the undo not work if we do registry

PR #4722, opened 2026-06-18
Fix disable updates getting reverted

ghost:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [x] Refactor
- [ ] UI/UX improvement

## Description
i removes some dupe messages

i removed the registry changes because they truly do nothing there is no disable key if you were to apply those registry changes without disabling services than nothing would happen updates would still work

i removed the disabling of BITS because i don't see a reason to disable it since windows updates works fine without it and it's not related to windows updates

i removed the start-service from UsoSvc because it will auto start if you go to windows update + it was getting ran before the enable UsoSvc which meant it never did anything

removed "Hiding Windows Updates from settings" because if you ran disable windows ai tweak alongside disabling updates than it would break it or the other way around

i remove the disabling of tasks because only 1 task would get disabled due to permission issues and it really does nothing to disable those windows updates still work fine

i added a file renaming to usosvc.dlle from usosvc.dll so windows update will fail to repair now windows updates might be able to recreate the file which is why this pr is a draft there is currently a person on discord tester wheater renaming the dll will permantly disables updates or not


ChrisTitusTech (changes requested):
Leave the set service and task schedules.

ghost:
> Leave the set service and task schedules.

i added back disabling services but adding back task schedules makes no sense because only 1 of them got disabled due to accses denied and it dosen't do anything

PR #4723, opened 2026-06-18
Delete functions/private/Invoke-WinUtilExplorerUpdate.ps1

ghost:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactor
- [ ] UI/UX improvement

## Description
Replacing it with `Stop-Process -Name explorer` makes more sense and allows win1shot to apply invokescript correctly without needing to set a function


ChrisTitusTech (changes requested):
This would need to start the explorer process after it is killed though.

ghost:
> This would need to start the explorer process after it is killed though.

stop-process on explorer will restart it not shut it down using taskkill.exe will stop explorer

PR #4744, opened 2026-06-23
Update preset configuration

seanh1995:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description

This pull request updates the `config/preset.json` file to ensure that the `WPFTweaksRestorePoint` tweak is included at the beginning of both the `Minimal` and `Advanced` preset groups. This change helps guarantee that a restore point is created before any other tweaks are applied, improving system safety and rollback capability.

Preset configuration improvements:

* Added `WPFTweaksRestorePoint` as the first entry in both the `Minimal` and `Advanced` preset groups in `config/preset.json`, ensuring a restore point is created prior to applying other tweaks.

## Issue related to PR

- Resolves #

### DIFF
--- a/config/preset.json
+++ b/config/preset.json
@@ -15,7 +15,6 @@
     "WPFTweaksRestorePoint"
   ],
   "Minimal": [
-    "WPFTweaksRestorePoint",
     "WPFTweaksConsumerFeatures",
     "WPFTweaksDeBloat",
     "WPFTweaksWPBT",
--- a/config/preset.json
+++ b/config/preset.json
@@ -15,6 +15,7 @@
     "WPFTweaksRestorePoint"
   ],
   "Minimal": [
+    "WPFTweaksRestorePoint",
     "WPFTweaksConsumerFeatures",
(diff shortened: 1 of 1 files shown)


ghost:
minimal shouldn't have restore point

seanh1995:
@GabiNun2 they all should so if something goes wrong then can easy revert back. better having one than not having one

ghost:
> @GabiNun2 they all should so if something goes wrong then can easy revert back. better having one than not having one

people who run minimal preset don't want a restore point nor those it do that much that it might need to get reverted also it's alot easier to test something like -preset if it dosen't create a restore point everytime

seanh1995:
@GabiNun2 ok i removed it from the minimal but i am defo keeping it in the advance

ghost:
> @GabiNun2 ok i removed it from the minimal but i am defo keeping it in the advance

sure that sounds good

ChrisTitusTech (approved):

PR #4768, opened 2026-06-28
Fix Multiplane Overlay toggle

### LINKED ISSUE #4755: MPO
### I have read the known issues

- [x] Yes, I did

### What part of WinUtil are you having issues with?

Tweaks Tab

### Provide a clear and concise description of the issue.

With MPO enabled by default with the registry values not present, the toggle is showing MPO is Disabled. When I toggle MPO to Enabled, it causes some weird bug with GSync where it breaks GSync and runs monitor at max refresh rate at all times. When I toggle MPO to Disabled, GSync starts working again but without MPO enabled. If I want to go back to MPO Enabled without breaking GSync, I have to manually go into the registry and delete the two values that the MPO toggle creates. I noticed the issue because my game felt really stutter-ish which led me to benchmark with superposition. My benchmark was abnormally low. I turned on my monitors refresh rate overlay and noticed it was pinned to 360hz. I disabled MPO using the toggle and GSync started working again, the refresh rate was dynamically downclocking to match the benchmarks fps. I tried enabling MPO using the toggle and GSync stopped working again. I deleted the registry values and rebooted. Winutil showed MPO as being Disabled even though the registry values were removed which is the default for MPO enabled. I then used DDU to completely uninstall and then i reinstalled NVIDIA drivers. Checked Winutil and sure enough, it's still showing MPO is Disabled even though I know it's Enabled because the two corresponding registry tweaks are not present. 

1. There seems to be a difference in how GSync is working when I Enable MPO using the toggle (breaks GSync) vs when I enable MPO by manually deleting the two registry tweaks that disable MPO. 

2. When MPO is Enabled manually by deleting the registry tweaks, the MPO toggle in Winutil is showing that MPO is Disabled. 

### List the tweaks you applied before the issue occurred.

MPO toggle

### Paste the full error output (if available) or Screenshot or Video.

_No response_

LKmXr:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description

HKLM:\SYSTEM\CurrentControlSet\Control\GraphicsDrivers\DisableOverlays should be 0 when enabling MPO
image

## Issue related to PR

- Resolves #4755

### DIFF
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -1512,9 +1512,9 @@
       {
         "Path": "HKLM:\\SYSTEM\\CurrentControlSet\\Control\\GraphicsDrivers",
         "Name": "DisableOverlays",
-        "Value": "1",
+        "Value": "0",
         "Type": "DWord",
-        "OriginalValue": "0",
+        "OriginalValue": "1",
         "DefaultState": "false"
       }
     ],
--- a/config/tweaks.json
+++ b/config/tweaks.json
@@ -1515,7 +1515,7 @@
         "Value": "0",
         "Type": "DWord",
         "OriginalValue": "1",
-        "DefaultState": "false"
+        "DefaultState": "true"
       }
     ],
     "link": "https://winutil.christitus.com/dev/tweaks/customize-preferences/multiplaneoverlay"


ChrisTitusTech (approved):
Approved after code review and external validation of the MPO registry value polarity. The change correctly makes `DisableOverlays = 0` the applied/enabled state and keeps the missing/default registry state treated as enabled.

PR #4773, opened 2026-06-29
Add other Discord client variants and Legcord

usrx215:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description

This PR adds the other variants of the Discord desktop client as well as the Legcord client  
The other variants in question are as follows:
- Discord PTB
- Discord Canary
- Discord Development Edition

added discord variants marked
  
All variants are installed through winget, as PTB and Development are not on Chocolatey and Canary was having issues, though if it gets fixed, the choco package name for Canary is `discord-canary`

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -262,7 +262,7 @@
     },
     "discord-canary": {
         "category": "Communications",
-        "choco": "discord-canary",
+        "choco": "na",
         "content": "Discord Canary",
         "description": "Discord is a popular communication platform with voice, video, and text chat, designed for gamers but used by a wide range of communities. (Canary Version)",
         "link": "https://discord.com/",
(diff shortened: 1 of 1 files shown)


FluffyPunk:
We got rid of PTB and Canary because they're too niche + all the custom clients have an ability to switch versions if needed. Unsure about Legcord, since it looks as almost any other client, but with aarch64 support, which is actual only for Linux users

edenchazard:
Personally - and this is just me! - I can't imagine there's a real appetite for the PTB and Canary variants. Happy to be proven wrong!

I'm unsure what Legcord offers for Windows users that Vesktop doesn't, so is there a summary?

Obviously really down to the guys maintaining this project, but my main concern is bloating the program selection with apps that aren't that useful to enough people.

PR #4780, opened 2026-06-30
Add package search fallback for missing install apps

vyas-devgna:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description
Adds a fallback search result when no app matches the local install list.

- Searches the selected source, Winget or Chocolatey, without requiring either CLI to be installed
- Shows the result in the existing app entry style under Search Results
- Keeps fallback results unchecked until selected
- Refreshes results when switching package manager during search
- Avoids adding `na` package IDs during install selection
- Caches repeated searches for faster UI response

## Screenshots
{982ADB21-FAE4-4056-90B1-0C66C2E2630E}
{9A1AD1CE-9057-4964-9BF6-A87507A0B272}
image

## Issue related to PR
- Resolves #

### DIFF
--- a/functions/private/Find-AppsByNameOrDescription.ps1
+++ b/functions/private/Find-AppsByNameOrDescription.ps1
@@ -40,6 +40,11 @@ function Find-AppsByNameOrDescription {
     try {
         # Reset the visibility if the search string is empty or the search is cleared
         if ([string]::IsNullOrWhiteSpace($SearchString)) {
+            if ($null -ne $sync.SearchResultsWrapPanel) {
+                $sync.SearchResultsWrapPanel.Children.Clear(); $Apps = $sync.configs.applicationsHashtable
+                $customApps = $sync.selectedApps | Where-Object { $_ -like "WPFInstallCustom_*" }
+                $customApps | ForEach-Object { $sync.$_ = Initialize-InstallAppEntry -TargetElement $sync.SearchResultsWrapPanel -appKey $_; $sync.$_.IsChecked = $true }
+            }
             $sync.ItemsControl.Items | ForEach-Object {
                 # Each item is a StackPanel container
                 $_.Visibility = [Windows.Visibility]::Visible
@@ -65,13 +70,19 @@ function Find-AppsByNameOrDescription {
                     }
                 }
             }
+            if ($null -ne $sync.SearchResultsWrapPanel) { $sync.SearchResultsWrapPanel.Parent.Visibility = if ($customApps) { [Windows.Visibility]::Visible } else { [Windows.Visibility]::Collapsed } }
             return
         }
 
         # Escape wildcard characters for literal matching
         $escapedSearchString = [System.Management.Automation.WildcardPattern]::Escape($SearchString)
+        if ($null -ne $sync.SearchResultsWrapPanel) {
+            $sync.SearchResultsWrapPanel.Children.Clear()
+            $sync.SearchResultsWrapPanel.Parent.Visibility = [Windows.Visibility]::Collapsed
+        }
 
         # Perform search
+        $anyMatch = $false
         $sync.ItemsControl.Items | ForEach-Object {
             # Each item is a StackPanel container with Children[0] = label, Children[1] = WrapPanel
             if ($_.Children.Count -ge 2) {
@@ -114,6 +125,7 @@ function Find-AppsByNameOrDescription {
 
                 # If category has matches, show the WrapPanel and update the category label to expanded state
                 if ($categoryHasMatch) {
+                    $anyMatch = $true
                     $wrapPanel.Visibility = [Windows.Visibility]::Visible
                     $_.Visibility = [Windows.Visibility]::Visible
                     # Update category label to show expanded state (-)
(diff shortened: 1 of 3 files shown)


ghost:
why do we need this?

vyas-devgna:
> why do we need this?

So if someone does a fresh windows install and want to install all of the needed apps they would require in one go it would be easier like this

ghost:
> > why do we need this?
> 
> So if someone does a fresh windows install and want to install all of the needed apps they would require in one go it would be easier like this

what? can you make a video to showcase this

vyas-devgna:
> > > why do we need this?
> > 
> > 
> > So if someone does a fresh windows install and want to install all of the needed apps they would require in one go it would be easier like this
> 
> what? can you make a video to showcase this

sometimes a user searches for an app that is available through Winget or Chocolatey, but it is not already listed inside WinUtil’s built-in install list. For example, after a fresh Windows install, someone may search for “Obsidian” along with their usual apps. If WinUtil’s local list does not contain it, the search currently feels like “nothing exists,” even though the app can still be installed from the selected package source. This fallback simply gives the user one extra search result from Winget or Chocolatey so they can select it like a normal app, instead of leaving WinUtil, searching manually, and installing it separately.

ghost:
> > > > why do we need this?
> > > 
> > > 
> > > So if someone does a fresh windows install and want to install all of the needed apps they would require in one go it would be easier like this
> > 
> > 
> > what? can you make a video to showcase this
> 
> sometimes a user searches for an app that is available through Winget or Chocolatey, but it is not already listed inside WinUtil’s built-in install list. For example, after a fresh Windows install, someone may search for “Obsidian” along with their usual apps. If WinUtil’s local list does not contain it, the search currently feels like “nothing exists,” even though the app can still be installed from the selected package source. This fallback simply gives the user one extra search result from Winget or Chocolatey so they can select it like a normal app, instead of leaving WinUtil, searching manually, and installing it separately.

okay i see but you should leave that to unigetui

vyas-devgna:
> > > > > why do we need this?
> > > > 
> > > > 
> > > > So if someone does a fresh windows install and want to install all of the needed apps they would require in one go it would be easier like this
> > > 
> > > 
> > > what? can you make a video to showcase this
> > 
> > 
> > sometimes a user searches for an app that is available through Winget or Chocolatey, but it is not already listed inside WinUtil’s built-in install list. For example, after a fresh Windows install, someone may search for “Obsidian” along with their usual apps. If WinUtil’s local list does not contain it, the search currently feels like “nothing exists,” even though the app can still be installed from the selected package source. This fallback simply gives the user one extra search result from Winget or Chocolatey so they can select it like a normal app, instead of leaving WinUtil, searching manually, and installing it separately.
> 
> okay i see but you should leave that to unigetui

yeah but if i can install all needed apps at once in the same interface then it would be better right ?

ghost:
it dosen't work
{64EAC96C-4227-4DFC-8ADD-D646C631C246}

```ps1
PS C:\Users\Gabi> winget search calc
Name                          Id                             Version       Match
--------------------------------------------------------------------------------------------------
Vistumbler                    AndrewCalcutt.Vistumbler       10.8.2
```

vyas-devgna:
> it dosen't work {64EAC96C-4227-4DFC-8ADD-D646C631C246}
> 
> ```powershell
> PS C:\Users\Gabi> winget search calc
> Name                          Id                             Version       Match
> --------------------------------------------------------------------------------------------------
> Vistumbler                    AndrewCalcutt.Vistumbler       10.8.2
> ```

oh damn

vyas-devgna:
> it dosen't work {64EAC96C-4227-4DFC-8ADD-D646C631C246}
> 
> ```powershell
> PS C:\Users\Gabi> winget search calc
> Name                          Id                             Version       Match
> --------------------------------------------------------------------------------------------------
> Vistumbler                    AndrewCalcutt.Vistumbler       10.8.2
> ```

i am using an unofficial web api maybe thats why , will work on this

ghost:
> > it dosen't work {64EAC96C-4227-4DFC-8ADD-D646C631C246}
> > ```powershell
> > PS C:\Users\Gabi> winget search calc
> > Name                          Id                             Version       Match
> > --------------------------------------------------------------------------------------------------
> > Vistumbler                    AndrewCalcutt.Vistumbler       10.8.2
> > ```
> 
> i am using an unofficial web api maybe thats why , will work on this

i still think it's a bad idea most people won't even tho about this and it feels very weird also it's hard to search for programs because if winutil already has something similar to what your searching for than it will show what is inside of winutil and not what you searched for overall i think it's a bad idea

vyas-devgna:
Fixed this. The search now uses the same local winget search that users run in PowerShell, so searching calc returns Vistumbler like your example. If winget is not available, it still falls back to the web lookup.

vyas-devgna:
> > > it dosen't work {64EAC96C-4227-4DFC-8ADD-D646C631C246}
> > > ```powershell
> > > PS C:\Users\Gabi> winget search calc
> > > Name                          Id                             Version       Match
> > > --------------------------------------------------------------------------------------------------
> > > Vistumbler                    AndrewCalcutt.Vistumbler       10.8.2
> > > ```
> > 
> > 
> > i am using an unofficial web api maybe thats why , will work on this
> 
> i still think it's a bad idea most people won't even tho about this and it feels very weird also it's hard to search for programs because if winutil already has something similar to what your searching for than it will show what is inside of winutil and not what you searched for overall i think it's a bad idea

i get it , will close this one

PR #4781, opened 2026-06-30
Add font scaling keyboard and mouse shortcuts

### LINKED ISSUE #4777: Feature Request: Font size should be adjustable via ctrl+/- & via ctrl mousewheel

When I first launched the application, my immediate instinct was to use Ctrl + ,  & Ctrl + Mouse Wheel to adjust the font size because the default text was too small for my eyes. When none of those shortcuts worked, my next reaction was to close the application. Only then did I notice the "A" button for changing the font size.

These shortcuts are widely supported across browsers & many day-to-day use applications, making them the most intuitive way to adjust text size. While having a dedicated accessibility control in the UI is great, common actions like changing the font size should also support the standard keyboard & mouse shortcuts that users naturally expect.

Supporting both approaches would improve discoverability, accessibility, & usability. The UI control would remain available for users who prefer it, while the familiar shortcuts would provide a seamless experience for user like me, who instinctively reach for them.

vyas-devgna:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description
Adds common font scaling shortcuts:
- Ctrl + Plus / Minus adjusts font size by 5%
- Ctrl + Mouse Wheel adjusts font size by 5%

Reuses the existing font scaling logic and keeps the UI slider in sync.

## Issue related to PR
- Resolves #4777

### DIFF
--- a/scripts/main.ps1
+++ b/scripts/main.ps1
@@ -268,6 +268,8 @@ $sync.SearchBarClearButton.Add_Click({
 })
 
 # add some shortcuts for people that don't like clicking
+function Set-WinUtilFontScale([double]$Step) { $sync.FontScalingSlider.Value = [math]::Max(0.75, [math]::Min(2.0, $sync.FontScalingSlider.Value + $Step)); Invoke-WinUtilFontScaling -ScaleFactor $sync.FontScalingSlider.Value }
+
 $commonKeyEvents = {
     # Prevent shortcuts from executing if a process is already running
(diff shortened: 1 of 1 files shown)


ChrisTitusTech (approved):
Verified the font scaling shortcut fix after updating the branch. Local compile passes and the refreshed GitHub Compile-and-Check run is green.

seanh1995:
i can confirm this works great

PR #4560, opened 2026-05-26
fix(iso): resolve windows update re-enabling failures on first boot

vyas-devgna:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] UI/UX improvement

## Description
Fixes the Windows Update settings page displaying 'something went wrong' on custom ISO setups (especially under offline/ARM conditions).
1. Bypasses Service Control Manager (SCM) tamper protection by using the Set-Service API in FirstLogon.ps1 rather than raw registry edits (reg.exe add) on the live system.
2. Uses error handling to catch setup failures on protected services (like UsoSvc/WaaSMedicSvc) and emits warnings rather than failing silently.

## Issue related to PR
- Resolves #4556

Issue #4556, opened 2026-05-26
Windows Update not re-enabling


hauomu:
### I have read the known issues

- [x] Yes, I did

### What part of Winutil are you having issues with?

Windows update not re-enabling after installation 

### Provide a clear and concise description of the issue.

*Disable Windows Update during OOBE — re-enabled automatically on first login

For Arm pcs, If installation was done offline with no access to internet (i.e because fresh install lacked wifi drivers), windows update **does not re-enable** even after pc is connected to internet subsequently
"something went wrong. Try to reopen Settings later.

### List the tweaks you applied before the issue occurred.

Versions issue occured on:
Windows 11 Arm Enterprise Iot LTSC
Windows 11 Arm Enterprise Iot

hauomu:
*Edit (Retried installing with the drivers injected and connected to internet) Issue occurs on both versions of windows irregardless of wether installation is done offline or online. Installations were done with Local account, because I lack education or coorporate email

Additionally (not sure if this is a bug on windows side), might be realted to #4543 Taskbar allignment from settings broken, unable to remove notification icon from taskbar debloated Windows 11 Arm Enterprise Iot LTSC, even after various regeedit re-enabling and then removing. Option does not show up

<img width="1536" height="2048" alt="Image" src="https://github.com/user-attachments/assets/101174ca-00fd-43cd-8fa1-058d7990ff1f" />

PR #4807, opened 2026-07-04
fix(iso): resolve windows update re-enabling failures on first boot

ChrisTitusTech:
## Summary

Recreates PR #4560 by changing the FirstLogon script embedded in `tools/autounattend.xml` to re-enable Windows Update related services through `Set-Service` instead of direct service `Start` registry writes.

## Why

Custom ISO setup can hit Windows service control/tamper behavior when restoring update services from raw registry edits. Using `Set-Service` follows the service control path and keeps the intended startup types for `BITS`, `wuauserv`, `UsoSvc`, and `WaaSMedicSvc`.

## Validation

- `./Compile.ps1`
- `Import-Module Pester -RequiredVersion 5.8.0 -Force; Invoke-Pester -Path 'pester/win11creator.Tests.ps1' -Output Detailed`

Recreates #4560. Addresses #4556.

### DIFF
--- a/pester/xaml.Tests.ps1
+++ b/pester/xaml.Tests.ps1
@@ -310,6 +310,7 @@ Describe "XAML and sync wiring" {
             "InstallAppEntriesRendered",
             "ProgressBar",
             "progressBarTextBlock",
+            "FontScaleFactor",
             "Win11ISOImageInfo",
             "Win11ISODriveLetter",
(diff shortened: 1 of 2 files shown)


wolfxb:
hay chris i have ask in you tools for make windows fast - i can create custom win 10 because my computer is too weak to run Windows 11

PR #4818, opened 2026-07-07
Add Fan Control to applications.json.

eduardodepaiva:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description

Adding FanControl to the `applications.json` list.

- **Added to Pro Tools:** Placed it alongside HWiNFO and HWMonitor since it's an advanced hardware control tool.
- **Why:** It's widely considered the best lightweight, highly customizable fan controlling software for Windows, replacing the need for bloated motherboard vendor software and can control fans that MSI Afterburner cannot.

Tested locally via `Compile.ps1` and it renders perfectly.

## Issue related to PR

N/A

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -413,6 +413,15 @@
         "winget": "EpicGames.EpicGamesLauncher",
         "foss": false
     },
+    "fancontrol": {
+        "category": "Pro Tools", 
+        "choco": "na",
+        "content": "Fan Control",
+        "description": "A deeply customizable and lightweight fan controlling software for Windows.",
+        "link": "https://getfancontrol.com/",
+        "winget": "Rem0o.FanControl",
+        "foss": false
+    },
     "files": {
         "category": "Utilities",
         "choco": "files",


FluffyPunk:
Eh... I'm not a fan(ba-dum, ts...) of this for 2 reasons:
- If your BIOS doesn't have quick setting for fan speed settings, your motherboard sucks.
- Is there anything not that old enough that is not controlled by Afterburner? Cause my old 2060 could be controlled through it, being old enough already .___.

eduardodepaiva:
@FluffyPunk , I appreciate the pun, but...

- Fan Control manages GPU, CPU, chassis fans, and liquid cooler pumps. Afterburner handles _only_ GPU fans.
- Fan Control lets you apply fan curves on the fly while benchmarking. BIOS requires a _reboot_ for every change.
- Fan Control can tie chassis fans to GPU temps (or a CPU/GPU mix). BIOS curves consider _only_ CPU temps.
- Plus, BIOS lacks step-up/down response time configuration.

Afterburner and BIOS work, but are incomplete.

PR #4829, opened 2026-07-13
Add maximize button logic and update assets

aran628:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [x] UI/UX improvement

## Description
adds a maximize button

## Issue related to PR
search bar will not be seen 
- Resolves #
only happens in laptop screens


mewclouds (changes requested):
I appreciate the addition of a maximize button, as it does provide more reading space, but please note that PR #4831 already addresses the search bar overlap issue. Additionally, the maximize icon itself is currently misaligned and inconsistent in weight compared to the other window controls, so that design element will need to be corrected before we can move forward

image

PR #4831, opened 2026-07-13
Fix search bar overflowing behind icon buttons at narrow window sizes

YayoRazo:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor
- [x] UI/UX improvement

## Description

The search bar (top toolbar) is invisible/hidden behind the theme/font/settings icon buttons whenever the window is smaller than its natural auto-sized width. Root cause: the search `Border`/`TextBox` had a hardcoded fixed `Width="{DynamicResource SearchBarWidth}"` (200px) with `HorizontalAlignment="Left"` instead of respecting the `2*` column already reserved for it - so it doesn't shrink with the window, it just overflows past its column boundary, and the icon buttons (declared later in the XAML, so painted on top) end up covering it.

Fix:
- Search `Border`/`TextBox` now use `HorizontalAlignment="Stretch"` with a `MinWidth="120"` floor instead of a fixed pixel width, so they fill the column's actual available space instead of overflowing it.
- `SearchBarClearButton` used a hardcoded `Margin="213,0,0,0"` assuming the old fixed 200px width - moved it to `HorizontalAlignment="Right"` so it tracks the search box's real (now flexible) right edge instead of a stale pixel offset.
- Raised the window's `MinWidth` from 800 to 1150 - at 800, the nav tab buttons plus the icon buttons alone already consumed nearly all the available width, leaving no room for a usable search box regardless of this fix.

## Verification

- `.\Compile.ps1` succeeds.
- Full Pester suite: 397/398 passing (the 1 failure is pre-existing on `main`, unrelated - see the progress-indicator PR for details).
- Manually resized the window down to its new minimum and confirmed the search bar stays visible and usable instead of disappearing behind the icon buttons.

## Issue related to PR
- Resolves #


mewclouds:
I can still reproduce the issue on this PR: when resizing the window narrower, the top-right search box is still pushed/hidden behind adjacent toolbar elements (tabs/icon buttons), so it is not consistently visible/usable at constrained widths.

I tested against the behavior described in this PR and expected the search field to remain visible down to the enforced minimum width, but in practice it still gets **occluded:**

image


Suggested follow-up:
- Revisit top bar layout priorities so the search container has guaranteed visible space at min width (or intentionally collapses in a controlled way).
- Consider responsive behavior for narrow widths (e.g., reduce/hide tab text, move actions into overflow, or switch search to icon-expand mode).
- Add a deterministic UI check at minimum width to prevent regressions.

Happy to re-test once updated!

mewclouds:
Thanks for the fix @YayoRazo! The search bar stays consistent even when shrinking the window now. *You can safely ignore the errors in the bg they're fixed in main*

image

mewclouds (approved):

ChrisTitusTech (approved):
Reviewed the updated diff after merging current `main`. The responsive search layout and runtime/XAML minimum-width values are consistent, the merge is clean, local compile plus focused XAML/sanity tests pass, and the refreshed GitHub checks are green. No blocking findings.

ChrisTitusTech:
@codex review

PR #4834, opened 2026-07-13
Fix null-reference errors in app entry rendering ($Apps.$appKey -> $app)

### LINKED ISSUE #4833: Initialize-InstallAppEntry.ps1 throws null-reference errors for every rendered app card ($Apps.$appKey should be $app)
## Problem

Every app-card render on the Install tab throws:
```
InvalidOperation: ...\winutil.ps1:954
 954 |          $fallback.Text = $Apps.$appKey.content.TrimStart(".").Substri …
     | You cannot call a method on a null-valued expression.
```
repeated once per app card, flooding the console.

## Cause

`functions/private/Initialize-InstallAppEntry.ps1` computes `$app = $sync.configs.applicationsHashtable.$appKey` early in the function, but 4 later references still use the stale/out-of-scope `$Apps.$appKey` (capital A) instead of the local `$app`: lines 72, 74, 78, and 81.

## Existing test coverage

The repo's own Pester suite already covers this exact scoping bug and it's currently failing on `main`: `pester/install-rendering.Tests.ps1` -> "Install app rendering startup contract - keeps app-entry metadata lookup independent from the old caller scope".

## Suggested fix

Replace the 4 `$Apps.$appKey` references with `$app` in `Initialize-InstallAppEntry.ps1`.

YayoRazo:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description

`Initialize-InstallAppEntry.ps1` computes the correctly-scoped `$app = $sync.configs.applicationsHashtable.$appKey` early in the function (and already uses `$app` at lines 22, 90, 93, 104, 105), but 4 later references still used the undefined `$Apps.$appKey` (capital A) instead - throwing `InvalidOperation: You cannot call a method on a null-valued expression` for every single app card rendered on the Install tab.

Beyond the console spam, this silently broke the app-icon logic: since `$Apps.$appKey.link` always threw/evaluated as falsy, every app card fell back to showing just a letter instead of its real favicon.

Fix: replace the 4 stale `$Apps.$appKey` references with `$app`, matching the rest of the function.

## Verification

- `.\Compile.ps1` succeeds.
- Full Pester suite: 398/398 passing - the repo's own pre-existing test for this exact bug (`pester/install-rendering.Tests.ps1` -> "keeps app-entry metadata lookup independent from the old caller scope") now passes; it was failing on `main` before this fix.
- Ran the app: no more errors in the console, and app icons now load correctly instead of falling back to letters.

## Issue related to PR
- Resolves #4833

### DIFF
--- a/functions/private/Initialize-InstallAppEntry.ps1
+++ b/functions/private/Initialize-InstallAppEntry.ps1
@@ -69,16 +69,16 @@ function Initialize-InstallAppEntry {
         $icon.SetResourceReference([Windows.FrameworkElement]::HeightProperty, "AppEntryIconSize")
         $icon.Margin = New-Object Windows.Thickness(0, 0, 8, 0)
         $fallback = New-Object Windows.Controls.TextBlock
-        $fallback.Text = $Apps.$appKey.content.TrimStart(".").Substring(0, 1).ToUpper()
+        $fallback.Text = $app.content.TrimStart(".").Substring(0, 1).ToUpper()
         $fallback.FontWeight = "Bold"; $fallback.HorizontalAlignment = "Center"; $fallback.VerticalAlignment = "Center"
-        if ($Apps.$appKey.link) { $fallback.Visibility = "Collapsed" }
+        if ($app.link) { $fallback.Visibility = "Collapsed" }
         $fallback.SetResourceReference([Windows.Controls.TextBlock]::FontSizeProperty, "AppEntryFontSize")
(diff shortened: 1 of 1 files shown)


mewclouds:
Ran pester suite and tested locally. Good changes Yayo!

image

ChrisTitusTech (approved):

PR #4835, opened 2026-07-13
Fix DNS dropdown styling: scope leaked ToggleButton style, theme the popup list

### LINKED ISSUE #4832: DNS dropdown text looks cramped - global ToggleButton style leaks into the ComboBox template
## Problem

In the Config tab, the "DNS - Set to:" dropdown (`WPFchangedns`, `config/tweaks.json:1835`) shows its selected text ("Default") visually cramped/clipped against the box edges instead of having normal padding.

## Cause (best guess, not yet confirmed live)

`xaml/inputXML.xaml:325` declares an implicit, unscoped `<Style TargetType="{x:Type ToggleButton}">` (no `x:Key`) meant for the big top-bar tab buttons. Because it has no key, WPF applies it to every `ToggleButton` in the visual tree - including the internal `ToggleButton` template part inside every `ComboBox` (the ComboBox's own style builds one at `xaml/inputXML.xaml:264`, wrapping a text/arrow `Grid` with `TextBlock Margin="6,3,2,3"`). The implicit tab-button style overrides that template's part with extra chrome (`ButtonGlow`/`BackgroundBorder` borders plus a `ContentPresenter Margin="10,2,10,2"`) sized for a full tab button, not a 25px-tall combo box, stacking margins/borders that squeeze the visible text area down.

## Suggested fix direction

Scope the tab-button style with an `x:Key` (or add an explicit override on the ComboBox's internal `ToggleButton` part) so it stops leaking into other controls' templates.

## Note

Traced via static XAML analysis, not a live run - flagging as an issue rather than opening a PR for a fix I haven't independently verified.

YayoRazo:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor
- [x] UI/UX improvement

## Description

Three compounding issues on the DNS dropdown (`WPFchangedns`), all traced back to the same control:

1. **Cramped closed-box text.** An implicit, unscoped `<Style TargetType="{x:Type ToggleButton}">` (meant only for the top-bar tab buttons - the file even had a stale comment saying `x:Key="TabToggleButton"`, but the actual `x:Key` attribute was missing) was leaking into the ComboBox's internal `ToggleButton` template part, stacking tab-button-sized borders/margins onto a 25px-tall combo box. Fixed by actually giving it the `x:Key` the comment already claimed, and referencing it explicitly from the tab buttons that want it.
2. **White-on-white dropdown list.** No `ComboBoxItem` style existed anywhere, so every item in the open dropdown fell back to WPF's default white background / black text regardless of theme. Added a themed `ComboBoxItem` style (background/foreground/hover/selected), plus an explicit `ContentTemplate` so item text isn't silently repainted by the app's own global implicit `TextBlock` style (which sets an unrelated label color).
3. **Closed-box background falling back to native chrome.** Once decoupled from the leaked tab-button style, the ComboBox's internal `ToggleButton` had no template of its own and fell back to the OS-default beveled/light `ToggleButton` chrome, which doesn't cleanly respect `Background="Transparent"`. Gave it a minimal dedicated template (just a transparent-passthrough `Border` + `ContentPresenter`) so the surrounding themed `OuterBorder` shows through correctly.

The branch was merged with current `main`; the conflict was resolved by preserving `main`'s removal of the AppX top-navigation button while keeping the scoped tab style on the five remaining navigation buttons. Focused regression coverage now verifies that no implicit `ToggleButton` style can leak into ComboBoxes.

## Verification

- `.\Compile.ps1` succeeds.
- Full Pester suite: **442/442 passing** with Pester 5.8.0.
- Focused XAML suite: **13/13 passing**, including the new style-isolation regression.
- `inputXML.xaml` loads successfully through WPF's runtime `XamlReader` in an STA Windows PowerShell process.
- Light and dark theme UI verification screenshots are attached in the PR discussion.

## Issue related to PR
- Resolves #4832


mewclouds:
Ran pester suite locally and produced the same results. Also verified UI changes on both themes.

image

image

mewclouds (approved):

ChrisTitusTech (approved):
Reviewed the updated diff after merging current main. The conflict resolution preserves main's AppX navigation change while keeping the scoped tab style on the remaining navigation buttons. The ComboBox-specific template and themed item style address the reported rendering issues without reintroducing an implicit ToggleButton style. Added focused regression coverage. Verified with Compile.ps1, runtime WPF XAML loading, the full Pester 5.8.0 suite (442/442), and the green Compile-and-Check workflow.

ChrisTitusTech (approved):

PR #4906, opened 2026-08-03
Refactor core UI rendering, search matching, and service maintenance loops

vyas-devgna:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactor
- [ ] UI/UX improvement

## Description
Reduce repeated allocations and pipeline overhead in UI rendering, tweak application, and installed-state detection while preserving the current job layer and startup flow.

- Build UI groups with generic lists and enumerate configuration properties directly.
- Match application names, descriptions, and preset keys literally, without wildcard interpretation.
- Simplify tweak/AppX search and restore categories that were collapsed before searching; support both direct StackPanel and ScrollViewer layouts.
- Consolidate repeated Windows Update service/progress loops without introducing an early abort before service restart.
- Check all requested package managers when both switches are supplied, while keeping detection quiet.
- Preserve main's live toggle/error handling, ISO wizard and mount cleanup, and transcript-safe session logging. ISO operations use the shared session log, which survives working-directory cleanup.

## Validation
- `.\Compile.ps1` succeeds.
- `Import-Module Pester -RequiredVersion 5.8.0 -Force; Invoke-Pester -Path 'pester/*.Tests.ps1' -Output Detailed -CI`: 842 passed, 0 failed, 1 existing skip because the compiled headless test would open a UAC prompt.
- `Add-Type -AssemblyName PresentationFramework` followed by the search-filter Pester suite: 17 passed with real WPF controls, including direct StackPanel and ScrollViewer category layouts.
- Script Analyzer 1.20.0 with `lint/PSScriptAnalyser.ps1`: no errors; existing convention/test-double warnings remain.
- `git diff origin/main --check`: clean. Generated tweak/feature reference pages, ISO orchestration, logging, and startup are unchanged from main.
- Local Codex review completed for the initial merge. Its generated-doc finding concerned incoming main commits, not PR changes; no generated reference pages were edited by this PR update. The final incremental Codex review found no actionable regressions.
- Full interactive GUI and destructive Windows Update/ISO operations were not run for this branch synchronization.

Optional-service detection retains main's behavior: missing optional services are neutral in normal UI detection, while strict environment-report collection marks them as mismatches. This supersedes the earlier conflicting review suggestions without changing main's apply/detect behavior.

## Issue related to PR
- None


functions/public/Invoke-WPFFixesUpdate.ps1
```diff
@@ -42,14 +42,10 @@ function Invoke-WPFFixesUpdate {
...
-    Write-Progress -Id 2 -ParentId 0 -Activity "Stopping Services" -Status "Stopping wuauserv..." -PercentComplete 20
-    Stop-Service -Name wuauserv -Force
-    Write-Progress -Id 2 -ParentId 0 -Activity "Stopping Services" -Status "Stopping appidsvc..." -PercentComplete 40
-    Stop-Service -Name appidsvc -Force
-    Write-Progress -Id 2 -ParentId 0 -Activity "Stopping Services" -Status "Stopping cryptsvc..." -PercentComplete 60
-    Stop-Service -Name cryptsvc -Force
+    foreach ($svc in @("BITS", "wuauserv", "appidsvc", "cryptsvc")) {
+        Write-Progress -Id 2 -ParentId 0 -Activity "Stopping Services" -Status "Stopping $svc..."
+        Stop-Service -Name $svc -Force -ErrorAction SilentlyContinue
+    }
```
mewclouds: The new loop also regresses progress reporting. Each `Write-Progress` call updates the status text, but `PercentComplete` is omitted. `Write-Progress` does not calculate the percentage automatically.

Please calculate and pass progress for each service, as the previous 20/40/60/etc. updates did.

Also, `-ErrorAction SilentlyContinue` suppresses `Stop-Service` failures and allows the repair to continue even if a service could not be stopped. The following Windows Update repair steps require these services to be stopped, so continuing after a failure can leave the repair in a partial or inconsistent state.

Please surface the failure and abort before continuing.

mewclouds (dismissed):

ChrisTitusTech (changes requested):
Feedback needs to be resolved or this PR will be closed out.

vyas-devgna:
All review feedback addressed in b4b04e6:

**CodeRabbit + Codex feedback:**
- **Find-TweaksByNameOrDescription**: Reset now respects collapsed category state (mirrors Find-AppsByNameOrDescription pattern). Renamed `` to `` to avoid shadowing the PowerShell automatic variable.
- **Invoke-WinUtilCurrentSystem**: Missing service now treated as a mismatch (`-not  -or`) instead of silently passing.
- **Invoke-WinUtilISO**: ISO is dismounted before throwing timeout error. Restored per-workdir log file for resume/export diagnostics.
- **Test-WinUtilPackageManager**: Both `-winget` and `-choco` switches are now evaluated when passed together.

**@mewclouds feedback:**
- **Invoke-WPFFixesUpdate**: Restored per-service `PercentComplete` calculation in the Stop-Service loop. Changed `-ErrorAction SilentlyContinue` to `-ErrorAction Stop` with a try/catch that aborts the repair if a service cannot be stopped.

mewclouds:
Thanks! Will review shortly.

mewclouds:
Could the ISO log write directly to `$workDir` from the start? `Win11ISOWorkDir` is set only after success, so early failures still have no per-workdir log. Also, the test suite produces 470 passed, 1 failed. The empty-search test needs updating for the new collapsed-category behavior. Other than that it looks good, thanks for your contributions!

vyas-devgna:
> Could the ISO log write directly to `$workDir` from the start? `Win11ISOWorkDir` is set only after success, so early failures still have no per-workdir log. Also, the test suite produces 470 passed, 1 failed. The empty-search test needs updating for the new collapsed-category behavior. Other than that it looks good, thanks for your contributions!

Both fair, thanks. The workdir log is still gated on Win11ISOWorkDir which is only set after the modify run succeeds, so I will point Log at the workDir the runspace already receives and create the folder before the first line, and since the failure path deletes workDir I will preserve the log before cleanup. For the test, the empty search case builds the category as "+ Privacy" and still expects everything visible, so I will update it to keep collapsed categories collapsed and add a "- Privacy" case for the other branch. Pushing shortly.

vyas-devgna:
Both fixed in d20279d.

The log now goes to `<workDir>.log` beside the working directory rather than inside it, which drops the `$sync["Win11ISOWorkDir"]` gate entirely — `%TEMP%` already exists, so the first line lands before the work directory is created, and the `Remove-Item -Recurse` on the failure path can't reach it. Writing inside `$workDir` would have needed a `New-Item` before the first line plus a move before cleanup, and that move sits in the catch block where a throw would replace the original ISO error.

The empty-search tweak test now covers both branches — `+ Privacy` stays collapsed, `- Explorer` gets restored — mirroring the app-side test. 471/471 locally and in CI, no new PSScriptAnalyzer findings, `Compile.ps1` clean. Also updated the one docs line that named the old log file.

Thanks for the catch.

vyas-devgna:
This PR heavily refactors `Find-AppsByNameOrDescription.ps1`. It should be merged **before** #4955. #4955 has already been rebased and stacked on top of this branch to pre-resolve the resulting merge conflicts.

mewclouds:
@coderabbitai resume

vyas-devgna:
@coderabbitai resume

ChrisTitusTech (approved):

ChrisTitusTech:
@codex review

PR #4924, opened 2026-08-05
Replace the FOSS dot with a corner badge on app entries

MyDrift-user:
## Type of Change

- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [x] UI/UX improvement

## Description

This changes how FOSS applications are marked on the Install tab. Previously a small green dot was appended to the application name, which was not very clear to be such kind of indication on first sight

FOSS applications now carry a green triangle in the top right corner of their entry, with a keyhole on it. The badge is drawn as vector paths in a `Viewbox`, meaning scalable.

Since there is not really a standardized logo for "FOSS" and to avoid missunderstandings i refrained from using official logos like the one of the Open Source Initiative ® and tried to do something similar for recognitial purposes.

The legend in the left panel uses the same badge as a circle. The legend also moved below the buttons. It was rendered above them because entries are sorted by type rank and `Note` shared rank 0 with checkboxes; notes now sort last. The `Order` field in `appnavigation.json` is not read by the renderer, so no config change was needed.

Every app entry now wraps its checkbox in a layout `Grid` rather than only the FOSS ones, so the checkbox handlers reach the border the same way for every entry. This does not change `applications.json` or the `foss` flag it carries.

### Comparison

| Before | After |
| --- | --- |
| winutil-foss-before-detail | winutil-foss-after-detail |
| Green dot after the application name, legend above the buttons | Corner badge on the entry, circle badge legend below the buttons |
### Verification

Verified with:

- local compilation

## Issue related to PR

N/A

### DIFF
--- a/functions/private/Initialize-InstallAppEntry.ps1
+++ b/functions/private/Initialize-InstallAppEntry.ps1
@@ -21,8 +21,9 @@ function Initialize-InstallAppEntry {
         $border.Tag = $appKey
         $border.ToolTip = $app.description
         $border.Add_MouseLeftButtonUp({
-            $childCheckbox = ($this.Child | Where-Object {$_.Template.TargetType -eq [System.Windows.Controls.Checkbox]})[0]
-            $childCheckBox.isChecked = -not $childCheckbox.IsChecked
+            # Resolve through $sync because the border's child is a layout Grid for FOSS entries
+            $childCheckbox = $sync.$($this.Tag)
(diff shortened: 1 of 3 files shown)


vyas-devgna:
This is really cool

mewclouds (approved):
Oh yeah, this looks great. Loving the changes to the UI 😍

ChrisTitusTech (approved):
Amazing Thanks Drift!

PR #4938, opened 2026-08-07
fix: Import now fills all tabs

v-amorim:
## Type of Change
- [x] Bug fix

## Description
Importing a JSON config only checked the boxes on whichever tab was currently open, forcing users to re-import once per tab. Tabs other than Install build their checkboxes lazily on first visit, and the import's `Reset-WPFCheckBoxes` call only touches checkboxes that already exist - so any tab not yet visited later builds its controls with the default unchecked state and never picks up the import.

Fix: re-run `Reset-WPFCheckBoxes` right after a tab's controls are built, so freshly-created checkboxes and toggles sync to whatever is already recorded (import or preset), regardless of visit order.

Added Pester coverage asserting this happens once per tab build. Verified manually by compiling and running the GUI, importing a config, then switching tabs.

## Issue related to PR
- Resolves #

### DIFF
--- a/functions/private/Initialize-WinUtilTabContent.ps1
+++ b/functions/private/Initialize-WinUtilTabContent.ps1
@@ -36,4 +36,7 @@ function Initialize-WinUtilTabContent {
     }
 
     $sync.InitializedTabs[$TabName] = $true
+
+    # Sync freshly built controls to any selections already in $sync.selected* (import/preset).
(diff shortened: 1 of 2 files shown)


ChrisTitusTech (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/ChrisTitusTech/winutil/pull/4938)</sub>

PR #4941, opened 2026-08-07
Fix UI helper errors during headless -Preset and -Config runs

hiphucto:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description

Running WinUtil headlessly with `-Preset` or `-Config` writes a wall of errors into the session log on every run.

`scripts/main.ps1` handles `-Preset` (lines 37-51) and `-Config` (lines 53-65) in early-return blocks that call `Invoke-WinUtilAutoRun` and then exit. Both return **before** line 67 loads `presentationframework` and **before** line 74 creates `$sync["Form"]`, so the tweak workflow calls UI-only helpers against a window that does not exist:

- `Invoke-WPFUIThread` runs `$sync.form.Dispatcher.Invoke(...)` on `$null` and throws `InvokeMethodOnNull`
- `Set-WinUtilTweaksProgressIndicator` cannot resolve `[Windows.Visibility]` and throws `TypeNotFound`

The errors are non-terminating, so tweaks still apply correctly, but a `-Preset Advanced` run logs six error records before the first tweak even starts. Sample from a real run:

```
[INFO] [Tweaks] Tweaks requested: 17 selected tweak(s), DNS provider: Default
You cannot call a method on a null-valued expression.
+     $sync.form.Dispatcher.Invoke([action]$ScriptBlock)
Unable to find type [Windows.Visibility].
+     $indicatorVisible = if ($Visible) { [Windows.Visibility]::Visible ...
```

### Why the guard goes in the helpers

Loading the WPF assembly earlier does not fix this. `System.Windows.Visibility` lives in **PresentationCore**, which only loads once a WPF object is actually instantiated, so `LoadWithPartialName('presentationframework')` leaves the type error unchanged. And giving `$sync.Form` a real window only trades the dispatcher errors for `PropertyNotFound` on `WPFTweaksProgressBar`, `WPFTweaksProgressLabel` and `WPFTweaksProgressValue`, which are only populated at `main.ps1:135`.

The repo already has a `$hasUI = $null -ne $sync.Form -and $null -ne $sync.Form.Dispatcher` convention in `Invoke-WPFInstall`, `Invoke-WPFUnInstall`, `Invoke-WPFAppxInstall`, `Invoke-WPFAppxRemoval` and `Invoke-WPFOOSU` — but it was never applied to `Invoke-WPFtweaksbutton`, `Invoke-WPFundoall` or `Invoke-WPFFeatureInstall`, which is exactly why these errors appear. Guarding the two helpers covers all three missed workflows in two lines instead of ~15, and any future call site along with them. The existing `$hasUI` blocks become redundant rather than wrong and are left untouched.

### Stray console output

A second commit removes two pieces of output that a GUI click handler silently discards but that reach the console on automation runs:

- `Invoke-WPFRunspace` returns an `IAsyncResult` that no call site uses, printing a `CompletedSynchronously / IsCompleted` block. The return value itself is kept because `pester/runspace.Tests.ps1` asserts it, so the suppression is applied at the four call sites reachable from `Invoke-WinUtilAutoRun`.
- `Set-WinUtilRegistry` was the only `New-PSDrive` call site not piping to `Out-Null`, printing a PSDrive table mid-run. `Invoke-WinUtilCurrentSystem` and `Get-WinUtilToggleStatus` already suppress it.

### Testing

- New `pester/headless-ui.Tests.ps1` (5 tests) covers the no-window cases for both helpers and asserts the GUI path still marshals onto the dispatcher and still updates the indicator controls. The three no-window tests were confirmed to fail against the pre-fix functions, and the two GUI-path tests pass both before and after, so the guard provably does not disable the GUI behaviour.
- Full Pester suite: **480 passed, 0 failed**.
- `.\Compile.ps1` succeeds.
- PSScriptAnalyzer with `lint/PSScriptAnalyser.ps1` on the changed files: no new findings.
- Verified against a live `-Preset Advanced` run that all 17 tweaks apply both before and after this change, including all 12 registry keys for `WPFTweaksTelemetry`. This PR only removes the error noise; it does not change which tweaks run.

No behaviour change in the GUI: when a window exists, both helpers work exactly as before.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -163,6 +163,7 @@ When the user corrects an agent approach, add or tighten one concrete rule here 
 - Import Pester 5.8.0 before running tests so `Invoke-Pester -Output Detailed -CI` does not resolve to Windows' inbox Pester 3.4.0.
 - Keep package install/uninstall process launches simple unless explicitly requested; do not add a separate stdout/stderr process logging helper for winget or Chocolatey.
 - When the active log file is owned by `Start-Transcript`, do not call `Add-Content` against that file; write to host output so the transcript captures the line in the same log file without recording a terminating-error diagnostic.
+- Keep UI helpers such as `Invoke-WPFUIThread` and `Set-WinUtilTweaksProgressIndicator` safe to call without a window; the `-Preset` and `-Config` paths run the workflows before the form is created and before PresentationCore is loaded.
 - Log install/uninstall package names and package-manager IDs before queuing background runspace work; do not rely on runspace host output for the package identity.
 - For Win11 Creator, start each new ISO modification in a fresh `WinUtil_Win11ISO_*` temp directory; existing-work detection is only for resuming/exporting already modified media.
 - For Win11 Creator driver injection, keep offline WIM servicing to one mount, one `/Add-Driver`, and one commit; do not export editions or run unrelated WIM cleanup, and reject damaged metadata before ISO export.
--- a/functions/private/Set-WinUtilRegistry.ps1
+++ b/functions/private/Set-WinUtilRegistry.ps1
@@ -28,7 +28,7 @@ function Set-WinUtilRegistry {
     )
 
     try {
-        if(!(Test-Path 'HKU:\')) {New-PSDrive -PSProvider Registry -Name HKU -Root HKEY_USERS}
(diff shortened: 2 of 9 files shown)


ChrisTitusTech:
@codex review

ChrisTitusTech (approved):

Issue #4877, opened 2026-07-25
Change in JSON file format provides An error occurred while importing: Key cannot be null.


obrientg:
### I have read the known issues

- [x] Yes, I did

### Version of WinUtil

WinUtil 26.07.17

### What part of WinUtil are you having issues with?

Program Install Tab

### Provide a clear and concise description of the issue.

Note I did 

- Review Known Issues (https://winutil.christitus.com/knownissues/)
- Try to test in the latest beta/dev version, but the “sprocket” in the upper right where you would access this capability is aligned with the “minimize” the entire window instead of the expected menu drop town providing access for import/export, amount/documentation/sponsors 
- Research around the internet, using various browsers regarding this issue/bug but could not find a relevant thread/comment/conversation. 

In building a new Win11 install for me today, I grabbed my notes and thumbdrive containing the JSON files from my last Win11 rebuild. Started CTT, went to import the last applications JSON I have but nothing happened – and on the PowerShell CLI the error message received: 

```
Invoke-WPFImpex : An error occurred while importing: Key cannot be null.
Parameter name: key
At line:14686 char:5
+     Invoke-WPFImpex -type "import"
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Invoke-WPFImpex
```

Steps taken to duplicate/troubleshoot issue(s):
-	On the same system being rebuilt, I tried to import older versions of my application JSON files, all giving the same error
-	Using my current daily driver system, I started CTT WinUtil 26.07.17 and attempted to import the same application JSON files, all giving the same error
-	Using CTT WinUtil 26.07.17 on my daily driver I randomly selected a handful of applications, exported as a “test” file and then imported the “test” file into the rebuild. This did not error out.
-	Comparing the old JSON to the newly created JSON identifies sections missing in the JSON file, which is prompting the error
I then went to import & implement my tweaks file, to the same error message. 

Summation/detail of problem: 
The new version of the JSON export is missing the "WPFFeature":  section as well as the “Install":  section that for each application lists the respective source (winget/choco). Currently the exported JSON looks like: 
```
[
    "WPFInstallungoogled",
    "WPFInstallvivaldi",
    "WPFInstallwaterfox",
    "WPFInstallbetterbird",
    "WPFInstalldiscord",
[...]

PR #4966, opened 2026-08-11
fix(system): correct service, update repair, and legacy imports

virtualtable:
## Summary

- pass the PowerShell 5 delayed-start arguments to `sc.exe` with the required spacing and report non-zero exit codes
- remove the WSUS client registry values with PowerShell registry commands instead of malformed `REG DELETE` paths
- import historical grouped WinUtil configurations while ignoring the legacy `Install` package metadata and retaining strict validation for selection keys
- merge the current `main` branch and add regression coverage for the historical import shape

## Validation

- `.\Compile.ps1`
- Pester 5.8.0 full suite: 562 passed, 0 failed
- focused UI-state suite: 20 passed, 0 failed
- PowerShell Script Analyzer completed with only existing accepted repository warnings
- local Codex review loop completed; no actionable findings remain
- legacy fixture confirmed apps, tweaks, toggles, and features import without treating package metadata as selections

## Manual validation

The destructive service and Windows Update repair operations were not run against the host. Their behavior is covered by source review and focused helper tests.


virtualtable:
@coderabbitai review

virtualtable:
@coderabbitai review

mewclouds:
Thanks for working on this. I raised the same concern on #4911, but a decision has not been made in #4877 about what backward compatibility should mean here.

The backup format changed on February 17, almost six months ago. It is up to Chris to decide whether full compatibility is necessary or partial import is acceptable.

This skips the old `Install` section. That may be necessary because the old `winget` and `choco` metadata cannot be passed directly to the current selection importer, but it also means the backup is not necessarily fully restored. That needs an explicit product decision rather than being presented as complete backward compatibility.

I also cannot verify this was tested. There is no legacy JSON fixture, import-path test, or example of the old input and expected selections.

Finally, every provider in `config/dns.json` has `Primary6` and `Secondary6`, so the IPv6 guard is always true. IPv4-only DNS providers are very rare.

virtualtable:
didnt know about #4877 and #4911, my bad. the legacy import thing was just to avoid the null key crash but yeah thats more of a product decision for chris on how far back compat should go. i can drop that + the ipv6 guard if youd prefer to keep this to just the sc.exe and reg fixes, lmk

mewclouds:
> didnt know about #4877 and #4911, my bad. the legacy import thing was just to avoid the null key crash but yeah thats more of a product decision for chris on how far back compat should go. i can drop that + the ipv6 guard if youd prefer to keep this to just the sc.exe and reg fixes, lmk

I'd say keep the legacy import implementation for now and lets wait for Chris' input. The IPv6 guard can be removed. Good work on the bugfixes 😄

virtualtable:
sounds good, removed the ipv6 guard commit as suggested! leaving the rest for chris to review 👍

ChrisTitusTech (approved):
Reviewed the final branch after syncing with main. Local compile and all 562 Pester tests pass, all review threads are resolved, and both Codex and CodeRabbit review loops have no remaining actionable findings.

Issue #4785, opened 2026-07-01
Language Selection


Woff3037:
### 💡 Describe the solution you'd like

Can you add language selection so I can advise this tool to my non English speaker friends (I will also help the translation)

Woff3037:
I decided I will give it a shot myself but please reply if you think it is fitting to the project.

FluffyPunk:
This was already discussed in a lot of places like #2993 #4103 #4032 #4026 etc. This feature is limited due to capabilities and structure of WinUtil, because making one language will add like half a size for compilation outcome.
Also even with fact of asked and even suggested translations, since WinUtil is done in English, priority of translation is way lower rather than some other projects, i.e. Chinese one that has no translations.

Please, track the progress on WinOneShot in CTT Discord, its roadmap includes locale support

/close 'not planned'

FluffyPunk:
@ChrisTitusTech pin this issue as final answer for this question pls

ChrisTitusTech:
Never have found a good solution for this

PR #4991, opened 2026-08-17
feat: add switchable UI language support (English + Simplified Chinese)

kakapo00:
Adds switchable UI language support (English + Simplified Chinese) to WinUtil.

## What's included

- **Runtime language switching** from the new language menu in the top bar (English / Simplified Chinese), persisted in `preferences.json`.
- **Localized UI layers**: tabs, menus, buttons, tooltips, accessibility names, config-driven entries (tweaks/features/AppX/apps), search results, workflow dialogs/progress/logs, and the Win11 ISO creator placeholders.
- **Localized search**: app and tweak search matches both English config strings and the localized display text.
- **Formatted text**: `Get-WinUtilFormattedText` handles `{n}` placeholders with safe English fallbacks; runtime strings go through `Get-WinUtilText`.
- **UTF-8 standardization** for config reading and compiled output.
- **Comprehensive tests**: translation coverage, forward/reverse traversal, reverse-restore idempotency, localized search, fallback behavior, and XAML/config coverage validation.

## Implementation notes

- Language codes live in `config/i18n.json`; `Set-WinUtilLanguage` validates against that config.
- Switching rebuilds rendered tabs and re-wires generated controls (click handlers, package-manager radios), re-applies active searches, and re-syncs ISO placeholder lines.
- A failed switch restores the previous language state, re-renders the current tab, and keeps `preferences.json` consistent.
- The forward traversal records each control's original English key (in its `Uid`) so controls that share one translation (e.g. "Documentation" and "Document" both → 文档) restore to their own key when switching back to English.


kakapo00:
Thanks for the reviews — I've addressed the findings and pushed the fixes to the `i18n` branch (5 commits, `6dfeffe..37c37d3`).

**Addressed**

*Codex P1 + CodeRabbit (functional correctness)*
- **Rebind event handlers after rebuilding**: a language switch now resets the button-registration cache so rebuilt config buttons get their click handlers re-attached, re-wires the WinGet/Chocolatey radio preference handlers, and re-applies the persisted package-manager selection.
- **Failed-switch rollback**: the preference is persisted only after the UI switch succeeds; the catch path restores the previous language, re-renders the current tab, and persists the restored preference, so a failed switch no longer leaves blank panels or a session/preferences mismatch.
- **Non-injective reverse table**: the forward traversal records each control's original English key in its `Uid`; the reverse pass restores from that record first, so "Documentation" and "Document" (both → 文档) come back to their own English keys. Added a regression test.
- **Unknown-state sentinel**: the synthetic combo item is now found/restored by a stable `Tag` marker instead of its localized Content.
- **Category auto-expansion keys**: the raw category travels on the toggle's `Tag` and the WrapPanel, so `AppCategoryAutoExpanded` stays language-independent.
- **LanguagePopup dismissal**: it now hides with the other toolbar popups on form click, window deactivation, and Theme/Settings toggles.
- **Active search and ISO placeholders**: switching re-applies the current Tweaks/AppX/Install query and re-syncs the ISO status-log/path placeholder lines so `Write-WinUtilISOLog` can still replace them.

*CodeRabbit (nitpicks / coverage)*
- `Set-WinUtilLanguage` validates against `config/i18n.json` instead of a hard-coded `ValidateSet`, and rejects an empty translation table.
- `Get-WinUtilInlineSegments` / `Get-WinUtilLanguageText` moved to their matching private function files.
- Traversal tests reset the translation table per test instead of relying on describe order.
- Added localized-search tests for `Find-AppsByNameOrDescription` (translated Content and Description).
- The Chinese menu item now uses the English source key "Simplified Chinese" with the translation in `config/i18n.json`.

**Not changed (false positive)**
- The `Invoke-WPFFixesUpdate` "missing keys" finding: both keys ("Reset Windows Update ", "Stock settings loaded.\n Please reboot your computer") exist in `config/i18n.json` — the checker compared against a literal `\n` instead of the decoded newline. Verified at runtime and by the locales coverage test.

All 617 Pester tests pass, and `Compile.ps1` succeeds.

FluffyPunk:
God's sake how many times should we tell that doing it in PowerShell is bloatful.

#4785 it won't be added

ChrisTitusTech:
Adding every language would break the script no localization is doable in powershell...

PR #4994, opened 2026-08-17
fix: preserve selections when config import is invalid

Suze021:
## Summary
- validate every imported selection before replacing the current selection state
- reject unsupported and unknown selection keys without leaving partial selections behind
- add coverage for valid replacement and invalid import preservation

## Validation
- ./Compile.ps1`n- Invoke-Pester -Path 'pester/*.Tests.ps1' -Output Detailed -CI (552 passed)


ChrisTitusTech (approved):
Reviewed the atomic import behavior, focused regression coverage, and stale-key documentation. Local validation and repository checks pass; the addressed review thread is resolved.

PR #5011, opened 2026-08-19
Add Cloudflare WARP to Pro Tools

untitledmag:
## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description

Added Cloudflare WARP to the WinUtil Pro Tools section.

Changes:
- Added Cloudflare WARP application entry to `config/applications.json`
- Added category, description, Winget ID, Chocolatey package ID, and official download link
- Allows users to easily install Cloudflare WARP through WinUtil

Cloudflare WARP provides a faster, more secure, and more private internet connection by using Cloudflare's network.

## Issue related to PR

No related issue.

## Testing

- Verified the application appears correctly in WinUtil.
- Verified Cloudflare WARP installs successfully through Winget.
- Verified Cloudflare WARP installs successfully through Chocolatey.

PR #5012, opened 2026-08-19
fix(apps): WARP prettify

FluffyPunk:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactor
- [ ] UI/UX improvement

## Description

Beg my pardon, but #5011 was made very plain. So proper description, proper category, proper naming for attribute and proper link, because sending people to dev docs is BAD IDEA

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -2077,13 +2077,13 @@
         "link": "https://github.com/rjpcomputing/luaforwindows",
         "foss": true
     },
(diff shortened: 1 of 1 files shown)


ChrisTitusTech:
meh cf-warp is a downgrade. I want to keep these as descriptive as possible so I can plainly read the JSON.

PR #5013, opened 2026-08-19
fix(apps): warp prettify

FluffyPunk:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [X] Refactor
- [ ] UI/UX improvement

## Description

Fine, let's redo #5012, but with proper property name. Still, other things are pretty misleading, because they don't explain anything close about WARP.
What's WARP client? Why person should go to dev docs to understand that it is VPN service?

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -2077,7 +2077,7 @@
         "link": "https://github.com/rjpcomputing/luaforwindows",
         "foss": true
     },
-    "cloudflare-warp": {
+    "CloudflareWARP": {
         "category": "Utilities",
         "choco": "warp",
         "winget": "Cloudflare.Warp",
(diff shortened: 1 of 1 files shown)


ChrisTitusTech:
I guess we should use the official marketing page, but personally I see why dev docs because it explains the service better. However, winutil is meant for normal users so the marketing page makes more sense.

FluffyPunk:
> I guess we should use the official marketing page, but personally I see why dev docs because it explains the service better. However, winutil is meant for normal users so the marketing page makes more sense. 

Yeah, and pure comparison tells why quadOne link is more informative than "nerd docs"

On quadOne on "learn more" it says in second article that "it is VPN for who don't even know what is VPN". While in dev docs... You open WARP modes, which still won't make sense, unless you read through the lines. So changes are reasonable

ChrisTitusTech:
I'd also try to not rename tweaks or function names unless absolutely needed.

ChrisTitusTech (approved):

PR #5053, opened 2026-09-03
style(ui): clarify scope in the note

eduardodepaiva:
Clarify which command affects only the current user and which affects all users.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [x] UI/UX improvement

## Description

Clarify, in the AppX Removal menu, which command affects only the current user and which affects all users in the already existing note.

## Issue related to PR

- Resolves #

### DIFF
--- a/xaml/inputXML.xaml
+++ b/xaml/inputXML.xaml
@@ -1937,13 +1937,13 @@
                             </Grid>
 
                             <Border Grid.Row="2" Style="{StaticResource BorderStyle}" Margin="5,15,5,5">
-                                <StackPanel Background="{DynamicResource MainBackgroundColor}" Orientation="Horizontal" HorizontalAlignment="Left">
+                                <WrapPanel Background="{DynamicResource MainBackgroundColor}" Orientation="Horizontal" HorizontalAlignment="Left">
                                     <TextBlock Padding="10" TextWrapping="Wrap" Foreground="{DynamicResource MainForegroundColor}">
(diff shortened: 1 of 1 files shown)


mewclouds:
Thanks for putting this together.

The main issue with this change is the phrasing "prevent them from being reinstalled for new users". WinUtil deprovisions the packages so new user profiles do not get them by default, but it does not apply policies or locks that prevent Windows from reinstalling them later. Deprovisioning stops Windows from giving the app to new profiles by default. It does nothing to stop the app from coming back through the Store, through Windows' suggested apps, or from someone reinstalling it. That's why "prevent them from being reinstalled" still promises more than the feature delivers, and I'd still keep the current wording. If you'd rather take another pass at it anyway, something like "Remove Selected removes packages for all users and stops new profiles from getting them by default" would describe it accurately.

The original description for "Install Selected" also explains the fallback behavior: it tries registering an offline local manifest first before falling back to the Microsoft Store via WinGet (when it does hit the network).

Because the existing note is more accurate about how both operations work under the hood, I'd suggest keeping the current wording for now.

eduardodepaiva:
@mewclouds Thanks for the detailed explanation — I wasn't aware of the distinction between deprovisioning and actually preventing reinstallation. I've updated the note to your suggested wording.

mewclouds:
@coderabbitai review

PR #5066, opened 2026-09-07
Add AB Download Manager

KarimBenkirane:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description
Adds **AB Download Manager** to the `Utilities` category.

AB Download Manager is an open-source download accelerator that offers multi-threaded downloading, queue scheduling, and browser integration. It serves as a great modern, FOSS alternative to Internet Download Manager (IDM).

### DIFF
--- a/config/applications.json
+++ b/config/applications.json
@@ -17,6 +17,15 @@
         "winget": "7zip.7zip",
         "foss": true
     },
+    "abdownloadmanager": {
+        "category": "Utilities",
+        "choco": "ab-download-manager",
+        "content": "AB Download Manager",
+        "description": "AB Download Manager is an open-source download accelerator and manager with multi-threaded downloads, queue scheduling, speed limiting, and browser integration.",
+        "link": "https://abdownloadmanager.com/",
+        "winget": "amir1376.ABDownloadManager",
+        "foss": true
+    },
     "adobe": {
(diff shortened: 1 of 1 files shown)


FluffyPunk:
Eeeeeeeeeeeeeeeehm... I can't understand impact of this app. I see almost 18k stars, but I can't understand when this app can be useful except downloading HELLA big files like some archives(that are probably present in torrent) and when hosting server(or its network) sucks(which is also rare case nowadays).
That feels like overcomplication of downloading process. Unless there is some real often case that I'm missing

KarimBenkirane:
> Eeeeeeeeeeeeeeeehm... I can't understand impact of this app. I see almost 18k stars, but I can't understand when this app can be useful except downloading HELLA big files like some archives(that are probably present in torrent) and when hosting server(or its network) sucks(which is also rare case nowadays). That feels like overcomplication of downloading process. Unless there is some real often case that I'm missing

I just thought about adding it since IDM is already in the Utilities list, and AB Download Manager is a great open-source, license-free alternative that I personally use and rely on frequently.

ChrisTitusTech (approved):

PR #5082, opened 2026-09-12
feat: expand application catalog and improve package detection

mhdnazrul:
## Type of Change
- [x] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactor
- [ ] UI/UX improvement

## Description

This PR expands the WinUtil application catalog and improves package detection and Winget handling.

### Application Catalog

- Added 33 new applications across the existing WinUtil categories.
- Added Google Antigravity IDE using its verified Winget package.
- Added and verified Chocolatey mappings where reliable packages are available.
- Kept the existing application catalog structure and categories.
- Removed Super Productivity because its currently referenced Winget installer returns HTTP 404.

### Package Management Improvements

- Added a safe Winget uninstall fallback for the `0x8A150014` source-metadata failure.
- Added Winget detection fallback for applications that do not have a Chocolatey mapping.
- Preserved the existing Chocolatey-preferred behavior where a valid Chocolatey package is available.

### Compatibility and Testing

- Improved Pester test compatibility with Windows PowerShell 5.1.
- Compile validation passes successfully.
- Pester: 832/832 tests pass.
- Winget detection/execution paths were validated.
- Chocolatey metadata was validated.

### Known Limitations

Real Chocolatey installation/uninstallation testing could not be performed in the development environment because administrative elevation was unavailable.

VLC uninstall behavior was also observed to hang at the Winget/installer/elevation level. No unsafe package-specific workaround was introduced.

### Validation

- `.\Compile.ps1` — PASS
- Pester — 832/832 PASS
- PowerShell 5.1 compatibility — validated
- PowerShell 7+ compatibility — validated
- Winget paths — validated
- Chocolatey metadata — validated
- Working tree — clean

No unrelated application catalog changes or unrelated refactors are included in this PR.


mewclouds:
This PR bundles four separate changes into one commit, and the new apps don't come with any reasoning. The app picks look reasonable: all 33 winget IDs exist and none are duplicates. Splitting it up would make this much easier to review.

**Please split this into separate PRs:**

1. **New apps**, ideally in smaller batches. For each app, add a line saying what it is and why it belongs in WinUtil. Please also double-check the Chocolatey IDs. For example, `appflowy` on Chocolatey is still on 0.6.8 while winget has 0.14.x.
2. **The winget uninstall retry**, with steps to reproduce the problem.
3. **The detection change** in `Invoke-WinUtilCurrentSystem`.

The PowerShell 5.1 change in `winoneshot-compat.Tests.ps1` can be dropped. CI runs Pester under pwsh, so that branch never runs.

**applications.json**

The file looks like a script loaded it and saved it again, instead of it being edited by hand. That's where the `\u0027` and `\u0026` come from. They're JSON escapes for `'` and `&`, so they parse the same, but they fill the diff with noise. The same save also turned Tailscale's `®` into `Â®`, which shows up in the app, added a BOM, and removed the trailing newline. Could you restore the file from main and add only the new entries by hand?

**Uninstall retry**

`0x8A150014` is `APPINSTALLER_CLI_ERROR_NO_APPLICATIONS_FOUND`, winget's normal "nothing matched" result. It isn't specific to lost source metadata. The retry would run for every app that simply isn't installed through winget. It also drops `--source`, which loosens the match during an uninstall. If you've hit a real case that needs this, please include the winget version, the package, the `winget list` output, and a Pester test in the new PR.

**Detection change**

The new block in the choco branch copies the existing winget branch. Two copies of the parsing will drift apart. The copy also ignores a failed `winget list`, where the original throws. In its own PR, please reuse the existing winget branch for entries without a Chocolatey ID, and add a test that checks the winget call.

Last thing: the description says Super Productivity was removed, but that isn't in this diff or on main. Please update the description to match what the PR actually changes.

FluffyPunk:
Also:
- Syncthing is a duplicate of #5064;
- KDE products on Windows are not welcome, because there are better things than KDE products(the only good thing I can call is Krita, but that should be a call of drawing artists, I heard they use either SAI or Krita)
- Ditto... Clipboard History is a joke?
- C++ is a debate, it is popular usage case, but count of devs is a question
- PDF viewers is a one whole big debate and @ChrisTitusTech pls f- ing deal with that, there is so much tools that do same thing and each person wants their own app
- Same goes for md editors, expanding list with choices when there are already decent choices, even for people who decline usage of "№1 choice"
- Gemini... I will just say no. Even that AI-skeptic guy as me will say "anything BUT Gemini"

Make list clearer, at least remove duplication of apps that raises question "Why you add Y, when you're already added X?". Cause at least I've seen 3 SQL viewers, which makes no sense at all

ChrisTitusTech:
Agreed with @mewclouds and @FluffyPunk feedback

Issue #5095, opened 2026-09-17
[Bug Report] - Win11 Creator leaves WaaSMedicSvc disabled; servicing changes appear to cause AppX/Store failures after Windows Update


darkchina811:
### I have read the known issues

- [x] Yes, I did

### Version of WinUtil

26.08.19

### What part of WinUtil are you having issues with?

Win11 Creator Tab

### Description of the issue

I created a Windows 11 Pro 25H2 ISO using WinUtil/Win11 Creator and performed a clean installation.

After installing all available Windows Updates, several inbox/MSIX applications stopped working:

* Microsoft Store would not launch
* Notepad reported a package dependency/conflict validation error
* Snipping Tool would not launch

DISM and SFC found/repaired corruption but did not fix the applications.

I investigated the generated ISO and found that its `install.wim` itself appears to be intact. The issue appears to be related instead to the post-install scripts added by Win11 Creator.

In particular, I found a reproducible issue with `WaaSMedicSvc`: the post-install script directly disables it by setting `Start=4`, while FirstLogon later attempts to restore it using `Set-Service`. On a fresh installation, that restoration fails with `Access is denied`, but the script uses `-ErrorAction SilentlyContinue`, so the failure is hidden.

I then modified only the setup/post-install scripts, rebuilt the ISO without changing `install.wim`, and clean-installed Windows again. After installing all Windows Updates, Microsoft Store, Notepad and Snipping Tool all continue to work correctly.

## Environment

Windows edition: Windows 11 Pro
Version: 25H2
Base build in generated WIM: 26200.6584
Architecture: x64

After Windows Update, the original affected installation was:

* Build 26200.9457
* Edition: Professional
* DisplayVersion: 25H2

## Reproduction

1. Create a Windows 11 Pro 25H2 ISO using Win11 Creator.
2. Clean-install Windows from the generated ISO.
3. Complete OOBE.
4. Install all available Windows Updates.
5. Test Microsoft Store, Notepad and Snipping Tool.

On my original installation, Store and Snipping Tool stopped launching and Notepad failed package dependency/conflict validation.

## AppX failure observed

Attempting to re-register Notepad produced:

```text
HRESULT: 0x80073CF3

Deployment failed with HRESULT: 0x80073CF3,
Package failed updates, dependency or conflict validation.

Microsoft.WindowsNotepad requires framework
Microsoft.WindowsAppRuntime.2
with minimum version 2.1.3.0,
but the framework could not be found.
```

[...]

PR #5096, opened 2026-09-17
Fix WaaSMedicSvc restoration using direct registry write

CozREV:
## Type of Change
- [x] Bug fix

## Description
Set-Service fails on WaaSMedicSvc with Access Denied since it's a
protected service (LaunchProtected=2). FirstLogon.ps1 silently
swallowed this failure via -ErrorAction SilentlyContinue, permanently
leaving WaaSMedicSvc disabled after setup.

Removes WaaSMedicSvc from the Set-Service restoration loop and
restores it via a direct registry write instead, matching the same
technique already used to disable it in WinUtil-PostInstall.ps1.

Added a test verifying the old Set-Service-based restoration is gone
and the new registry-write fix is present.

Addresses the WaaSMedicSvc portion of #5095; the broader AppX/Store
failure investigation is out of scope for this change.

Update: also changed -ErrorAction SilentlyContinue to Continue on
the Set-Service restoration loop and the WaaSMedicSvc registry
write. SilentlyContinue was suppressing errors before they reached
the output stream FirstLogon.ps1 redirects into its log file, so
any restoration failure — not just WaaSMedicSvc's — was invisible
even in the log. This surfaces those failures in FirstLogon.log
without halting the script.

## Issue related to PR
- Related to #5095


mewclouds (approved):
LGTM !

